Issue #15, opened 2014-10-07
alpm: pacman updates are not tracked


fbourigault:
From Jan de Groot :
> System updates by pacman are not tracked, after running pacman –Syu packageKit will not see that updates are already installed

From kalev :
> the plan is to set up a file monitor on the package database file and use it to remove the prepared updates if anything external touches the database

The alpm local database is a directory tree. Here is the structure :
```
/var/lib/pacman/local/
├── aalib-1.4rc5-10
│   ├── desc
│   ├── files
│   ├── install
│   └── mtree
├── accountsservice-0.6.38-1
│   ├── desc
│   ├── files
│   └── mtree
├── acl-2.2.52-2
│   ├── desc
│   ├── files
│   └── mtree
├── alsa-lib-1.0.28-1
│   ├── desc
│   ├── files
│   └── mtree
├── alsa-plugins-1.0.28-2
│   ├── desc
...
```
Watching the `/var/lib/pacman/local/` directory for changes should be enough.

As this will be implemented in the backend code, I'm waiting for the Hif implementation and then will do the alpm one.

fbourigault:
Hif implementation : 2b7db3423ec96b65f1b5fe4d5b89335813313a64

I will try to do this soon

fbourigault:
In which case I need to invalidate the database ? I think change in the local db (i.e Package was installed/upgraded/downgraded/re-installed/removed). Does I also need to watch for repos package list refresh ? And did I missed an other case ? cc @kalev @hughsie

aleixpol:
Confirmed.

hughsie:
The hif backend inhibits just before it does any action on the package database, and then uninhibits when it's done. This means that any invalidation calls that get triggered get ignored (otherwise, PackageKit does the modification, then notifies itself that it did the action).

fbourigault:
Reopening because I probably wrote a bad implementation.

fbourigault:
I think the local db pkgcache need to be invalidated too. I asked pacman-dev mailing list to know how can I do. https://lists.archlinux.org/pipermail/pacman-dev/2015-March/019997.html

PR #26, opened 2014-12-27
Pacman 4.2 support

piernov:
I'm not sure about all the g_return_if_fail().
Most of the changes are inspired from pacman 4.2 's  src/pacman/callback.c

### DIFF
--- a/backends/alpm/pk-alpm-transaction.c
+++ b/backends/alpm/pk-alpm-transaction.c
@@ -247,25 +247,25 @@ pk_alpm_transaction_progress_cb (alpm_progress_t type, const gchar *target,
 }
 
 static void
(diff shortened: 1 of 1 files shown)


fbourigault:
If backward compatibility is required, contributions are welcome !

fbourigault:
@hughsie is it possible to have a bugfix release ? This will allow PackageKit to be back in Arch Linux repositories !

eworm-de:
I think we still suffer the update problems, no?

fbourigault:
If you mean #15, then yes but PackageKit was in official repos even wiht this known bug. It's my next task but I don't know when it will be ok.

eworm-de:
No, #15 is not what I mean. I've described the problem above:

* Installing packages works, though package is installed, then reinstalled.
* Updates are not installed. Packagekit is busy for some time, then lists the same available updates again.

piernov reported the same, see Arch Linux forum link.

PR #28, opened 2014-12-31
Alpm

eworm-de:
Let's prepare for a proper alpm 9.0.0 / pacman 4.2.0 patch.


fbourigault:
I'm ok with this patch. @hughsie are you feeling good with this ?

hughsie:
@fbourigault -- dude, you're the boss of this backend :)

PR #30, opened 2015-01-04
backends/hif: Move hif_source_is_supported() here

cgwalters:
I think this concept should be based on user's trust of a GPG key or
something more flexible.  Moving it to PackageKit is cleaner as it's
PK that has the concept of "supported", not libhif.


cgwalters:
See also https://github.com/hughsie/libhif/pull/25

PR #31, opened 2015-01-08
alpm: bump required libalpm version number and fix version define

eworm-de:
ALPM_EVENT_PACKAGE_OPERATION_START is enumeration, not define. So we can
not use that in preprocessor.

Bump required libalpm version to 9.0.0 (pacman 4.2.0) and define that.

### DIFF
--- a/backends/alpm/pk-backend-alpm.h
+++ b/backends/alpm/pk-backend-alpm.h
@@ -29,16 +29,11 @@
  * versioning. A patch has been proposed, though:
  * https://lists.archlinux.org/pipermail/pacman-dev/2014-December/019759.html
  *
- * If ALPM_VERSION_NUMBER is *not* defined we test for
- * ALPM_EVENT_PACKAGE_OPERATION_START, which is libalpm >= 9.0.0 only and
- * define ALPM_VERSION_NUMBER on our own. */
+ * There's no usable define in alpm.h in libalpm up to and including 9.0.0.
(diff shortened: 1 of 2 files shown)


eworm-de:
Sorry for the noise... I did not try to use ALPM_EVENT_PACKAGE_OPERATION_START in preprocessor before and noticed it fails while prepare my patch to support older libalpm version.

Looks like we are stuck for now... Anybody with a better idea?

fbourigault:
Maybe we can do something within the configure.ac to define some preprocessor macros for libalpm. I'm not so comfortable with autotools so maybe it's not possible. cc @hughsie

eworm-de:
Prepared somthing in eworm-de/PackageKit branch alpm:
alpm: move versioning to configure: 0c7c459ac309b726bc868d2cf8b61fca1e22598a
alpm: re-introduce support for libalpm < 9.0.0: 6d052353f3c9f074cd13f662ae09afd33e3dc4ad

Does that make sense?

It does compile with libalpm 9.0.0, did not test with libalpm 8.2.0.

fbourigault:
This looks like what I have in mind. Could you do add it to this PR ? I will try it with 4.2 and 4.1 (will need more work !)

eworm-de:
I opened a new PR. So please just merge either of both... ;)

fbourigault:
I will use #33

PR #33, opened 2015-01-08
alpm: reintroduce pacman < 4.2 compatibility

eworm-de:
Try to make it work with libalpm 8.2.0 and 9.0.0.


fbourigault:
Testing with pacman < 4.2 should be possible with http://archlinux.mirrors.ovh.net/archlinux/iso/2014.12.01/

fbourigault:
Before doing anything with this PR, does any distro is still running an older pacman version ? If not I think we just have to drop pacman < 4.2 support to keep codebase clean.

PR #35, opened 2015-01-08
alpm: fix SIGSEV when asking for package files.

fbourigault:


backends/alpm/pk-alpm-packages.c
```diff
@@ -304,17 +306,13 @@ pk_backend_get_files_thread (PkBackendJob *job, GVariant* params, gpointer p)
 		if (pkg == NULL)
 			break;
 
-		root = alpm_option_get_root (priv->alpm);
-		files = g_string_new ("");
-
 		filelist = alpm_pkg_get_files (pkg);
+		files = g_new0 (gchar*, filelist->count + 1);
```
fbourigault: Does this code is leaking memory at each iteration or _cleanup_strv_free_ called each time the end of loop block reached ?

ximion: This is fine, the cleanup function will be called each time the the for-loop is executed.

PR #44, opened 2015-02-03
alpm: track local db changes and invalidate cache

fbourigault:
This PR is about fixing #15 . Here I'm monitoring /var/lib/pacman/local for changes.

As I asked in #15, is this what I must do here or I should also watch repos synced dbs or only the synced db or something else ?

I also have an other question about `pk_backend_transaction_inhibit_start` are those required and in which case I should call it ? I looked at hif calls but I don't have enought knowledge of the hif backend to understand the behavior.

About the code, I think I'm doing things right with threads and memory. But any reviews are welcome.

cc @hughsie @aleixpol @kalev

### DIFF
--- a/backends/alpm/pk-alpm-transaction.c
+++ b/backends/alpm/pk-alpm-transaction.c
@@ -1011,6 +1011,7 @@ pk_alpm_transaction_commit (PkBackendJob *job, GError **error)
 	PkBackendAlpmPrivate *priv = pk_backend_get_user_data (backend);
 	alpm_list_t *data = NULL;
 	_cleanup_free_ gchar *prefix = NULL;
+	gint commit_result;
 
 	if (pk_backend_job_is_cancelled (job))
 		return TRUE;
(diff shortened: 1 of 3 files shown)


aleixpol:
Shouldn't this be checking for changes in /var/lib/pacman/sync/ as well?

hughsie:
I think calling pk_backend_installed_db_changed() like this is the right thing to do, but you'll want to inhibit too.

fbourigault:
I added the inibition calls. Reviews are welcome before merging.

eworm-de:
Looks like this still does not work as expected:

```
root@08-00-27-f1-27-24 ~ # pacman -Q packagekit pacman
packagekit 1.0.5-2
pacman 4.2.1-1
root@08-00-27-f1-27-24 ~ # pkcon get-details wget
Resolving                     [=========================]
Getting details               [=========================]
Package description
  package:     wget-1.16.2-1.x86_64
  summary:     (null)
  license:     GPL3
  group:       other
  description: A network utility to retrieve files from the Web
  size:        2412544 bytes
  url:         http://www.gnu.org/software/wget/wget.html
root@08-00-27-f1-27-24 ~ # pacman -S wget
resolving dependencies...
looking for conflicting packages...

Packages (1) wget-1.16.3-1

Total Download Size:   0.47 MiB
Total Installed Size:  2.33 MiB
Net Upgrade Size:      0.03 MiB

:: Proceed with installation? [Y/n]
:: Retrieving packages ...
 wget-1.16.3-1-x86_64              479.6 KiB   999K/s 00:00 [#############] 100%
(1/1) checking keys in keyring                              [#############] 100%
(1/1) checking package integrity                            [#############] 100%
(1/1) loading package files                                 [#############] 100%
(1/1) checking for file conflicts                           [#############] 100%
(1/1) checking available disk space                         [#############] 100%
(1/1) upgrading wget                                        [#############] 100%
root@08-00-27-f1-27-24 ~ # pkcon get-details wget
Resolving                     [=========================]
Getting details               [=========================]
Package description
  package:     wget-1.16.2-1.x86_64
  summary:     (null)
  license:     GPL3
  group:       other
  description: A network utility to retrieve files from the Web
  size:        2412544 bytes
  url:         http://www.gnu.org/software/wget/wget.html
```

pacman successfully updated package wget to 1.16.3-1, but packagekit does not track the change.

PR #46, opened 2015-02-12
Add dbus method for returning prepared packages

### LINKED ISSUE #45: DBUS method for listing offline updates
It would be nice to expose what updates have been prepared via dbus, similar to the way that pkconn offline-get-prepared does.

petervo:
Fixes #45


petervo:
Changed the check to package_ids == NULL

petervo:
Squashed and rebased.

PR #47, opened 2015-02-13
zypp: Update arch compatiblity check for x86

thp:
Allow installing 32-bit x86 packages on 64-bit x86.

### DIFF
--- a/backends/zypp/pk-backend-zypp.cpp
+++ b/backends/zypp/pk-backend-zypp.cpp
@@ -1022,13 +1022,6 @@ zypp_refresh_meta_and_cache (RepoManager &manager, RepoInfo &repo, bool force = 
 
 
 static gboolean
-system_and_package_are_x86 (sat::Solvable item)
-{
-	// i586, i686, ... all should be considered the same arch for our comparison
-	return ( item.arch() == Arch_i586 && ZConfig::defaultSystemArchitecture() == Arch_i686 );
-}
-
-static gboolean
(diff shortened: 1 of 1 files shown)


thp:
Patch from Nemo Mobile. CC'ing @DimStar77 @sreeves @mlandres for patch review if that arch equality check is also valid on openSUSE (if one can install 32-bit x86 packages on 64-bit x86 installs).

mlandres:
Not sure what the purpose of the ```system_and_package_are_x86 ``` function is (esp. why it is restricted to x68).

If the question is whether ```item.arch()``` is installable in the system, I'd use ```Arch::compatibleWith```. If the restriction to  x68 is artifical, remove the ```if```:
```C++
const Arch &systemArch = ZConfig::defaultSystemArchitecture(); // hw arch
const Arch &systemBaseArch = systemArch.baseArch();            // i386 or x86_64 on x86

if ( systemBaseArch == Arch_x86_64 || systemBaseArch == Arch_i386 )
{
  return item.arch().compatibleWith( systemArch );
}
return false;
``

thp:
PR updated based on Michael's suggestion. The x86 restriction is probably only historic, and after removing the restriction the function decomposes into just a call to `item.arch().compatibleWith(...)`, so I removed the function and replaced it with a direct call to that.

This should now do the right thing for ARM and other architectures as well:

http://doc.opensuse.org/projects/libzypp/13.1/Arch_8cc_source.html#l00274

thp:
*bump* - this should be good to go in now?

PR #50, opened 2015-03-16
Add pk_backend_is_transaction_inhibited

fbourigault:
This let backend know if the current transaction is inhibited.


fbourigault:
@hughsie does this looks like good for you ? You can find the use case in https://github.com/hughsie/PackageKit/pull/51/files#diff-7aa64083b99a8da6a60d638f94bdbb7dR107

I'm open to any suggestion !

PR #56, opened 2015-03-19
alpm: remove RepoEnable

fbourigault:
This remove RepoEnable support from alpm backend because it only toggle repos
status in backend internals. For a good implementation of RepoEnable, we need to
write changes in /etc/pacman.conf that is quite tricky for now. This feature may
be reintroduced later with a good implementation.

More commits will follow to clean the backend code.


aleixpol:
+1 makes sense to me.

fbourigault:
I merged this now as all disabled code is now removed but I think more refactoring should be done in pk-alpm-databases.c

PR #60, opened 2015-04-18
zsh support for command-not-found

wjt:
(At least on Fedora,) `/etc/profile.d/PackageKit.sh` is sourced by zsh already, but the hook function is [named differently in zsh](http://zsh.sourceforge.net/Doc/Release/Command-Execution.html). The first patch defines that function when running under zsh, which makes everything *work*.

The second is cosmetic: the name `bash` is currently hardcoded, which makes the mimicked "command not found" messages stick out. However, as described by [the zsh docs](http://zsh.sourceforge.net/Doc/Release/Command-Execution.html), zsh prints out its own message after the hook returns non-zero. So we get something like this:

```
[wjt@gelf]~/src/PackageKit% does-not-exist
zsh: does-not-exist: command not found...
zsh: command not found: does-not-exist
```

where the first message comes from `pk-command-not-found` and the second from zsh itself. If CNF has any helpful suggestions, they'll appear in between the two messages, making it look less silly. I don't think this is worth trying to do something about, really.

### DIFF
--- a/contrib/command-not-found/PackageKit.sh.in
+++ b/contrib/command-not-found/PackageKit.sh.in
@@ -24,7 +24,8 @@ command_not_found_handle () {
 		@LIBEXECDIR@/pk-command-not-found "$@"
 		retval=$?
 	else
-		echo "bash: $1: command not found"
+		local shell=`basename "$SHELL"`
+		echo "$shell: $1: command not found"
 	fi
 
 	# return success or failure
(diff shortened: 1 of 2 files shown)


hughsie:
A couple of style niggles, but with those looks good, thanks.

hughsie:
Can you rebase patch 3 on top of patch 2 pls. You can use --force if you're using a local branch and this issue will magically update :)

wjt:
@hughsie sure! wasn't sure which workflow you prefer ☹

wjt:
(Let me know if that new squashed patch isn't what you meant.)

PR #61, opened 2015-05-05
Correct punctuation displayed to users while applying offline updates

mattdm:
This is a comma splice (http://en.wikipedia.org/wiki/Comma_splice), and while it's probably horribly pedantic to even mention it, it stares me in the face, taunting, whenever I update my system — so I eventually had to say something.  :)


allanday:
Works for me.

mattdm:
Thanks :)

PR #72, opened 2015-07-21
Failure handling

keszybz:
While testing offline updates I noticed that reboot was scheduled twice when the packagekit offline update failed (see patch one for logs). Two fixes to make sure that either failure is returned or a reboot/shutdown in scheduled.

### DIFF
--- a/client/pk-offline-update.c
+++ b/client/pk-offline-update.c
@@ -183,7 +183,7 @@ pk_offline_update_progress_cb (PkProgress *progress,
 /**
  * pk_offline_update_reboot:
  **/
-static void
+static int
 pk_offline_update_reboot (void)
 {
 	g_autoptr(GError) error = NULL;
@@ -200,7 +200,7 @@ pk_offline_update_reboot (void)
 		sd_journal_print (LOG_WARNING,
 				  "Failed to get system bus connection: %s",
 				  error->message);
-		return;
+		return EXIT_FAILURE;
 	}
 	val = g_dbus_connection_call_sync (connection,
 					   "org.freedesktop.systemd1",
@@ -217,14 +217,16 @@ pk_offline_update_reboot (void)
 		sd_journal_print (LOG_WARNING,
 				  "Failed to reboot: %s",
 				  error->message);
-		return;
+		return EXIT_FAILURE;
 	}
+
+	return EXIT_SUCCESS;
 }
 
 /**
  * pk_offline_update_power_off:
  **/
-static void
+static int
 pk_offline_update_power_off (void)
 {
 	g_autoptr(GError) error = NULL;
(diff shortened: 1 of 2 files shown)


keszybz:
Hm, still discussing this on the mailing list. I think the patches are an improvement over status quo, but you might want to wait a bit to see if the semantics change.

hughsie:
Both patches seem sane to me, but please keep me in the loop on the mailing list discussion. FWIW, fwupd also uses the offline update mechanism too.

keszybz:
On Mon, Jul 27, 2015 at 08:55:23AM -0700, Richard Hughes wrote:
> Both patches seem sane to me, but please keep me in the loop on the mailing list discussion.

There hasn't been much discussion on systemd-devel yet. @poettering
is recently back from vacation, and people are working on a release.
Hopefully the subject will be picked up after that.

> FWIW, fwupd also uses the offline update mechanism too.

It most likely only works as a race condition then :)

Conan-Kudo:
@keszybz Is this still something that needs to be dealt with?

keszybz:
Yeah, those two patches seem still relevant.

Conan-Kudo:
@keszybz Could you rebase the commits against master? And what happened with the mailing list discussion?

kalev:
@keszybz Do you have a link to the mailing list discussion that I could read?

keszybz:
Rebased and updated with a commit that should fix https://bugzilla.redhat.com/show_bug.cgi?id=1430920. Please have a look and test.

> And what happened with the mailing list discussion?

There were a few discussions. Anyway, the "official" recommendations are all gathered in https://www.freedesktop.org/software/systemd/man/systemd.offline-updates.html. If anything is missing, issues/PRs against that page in the systemd bug tracker will be the most effective.

kalev:
Awesome, thanks for updating these, Zbigniew! I'll go and apply the 3rd patch downstream in Fedora to get the blocker fix in.

I'm a bit unsure about the first patch, "pk-offline-update: only try to shutdown or reboot on success". If the user explicitly wanted to shut down how is it a good idea to then reboot on failure? I think I'd prefer to shut down in that case as well.

Let's say a user has hacked the whole day on systemd at home and her eyes are bleeding and she wants to go to bed. She unplugs the computer and goes to turn it off. When doing that she notices that GNOME Shell offers updates as well. She chooses "Install pending software updates" and clicks Power Off and walks away. Next morning, she puts the computer in her bag, knowing it's fully updated and nearly fully charged as well, and heads off to give a presentation. When there, the computer doesn't turn on ... and she realizes that it must have rebooted instead of shutting down and it was actually powered on all nighty and the battery is fully drained now.

I'd like to avoid that if possible :)

keszybz:
> If the user explicitly wanted to shut down how is it a good idea to then reboot on failure? I think I'd prefer to shut down in that case as well.

True, if you requested a shutdown, you probably want a shutdown even if the upgrade fails.

Existing code was subject to a race: a reboot or a shutdown was scheduled, but the unit has OnFailure=reboot.target, so it *might* be possible for the shutdown to be overridden by a reboot anyway. So it seems the right solution is to issue a big warning on failure, and schedule a shutdown or reboot, and exit with success, to avoid the OnFailure handling. If this approach is acceptable, I'll rework the patch.

kalev:
That sounds great to me! Thanks for working on this.

keszybz:
Updated. I squashed the first two commits together, they make more sense as one step now.

PR #75, opened 2015-08-07
libpackagekit-glib: Fix introspection annotations for pk_task_resolve_*

robert-ancell:
Fix 'packages' arguments in pk_task_resolve_sync / pk_task_resolve_async
which weren't marked as string arrays.


ximion:
Thanks for the patch!

PR #77, opened 2015-08-09
libpackagekit-glib: Generate Vala bindings

robert-ancell:
Generate the Vala bindings from the .gir. This means that the Vala package can
stop providing these.


ximion:
Patch looks good & works here, but @hughsie has the last word :-)

robert-ancell:
Sorry for updating after the first review. I added some missing annotations that are not currently present in the version in Vala.

mcatanzaro:
It will be interesting to learn what you are doing with PackageKit!

hughsie:
Looks fine to me,. Feel free to push to master, you've now got commit access. Thanks!

robert-ancell:
The main reason I'm fixing these things is I've updated simple-scan to install SANE drivers if it detects a USB scanner but SANE doesn't find it. I'm also keen to get PackageKit 1.0 into Ubuntu before 16.04.

PR #80, opened 2015-08-19
add .gitignore for yum

eworm-de:


hughsie:
We shouldn't check in these autogenerated files

eworm-de:
Ups, I had some binary files around before .gitignore files were generated...

PR #82, opened 2015-09-06
Gentoo fixes

EvaSDK:
So it appears someone pushed these commits stashed and without asking whether I could submit them for proper credit while I had all this in public here. Maybe a git merge --ours to get back this little bit of history in the main repository would be nice.


hughsie:
Can you prefix all the commits with "portage: " please; this makes writing the release notes easier. I have a feeling a few of the commits could be squashed together with a more user-understandable commit message too. Thanks!

EvaSDK:
I will fix the summaries. As for the number of commits I did as much independant commits as possible as I progressed through the code. I could probably merge the "mostly cosmetics" commits together but they all serve a different purpose. One is PEP8, second is using more pythonic code, last is true cosmetics (alignement, variable names, etc).

EvaSDK:
Should be good now.

EvaSDK:
Thank you.

PR #83, opened 2015-09-07
backend job: Fix a race with thread creation

kalev:
pk_backend_job_thread_setup may run and try to free the thread object
before g_thread_new has returned. The fix is to not store the thread
object inside the job; just do g_thread_unref (g_thread_new (...));

Thanks to Matthias Clasen for pointing out the potential fix.

https://bugzilla.redhat.com/show_bug.cgi?id=1057781


ximion:
Eww, really nasty bug.... I can't see any drawbacks from the patch, so +1 from me.

PR #84, opened 2015-09-22
hif: add support for HTTP proxy

marmarek:


hughsie:
Pushed with a small memory fix; thanks.

PR #88, opened 2015-10-23
Make pk_details_get_size documentation more specific

aleixpol:


ximion:
"for the rest the package size." - I would rather use "otherwise it returns the download size", since "for the rest" sounds strange in that regard...
Patch looks good otherwise, clarifying the documentation here is definitively needed.

PR #90, opened 2015-10-28
Added the variadic functions and macros for the Vala bindings.

tintou:
Added some useful macros in pk-bitfield.h to the Vala bindings.
Allow the use of variadic functions in Vala


hughsie:
Does it make sense to export the PkBitfield stuff into vala, it's really just a simple wrapper around some bit-wise operations.

tintou:
Yeah after some more thinking it's not even really useful

PR #92, opened 2015-10-30
zypp: Raise priority of ZYPP stack updates if a security patch is sha…

mlandres:
…dowed (bsc#951592)

A pending ZYPP stack update must be installed first, before further updates can be installed. Due to this a minor ZYPP stack update may shadow applicable security patches.

The patch raises the priority of a ZYPP stack update to PK_INFO_ENUM_SECURITY, if it prevents the installation of pending security patches. The former `gboolean _updating_self` that tracks the presence of a ZYPP stack update is replaced by `enum class SelfUpdate` to keep the additional info.


coolo:
:+1:

PR #95, opened 2015-11-17
vala: Add sentinel flag to variadic functions.

tintou:
The default sentinel is NULL which is wrong for these two functions.


ximion:
Looks good!

PR #99, opened 2015-12-08
require dbus.socket for offline update

eworm-de:
Looks like anything changed in systemd v228 that makes it skip to start
dbus. We need it for offline update, so add a Requires= to unit file.

This should fix Arch Linux downstream bug:
https://bugs.archlinux.org/task/47234


hughsie:
Sorry, +1 was meant to mean "please push to master" -- I'll do it now :)

PR #100, opened 2015-12-11
zypp: fix build

DimStar77:
b47f98 changed a bunch of gchar* to const gchar* to address some memory leaks.
uri should not have been touched - and it is being g_free()'d as needed.


hughsie:
+1, please merge

PR #101, opened 2015-12-16
alpm: get version macros from pkg-config and write header file

eworm-de:


hughsie:
This does look a bit odd. Doesn't alpm-version.h.in belong *in* libalpm so the defines can be consumed by PK? It seems the wrong way round of doing this imho.

eworm-de:
Yeah, would be my preferred solution. But Allan thinks the other way round:
https://lists.archlinux.org/pipermail/pacman-dev/2015-November/020603.html

I did provide a patch to pacman, but it does not get accepted.

hughsie:
I think Allan is wrong, and providing #defines for version information is something basically every library serious about versioning already does. It's really hard to use the values in a pc file from a .c file, and libalpm is supposed to be making it easy for program to consume the API.

eworm-de:
Full acknowledge. ;)
But I have to tell and convince Allan.

PR #102, opened 2015-12-18
Add getters for PkProgress

robert-ancell:
Makes it easier to be type-safe and works better with language bindings.


ximion:
Wouldn't this be `Since: 1.0.12`? (1.0.12 hasn't been released yet, current release is 1.0.11)
Change looks reasonable to me, but @hughsie has the final word :-)

robert-ancell:
I was just about to ask someone to check that, as I wasn't sure. I'll update it.

hughsie:
Can you rebase these together please, with that +1, thanks.

robert-ancell:
Merged as requested

PR #104, opened 2016-01-25
Improve apts "history.log" logfile when using packagekit

mvo5:
The apt history.log file contains log information when apt runs. By setting the "PACKAGEKIT_CALLER_UID" environment apt will log the calling user in the new "requested-by" tag. 

The "CommandLine::AsString" config option is also recorded in the log. The information stored is mostly useful so that an admin can see that this operation was performed using packagekit.


ximion:
Neat, didn't know about this feature yet!
This is pretty nice!

PR #121, opened 2016-05-01
hif: Add support for build-time vendor configurations

### LINKED ISSUE #119: [Hif] RFE: Build time switch for setting supported sources
<strike>For Mageia, I intend to use `/etc/distro.repos.d` (that will eventually be used in Fedora, planned for F25, see [this PR](https://github.com/rpm-software-management/dnf/pull/416) for details) for my repo configuration directory.

In addition, </strike>The hardcoded list of supported repositories is wrong for Mageia. We have our own list of supported repositories which differs quite a bit from Fedora's (for obvious reasons).

I *could* make a patch to change it when being built in Mageia, but I would prefer if a system to support different vendor configurations at build-time could be added to the Hif backend, so that Fedora and Mageia configurations can be maintained upstream.

Conan-Kudo:
With Mageia moving forward to adopt the PackageKit-Hif backend, it is now necessary to support switching out Fedora-specific and Mageia-specific configurations at build time.

For now, this only contains <strike>swapping out repository directory locations and validating the appropriate prefixes for verifying</strike> supported sources.<strike> Fedora isn't yet using `/etc/distro.repos.d`, and other distributions in the future may use other locations</strike>. Obviously sources would differ among distributions.

In the future, support may be added for dealing with groups, which different distributions use different conventions.

I'm somewhat uncertain about the `configure.ac` modifications, because I'm not very good with autotools, but it seemed to select as I wanted to.

This pull request fixes #119.

### DIFF
--- a/backends/hif/hif-backend-vendor-fedora.c
+++ b/backends/hif/hif-backend-vendor-fedora.c
@@ -1,0 +1,46 @@
+/* -*- Mode: C; tab-width: 8; indent-tabs-mode: t; c-basic-offset: 8 -*-
+ *
+ * Copyright (C) 2016 Neal Gompa <ngompa13@gmail.com>
+ *
+ * Licensed under the GNU Lesser General Public License Version 2.1
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301 USA
+ */
+
+#include "hif-backend-vendor.h"
+
+gboolean
+hif_validate_supported_source(const gchar* id)
+{
+	guint i;
+	const gchar *valid[] = { "fedora",
+				 "fedora-debuginfo",
+				 "fedora-source",
+				 "rawhide",
+				 "rawhide-debuginfo",
+				 "rawhide-source",
+				 "updates",
+				 "updates-debuginfo",
+				 "updates-source",
+				 "updates-testing",
+				 "updates-testing-debuginfo",
+				 "updates-testing-source",
+				 NULL };
+	for (i = 0; valid[i] != NULL; i++) {
+		if (g_strcmp0 (id, valid[i]) == 0)
+			return TRUE;
+	}
+	return FALSE;
+}
--- a/backends/hif/hif-backend-vendor-mageia.c
+++ b/backends/hif/hif-backend-vendor-mageia.c
@@ -1,0 +1,67 @@
+/* -*- Mode: C; tab-width: 8; indent-tabs-mode: t; c-basic-offset: 8 -*-
+ *
+ * Copyright (C) 2016 Neal Gompa <ngompa13@gmail.com>
+ *
+ * Licensed under the GNU Lesser General Public License Version 2.1
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301 USA
+ */
+
+#include "hif-backend-vendor.h"
+
+gboolean
+hif_validate_supported_source(const gchar* id)
+{
+	guint i, j, k, l;
+
+	const gchar *valid_sourcesect[] = { "",
+					  "-tainted",
+					  "-nonfree",
+					  NULL };
+
+	const gchar *valid_sourcetype[] = { "",
+					  "-debuginfo",
+					  "-source",
+					  NULL };
+
+	const gchar *valid_arch[] = { "x86_64",
+				      "i586",
+				      "armv7hl",
+				      "armv5tl",
+				      NULL };
+
+	const gchar *valid[] = { "mageia",
+				 "updates",
+				 "updates_testing",
+				 "backports",
+				 "backports_testing",
+				 "cauldron",
+				 NULL };
+
+	/* Iterate over the ID arrays to find a matching identifier */
+	for (i = 0; valid[i] != NULL; i++) {
+		for (j = 0; valid_arch[j] != NULL; j++) {
+			for (k = 0; valid_sourcesect[k] != NULL; k++) {
+				for (l = 0; valid_sourcetype[l] != NULL; l++) {
+					g_autofree gchar *source_entry = g_strconcat(valid[i], "-", valid_arch[j], valid_sourcesect[k], valid_sourcetype[l], NULL);
+					if (g_strcmp0 (id, source_entry) == 0) {
+						return TRUE;
+					}
+				}
+			}
+		}
+	}
+	return FALSE;
+}
--- a/backends/hif/hif-backend-vendor.h
+++ b/backends/hif/hif-backend-vendor.h
@@ -1,0 +1,31 @@
+/* -*- Mode: C; tab-width: 8; indent-tabs-mode: t; c-basic-offset: 8 -*-
+ *
+ * Copyright (C) 2016 Neal Gompa <ngompa13@gmail.com>
+ *
+ * Licensed under the GNU Lesser General Public License Version 2.1
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301 USA
+ */
+
+#ifndef __HIF_BACKEND_VENDOR_H
+#define __HIF_BACKEND_VENDOR_H
+
+#include <glib.h>
+
+gboolean	hif_validate_supported_source(const gchar*);
+
+G_END_DECLS
+
+#endif /* __HIF_BACKEND_VENDOR_H */
--- a/backends/hif/pk-backend-hif.c
+++ b/backends/hif/pk-backend-hif.c
@@ -49,6 +49,7 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC(HifContext, g_object_unref)
 #include <hawkey/util.h>
 #include <librepo/librepo.h>
 
+#include "hif-backend-vendor.h"
 #include "hif-backend.h"
 
 typedef struct {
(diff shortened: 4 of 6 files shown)


backends/hif/pk-backend-hif.c
```diff
@@ -1171,30 +1172,20 @@ pk_backend_get_updates (PkBackend *backend,
...
-				 "updates-testing",
-				 "updates-testing-debuginfo",
-				 "updates-testing-source",
-				 NULL };
+
+	const gchar **valid = hif_get_supported_source_prefixes();
+
 	for (i = 0; valid[i] != NULL; i++) {
-		if (g_strcmp0 (id, valid[i]) == 0)
+		if (g_str_has_prefix (id, valid[i]) == TRUE)
```
kalev: A random drive-by comment: not sure it's a good idea to use g_str_has_prefix() to match repo names; it could very easily lead to matching something like "updates-rpmfusion" as a valid Fedora repo.

backends/hif/pk-backend-hif.c
```diff
@@ -186,7 +187,7 @@ pk_backend_setup_hif_context (HifContext *context, GKeyFile *conf, const gchar *
 	hif_context_set_cache_dir (context, cache_dir);
 	solv_dir = g_build_filename (destdir, "/var/cache/PackageKit", release_ver, "hawkey", NULL);
 	hif_context_set_solv_dir (context, solv_dir);
-	repo_dir = g_build_filename (destdir, "/etc/yum.repos.d", NULL);
+	repo_dir = g_build_filename (destdir, hif_get_reposdir(), NULL);
```
kalev: I am not sure how DNF does fallback between /etc/yum.repos.d/ and /etc/distro.repos.d/, but I suspect it may be necessary to look at _both_ directories, instead of just one or the other, in order to support 3rd party repos that may drop .repo files in either of the directories.

backends/hif/hif-backend-vendor-fedora.c
```diff
@@ -0,0 +1,41 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301 USA
+ */
+
+#include "hif-backend-vendor.h"
+#include <glib.h>
+
+const gchar *
+hif_get_reposdir()
+{
+	return g_strdup ("/etc/yum.repos.d");
```
ignatenkobrain: `const gchar*` & `g_stdrup ()` not compatible ;)

backends/hif/pk-backend-hif.c
Conan-Kudo: How do you suggest I handle this? I'm trying to avoid having a huge list of every single repository, especially because Mageia's approach to these things is to enable the full repository of every arch that can be used on a given architecture. So 64-bit x86 systems have both the 32-bit and the 64-bit repositories, most of the time.

backends/hif/pk-backend-hif.c
Conan-Kudo: That's not the point of this code or pull request. `hif_get_reposdir()` returns only a string. For Fedora, it returns `/etc/yum.repos.d`. For Mageia, it returns `/etc/distro.repos.d`. If the mechanism of retrieving reposdirs changes in PackageKit-Hif, then this function should change as well.

Conan-Kudo:
Rebased patch on master.

Conan-Kudo:
This pull request is somewhat dependent on rpm-software-management/dnf#498 and rpm-software-management/dnf#481.

[Fedora looks like it won't change from `/etc/yum.repos.d`](https://fedorahosted.org/fesco/ticket/1578), so the vendor configuration as-is remains valid.

hughsie:
So for the /etc/sys.repos.d bit, I just think we want to look at both paths on both distros. I don't think this needs any kind of configure step.

Conan-Kudo:
@hughsie, @kalev, @ignatenkobrain: I've reworked this to take out the repo configuration and it just handles the logic for validating a repository ID for Fedora or Mageia.

backends/hif/hif-backend-vendor-fedora.c
```diff
@@ -0,0 +1,47 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301 USA
+ */
+
+#include "hif-backend-vendor.h"
+#include <glib.h>
+
+static gboolean
+hif_validate_supported_source(const char* id)
+{
+	guint i;
```
ignatenkobrain: I would move this into for-loop if @hughsie don't mind

backends/hif/hif-backend-vendor-fedora.c
```diff
@@ -0,0 +1,47 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301 USA
+ */
+
+#include "hif-backend-vendor.h"
+#include <glib.h>
+
+static gboolean
+hif_validate_supported_source(const char* id)
```
ignatenkobrain: `const gchar *id`

backends/hif/hif-backend-vendor-mageia.c
```diff
@@ -0,0 +1,70 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301 USA
+ */
+
+#include "hif-backend-vendor.h"
+#include <glib.h>
+
+static gboolean
+hif_validate_supported_source(const char* id)
+{
+	guint i, j, k, l;
```
ignatenkobrain: same about c99 thing. do we really want such 4 loops? ;) I guess it could be simplified

backends/hif/hif-backend-vendor-mageia.c
```diff
@@ -0,0 +1,70 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301 USA
+ */
+
+#include "hif-backend-vendor.h"
+#include <glib.h>
```
ignatenkobrain: not needed

backends/hif/hif-backend-vendor-fedora.c
```diff
@@ -0,0 +1,47 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301 USA
+ */
+
+#include "hif-backend-vendor.h"
+#include <glib.h>
```
ignatenkobrain: not needed

backends/hif/hif-backend-vendor-fedora.c
Conan-Kudo: AFAIK, @hughsie follows C89/90 standard, so that means I can't initialize it in the for loop.

backends/hif/hif-backend-vendor-mageia.c
Conan-Kudo: I'm not sure it's possible to avoid it. These loops allow me to construct all the possible valid IDs used by official Mageia sources. The alternative is a *huge* list that is much harder to manage.

backends/hif/hif-backend-vendor-fedora.c
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301 USA
+ */
+
+#include "hif-backend-vendor.h"
+
+static gboolean
+hif_validate_supported_source(const gchar* id)
```
kalev: const gchar *id

backends/hif/hif-backend-vendor-fedora.c
```diff
@@ -0,0 +1,46 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301 USA
+ */
+
+#include "hif-backend-vendor.h"
+
+static gboolean
```
kalev: Does this actually work when the function is marked 'static', as in 'static' means that it should only be visible in a single compilation unit? I see it is called from another source file (another compilation unit).

backends/hif/hif-backend-vendor-mageia.c
```diff
@@ -0,0 +1,69 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301 USA
+ */
+
+#include "hif-backend-vendor.h"
+
+static gboolean
```
kalev: Same remark about static as above

backends/hif/hif-backend-vendor-mageia.c
```diff
@@ -0,0 +1,69 @@
...
+hif_validate_supported_source(const gchar* id)
```
kalev: const gchar *id

backends/hif/hif-backend-vendor-fedora.c
```diff
@@ -0,0 +1,46 @@
...
+				 "updates-debuginfo",
+				 "updates-source",
+				 "updates-testing",
+				 "updates-testing-debuginfo",
+				 "updates-testing-source",
+				 NULL };
+	for (i = 0; valid[i] != NULL; i++) {
+		if (g_strcmp0 (id, valid[i]) == 0)
+			return TRUE;
+	}
```
ignatenkobrain: no need for braces here

backends/hif/hif-backend-vendor-mageia.c
ignatenkobrain: `const gchar *id`

backends/hif/hif-backend-vendor-fedora.c
Conan-Kudo: Only one of them is included during the build process. So as far as I know, that means `static` is fine. You can either have `fedora` vendor or the `mageia` vendor, not both.

backends/hif/hif-backend-vendor-mageia.c
```diff
@@ -0,0 +1,69 @@
...
+				 "backports_testing",
+				 "cauldron",
+				 NULL };
+
+	/* Iterate over the ID arrays to find a matching identifier */
+	for (i = 0; valid[i] != NULL; i++) {
+		for (j = 0; valid_arch[j] != NULL; j++) {
+			for (k = 0; valid_sourcesect[k] != NULL; k++) {
+				for (l = 0; valid_sourcetype[l] != NULL; l++) {
+					gchar *source_entry = g_strconcat(valid[i], "-", valid_arch[j], valid_sourcesect[k], valid_sourcetype[l], NULL);
```
kalev: 'g_autofree gchar *source_entry' would be better, so that you wouldn't have to manually free it in both codepaths

backends/hif/hif-backend-vendor.h
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+#ifndef __HIF_BACKEND_VENDOR_H
+#define __HIF_BACKEND_VENDOR_H
+
+#include <glib.h>
+
+static gboolean	hif_validate_supported_source(const gchar*);
+
+G_END_DECLS
```
kalev: Missing matching G_BEGIN_DECLS

backends/hif/hif-backend-vendor.h
```diff
@@ -0,0 +1,31 @@
...
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301 USA
+ */
+
+#ifndef __HIF_BACKEND_VENDOR_H
+#define __HIF_BACKEND_VENDOR_H
+
+#include <glib.h>
+
+static gboolean	hif_validate_supported_source(const gchar*);
```
kalev: Again, shouldn't be 'static'.

Missing space before paren, missing tabs. Please use:

hif_validate_supported_source<tab>(const gchar<tab>*id);

backends/hif/pk-backend-hif.c
```diff
@@ -138,7 +139,7 @@ pk_backend_sack_cache_invalidate (PkBackend *backend, const gchar *why)
 static void
 pk_backend_hif_repos_changed_cb (HifRepos *self, PkBackend *backend)
 {
-	pk_backend_sack_cache_invalidate (backend, "yum.repos.d changed");
+	pk_backend_sack_cache_invalidate (backend, "repository configuration changed");
```
kalev: unrelated change

backends/hif/pk-backend-hif.c
```diff
@@ -1171,33 +1172,14 @@ pk_backend_get_updates (PkBackend *backend,
...
-				 "updates-testing",
-				 "updates-testing-debuginfo",
-				 "updates-testing-source",
-				 NULL };
-	for (i = 0; valid[i] != NULL; i++) {
-		if (g_strcmp0 (id, valid[i]) == 0)
-			return TRUE;
-	}
-	return FALSE;
+	return hif_validate_supported_source(hif_source_get_id (source));
```
kalev: missing paren before opening (

Conan-Kudo: What missing parens?

kalev: Sorry, missing space before opening paren

backends/hif/hif-backend-vendor-fedora.c
Conan-Kudo: That came from the original code, and I think it's part of his code style, so I don't think I should touch that.

Conan-Kudo:
Rebased against master and commits squashed.

PR #123, opened 2016-05-05
alpm :: Ignore Usage key in pacman.conf. Fixes #122

### LINKED ISSUE #122: Backend :: ALPM :: Add support for "Usage" key in config file
The Usage key was added to the default [config file](https://www.archlinux.org/pacman/pacman.conf.5.html#_repository_sections) in pacman v4.2. PackageKit currently crashes when the key is found in the config file. Pull request incoming....

lots0logs:
~~**DISCLAIMER:** I'm not entirely sure that I have done this correctly. It seems to work but of course that doesn't mean its correct. Please review it and let me know what changes are needed. Thanks.~~

### DIFF
--- a/backends/alpm/pk-alpm-config.c
+++ b/backends/alpm/pk-alpm-config.c
@@ -611,6 +611,11 @@ pk_alpm_config_parse (PkAlpmConfig *config, const gchar *filename,
 			continue;
 		}
 
+		if (g_strcmp0 (key, "Usage") == 0) {
+			/* Ignore the "Usage" key, alpm will handle it. */
+			continue;
+		}
+
 		/* report errors from above */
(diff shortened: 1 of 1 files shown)


aleixpol:
What happened here?

lots0logs:
@aleixpol I had hoped that simply ignoring the keys would resolve the issue (which I guess technically it does resolve part of it since it no longer crashes) However, I was wrong about the pk backend not needing to worry about repo usage setting. After further testing I see that packagekit will try to install updates from any repo it finds in pacman.conf regardless of the `Usage` key for the repo. I still have the work I did locally. I will push it to github so you can take a look. Maybe you can tell me what I am missing :grimacing:

lots0logs:
@aleixpol fe97a3bb6c9ba7eefd49c6606d9f2fcfa718869c

PR #126, opened 2016-05-23
Add Nix as a backend

matthewbauer:
I'm going to put this here while I'm still working on it to get an early review if you can. I've implemented the barebones PackageKit vfuncs for Nix but it's still pretty buggy. If anyone can correct any obvious errors or issues that would be helpful. Right now, I've just been using pkcon and not any PackageKit GUI, so I'll have to figure that out as well.

### DIFF
--- a/backends/Makefile.am
+++ b/backends/Makefile.am
@@ -52,6 +52,10 @@ if BACKEND_TYPE_PORTS
 SUBDIRS += ports
 endif
 
+if BACKEND_TYPE_NIX
+SUBDIRS += nix
+endif
+
 clean-local :
 	rm -f *~
 
--- a/backends/nix/Makefile.am
+++ b/backends/nix/Makefile.am
@@ -1,0 +1,13 @@
+plugindir = $(PK_PLUGIN_DIR)
+plugin_LTLIBRARIES = libpk_backend_nix.la
+
+libpk_backend_nix_la_SOURCES = \
+  pk-backend-nix.cc \
+  nix-helpers.cc \
+  nix-lib-plus.cc
+libpk_backend_nix_la_LIBADD = -lnixmain $(PK_PLUGIN_LIBS) $(NIX_LIBS)
+libpk_backend_nix_la_LDFLAGS = -module -avoid-version
+libpk_backend_nix_la_CFLAGS = $(PK_PLUGIN_CFLAGS) $(AM_CPPFLAGS)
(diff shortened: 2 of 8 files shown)


hughsie:
In concept looks good; I'm no C++ programmer but it looks a good start. If you want a real-world application to try, I'd suggest starting with gpk-application and gpk-update-viewer and then progress onto gnome-software. The former is quite demanding for status updates, and the latter needs requests to complete in lightning quick time. Yell if you need any sample AppStream data.

matthewbauer:
I'm moving this to ready for review because I'm able to get everything working properly in "pkcon" and "gkp-application". There may still be some issues going forward. I'm not sure what your policy for adding new backends are but it's certainly in working condition right now.

I'm going to try to be working on generating the appstream data for NixOS, but it's unclear how to work with the way Nix packages stuff.

ximion:
As long as the backend is maintained and working, we can have it in master - but talk to @hughsie about it, it might make sense to give you commit access to the repo so you can maintain the backend more easily.

> I'm going to try to be working on generating the appstream data for NixOS, but it's unclear how to work with the way Nix packages stuff.

Could you be more specific on that? Probably not the right place here though, a mail to appstream[at]lists.freedesktop.org is probably better (or a bug report against one of the AppStream metadata generators, in case you encounter a specific issue).

hughsie:
Agreed, I'd rather you just have commit access if you want to be the maintainer for the nix backend.

matthewbauer:
Pinging @edolstra, @domenkozar, for review of Nix related parts. Specifically, whether the "nix_get_all_derivations" and "nix_get_state" are properly working.

@ximion
Well eventually we are going to want to make an "appstream-builder" for the Nix build system that generates the appstream data. The issue right now is that there's not really a good way to tell between "apps" vs "non-apps" in Nix. In fact, there's not really an easy way to even tell what files are in a given derivation. Regardless, I think it's going to be easier getting PackageKit and related packages working first before tackling appstream.

matthewbauer:
One more thing for @hughsie:

How is the arch part of "package id" used? Right now the way Nix encodes the "system" it looks like "x86_64-linux" or "i686-linux" while most of the other backends seem to just show "i686" or "x86_64". Is that data used anywhere besides just for display?

hughsie:
@matthewbauer I think you can use the arch how you like.

matthewbauer:
Okay, so this should be ready to merge then.

hughsie:
I've given you commit access to master. Please push fixes to this backend only directly to master, and cherry pick them back into stable versions as appropriate. For any other fixes in the source tree please just grab a +1 from any of us. Thanks!

matthewbauer:
Okay thanks!

Issue #127, opened 2016-05-26
Resolve query with a missing package fails completely


aleixpol:
In _Discover_ we used to resolve appstream packages when it's loaded, to initialize whether they're installed or not. This turned out to be a problem when archlinux appstream db started offering an asset whose packagename wasn't there. I'm unsure if it's a design decision in PackageKit or an implementation error in the ALPM backend.

Here's how to test:
```
$ pkcon resolve vlc gcc
Resolving                     [=========================]         
Installed       gcc-6.1.1-1.x86_64 (installed)                                  The GNU Compiler Collection - C and C++ frontends
Installed       vlc-2.2.3-3.x86_64 (installed)                                  A multi-platform MPEG, VCD/DVD, and DivX player
$ pkcon resolve vlc gcc wrong
Resolving                     [=========================]         
Fatal error: wrong: could not find or read package
```

hughsie:
Looks like a backend problem to me:

$ pkcon resolve gcc wrong
Resolving                     [=========================]         
Installed   	gcc-6.1.1-2.fc24.x86_64 (installed:fedora)                  	Various compilers (C, C++, Objective-C, Java, ...)

the resolve method is perfect for finding out if a package exists.

PR #128, opened 2016-05-26
Don't discard a resolve call just because one package is missing

aleixpol:
Should solve issue #127

### DIFF
--- a/backends/alpm/pk-alpm-packages.c
+++ b/backends/alpm/pk-alpm-packages.c
@@ -184,7 +184,6 @@ pk_backend_resolve_thread (PkBackendJob *job, GVariant* params, gpointer p)
 {
 	const gchar **search;
 	PkBitfield filters;
-	g_autoptr(GError) error = NULL;
 
 	g_variant_get (params, "(t^a&s)", &filters, &search);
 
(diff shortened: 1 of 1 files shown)


hughsie:
is error being leaked?

ximion:
This looks like it will break if there are multiple errors (multiple packages missing) which causes a leak in GError.
You need to free-amd-null the error if it was already set before passing it, and create a new error containing both error messages which can be emitted in the end.

ximion:
```diff
diff --git a/backends/alpm/pk-alpm-packages.c b/backends/alpm/pk-alpm-packages.c
index b860e7b..591969e 100644
--- a/backends/alpm/pk-alpm-packages.c
+++ b/backends/alpm/pk-alpm-packages.c
@@ -189,16 +189,25 @@ pk_backend_resolve_thread (PkBackendJob *job, GVariant* params, gpointer p)
 	g_variant_get (params, "(t^a&s)", &filters, &search);
 
 	for (; *search != NULL; ++search) {
+		g_autoptr(GError) tmp_error = NULL;
+
 		if (pk_backend_job_is_cancelled (job))
 			break;
 
 		/* find a package with the given id or name */
 		if (pk_package_id_check (*search)) {
-			if (!pk_backend_resolve_package (job, *search, filters, &error))
-				break;
+			pk_backend_resolve_package (job, *search, filters, &tmp_error);
 		} else {
-			if (!pk_backend_resolve_name (job, *search, filters, &error))
-				break;
+			pk_backend_resolve_name (job, *search, filters, &tmp_error);
+		}
+
+		if (tmp_error != NULL) {
+			if (error == NULL) {
+				g_propagate_error (&error, tmp_error);
+				tmp_error = NULL;
+			} else {
+				g_prefix_error (&error, "%s\n", tmp_error->message);
+			}
 		}
 	}
```
It's really messy, because we swallow the code and domain of all subsequent errors, so emitting errors instead of making the transaction fail might be better.
But as an immediate fix, this isn't too bad (I don't know the alpm backend well enough, and have no way to test any code).

aleixpol:
I looked into it again, my conclusion was that on other platforms we aren't emitting the error (see @hughsie's reply to the issue, also I tested it on _KDE Neon_, apt-based).

So basically I changed it to just never emit an error. I hope this makes sense. :D

ximion:
Or only emit an error if there was an unfixable issue with the backend/alpm itself, but ignore the case when a package isn't present.
With dropping the error handling completely, you might loose issue information if there is any (I haven't checked for that).
IMHO that patch is fine though.

PR #129, opened 2016-05-30
hif: Use new libhif API to run a test transaction for offline updates

kalev:
Set the right flags so that libhif runs a test transaction after
we've finished downloading rpms for an offline update. This makes it
possible for us to error out before writing out the prepared offline
updated files and makes the offline update user experience nicer,
allowing us to catch transaction errors before rebooting.

https://github.com/rpm-software-management/libhif/pull/130 is the matching libhif change.


Conan-Kudo:
@kalev This looks like it's in the master branch already... Shouldn't this PR be closed. More importantly, why does GitHub think this isn't in master already?!

kalev:
Yup, it's in master indeed.

PR #132, opened 2016-06-15
zypp: PK_FILTER_ENUM_ARCH is supposed to filter on primary arch

DimStar77:


DimStar77:
@mlandres would be nice if you could confirm that this does not cause weird issues for openSUSE / SUSE Linux (but as far as I understand, this should have been the meaning of this filter)

mlandres:
Well, I can't tell whether  this will cause issues, as I do not know why and where this filter is used. 

However if the purpose of this filter is to return just packages exactly matching the systems architecture (e.g. x86_64) and not also compatible architectures (e.g. iX86,noarch), the patch is correct.

DimStar77:
@hughsie if you could please confirm the intention of this filter? Also: I failed to find a documentation if it was to filter pri arch or compat arch.

kalev:
AFAIK this should match both the main architecture (x86_64) + noarch and filter out the compat arch (i686)

DimStar77:
So this patch is no good - it only shows main architecture (no noarch)

kalev:
No, I mean, it should show noarch too

DimStar77:
right - which is why this patch is no good :)

kalev:
Ahh yes, right! :)

DimStar77:
ok - now 'arch' filter shows "noarch" and "system primary architecture"

Issue #134, opened 2016-06-28
cnf: zsh command_not_found_handler should return the same as its bash equivalent


timspeetjens:
The zsh command_not_found_handler does not have to return non-zero. In this respect, it behaves the same as its bash equivalent.

http://www.zsh.org/mla/workers/2007/msg00592.html and http://www.zsh.org/mla/workers/2007/msg00601.html

PR #135, opened 2016-06-28
cnf: zsh command_not_found_handler should return the same as its bash equivalent

timspeetjens:


hughsie:
3ad2c60fc9059998f314df33b59c97eb2f237782 seemed to suggest 127 was the right value...

timspeetjens:
The documentation (http://zsh.sourceforge.net/Doc/Release/Command-Execution.html) says it should return 0 if it successfully handled the command, or non-zero otherwise. In that case, the zsh itself returns 127 after printing a file not found message.

I think zsh should revise its behaviour, but that's beyond the scope of this issue / PR.

This PR fixes #134

timspeetjens:
As per http://www.zsh.org/mla/workers/2016/msg02049.html bash and zsh will behave equally, though the name difference will be maintained, starting from zsh 5.2

hughsie:
So this PR can be closed?

timspeetjens:
The referenced change in zsh does not change the issue or its solution.

PR #149, opened 2016-07-27
fix updates and change code to aid in not breaking this again

apachelogger:
Updates were broken in the previous local-file resolution change by eliminating it implicitly being handled in the resolution's else case. They are now once again handled as regular installs.

Additionally force expressiveness in the code by making the else case an error condition (couldn't do anything with the role) and explicitly handling local file in an else-if case.

### DIFF
--- a/backends/aptcc/pk-backend-aptcc.cpp
+++ b/backends/aptcc/pk-backend-aptcc.cpp
@@ -853,8 +853,13 @@ static void backend_manage_packages_thread(PkBackendJob *job, GVariant *params, 
             // Updates are like installs for the purposes of resolution, we
             // want to install the updates essentially.
             installPkgs = apt->resolvePackageIds(package_ids);
-        } else {
+        } else if (role == PK_ROLE_ENUM_INSTALL_FILES) {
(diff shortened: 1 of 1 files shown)


apachelogger:
@dantti ^

ximion:
LGTM

dantti:
@apachelogger The first is ok, the second I can't understand it, there is no other role that will call that function so it's a fix for a possible programmer mistake which is useless to me.

apachelogger:
There is no other role **right now**.

Worst case the code path is never hit. Best case the logic there doesn't get broken again by implicitly handling stuff in the else, which as you put it is a "fix for a possible programmer mistake", which has value to you because you literally made that mistake.
For me it has value too because I don't want to spend another 3 hours of my life hunting down an issue that claims to be related to "can't find package" but really is "shitty-if detected".

aleixpol:
+1, please let's get this one in.

dantti:
Ok but lets use an assert then

dantti:
I have some furthers optimizations and refactoring to do so merging this to have it fixed now, thanks

PR #166, opened 2016-09-20
aptcc: make toinstall more apt compliant

### LINKED ISSUE #145: aptcc not installing recommends
- _foobar_ 1.0 doesn't recommend _kitteh_
- _foobar_ 2.0 recommends _kitteh_
- pkcon update from 1.0 to 2.0 doesn't want to install _kitteh_

This is inconsistent with apt where both upgrade and dist-upgrade will install new recommends. Furthermore, this actually can have negative impact on any form of rolling system where this sort of thing might occur regularly.

### LINKED ISSUE #151: [aptcc] updates always installs with fromuser=true
when updating a package it runs through
`AptCacheFile::tryToInstall`
which calls
`    GetDepCache()->MarkInstall(Pkg, false);`
which ultimately has this signature
```
   bool MarkInstall(PkgIterator const &Pkg,bool AutoInst = true,
		    unsigned long Depth = 0, bool FromUser = true,
		    bool ForceImportantDeps = false);
```
which IIRC means that every package ever looses its `pkgCache::Flag::Auto` which means apt won't suggest it be removed if all it's dependees were removed. Namely aptcc defuncts the entire `pkgCache::Flag::Auto` use-case

Should be something like
```
/***************************************************************************
 *   Copyright © 2010-2012 Jonathan Thomas <echidnaman@kubuntu.org>        *
 *                                                                         *
 *   This program is free software; you can redistribute it and/or         *
 *   modify it under the terms of the GNU General Public License as        *
 *   published by the Free Software Foundation; either version 2 of        *
 *   the License or (at your option) version 3 or any later version        *
 *   accepted by the membership of KDE e.V. (or its successor approved     *
 *   by the membership of KDE e.V.), which shall act as a proxy            *
 *   defined in Section 14 of version 3 of the license.                    *
 *                                                                         *
 *   This program is distributed in the hope that it will be useful,       *
 *   but WITHOUT ANY WARRANTY; without even the implied warranty of        *
 *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         *
 *   GNU General Public License for more details.                          *
 *                                                                         *
 *   You should have received a copy of the GNU General Public License     *
 *   along with this program.  If not, see <http://www.gnu.org/licenses/>. *
 ***************************************************************************/
            bool fromUser = !(State.Flags & pkgCache::Flag::Auto);
            (*m_cache)->MarkInstall(iter, true, 0, fromUser);
```

### LINKED ISSUE #153: [aptcc] updates never get autoinst resolution
updates call `AptCacheFile::tryToInstall` which calls `GetDepCache()->MarkInstall(Pkg, false);` which never ever manages to resolve involved problems since this explicitly sets autoinst=false BUT then never lets the resolver attempt to resolve this by setting autoinst. Notably what happens is that apt-pkg will then have all packages marked for install, but not be able to remove (and probably install new) packages which would be in conflict with the now strictly marked packages.

Example:

- A installed, A depends B
- B installed, new B available, conflicts C
- C is installed and has an upgrade
- Updating A, B, and C
 - A marked, B marked, C marked
 - A gets resolved by installing B, C can't be resolved because we explicitly want C and explicitly want A and explicitly want B but A through B conflicts C, B directly conflicts C. We can't resolve this by removing C because C was explicitly marked as noautoinst.

What should happen is:

- run through intended changes ⇒ mark everything install autoinst=false
- run through intended changes AGAIN ⇒ mark everything install autoinst=true
- at this point packages will be marked for remove ⇒ depending on policy mark removed packages as keep to prevent them from actually getting removed (which I think is actually incompatible with packagekit api, so this behavior probably makes no sense)
 - resolveByKeep (as mentioned, I think upgrade-by-not-removing-anything isn't really a thing that should be supported)

apachelogger:
new arguments: autoInst and preserveAuto

apt's markInstall can automatically try to resolve dependencies (incl.
recommends if applicable) without having the resolver deal with broken
dependencies at all. to that end it will mark dependencies of $package as
auto-install which gives them a slightly lower resolution score and the
auto flag (so they can be autoremoved).
To facilitate its use, the interface now calls toinstall twice: once
without autoInst to only mark explicitly wanted packages, and a second time
with autoInst to then also mark all needed dependencies. This is
specifically meant to allow or-relationships to resolve as desired.

e.g.
```
Package: foobar
Depends: foo|bar
```
with `pkcon install foobar bar` will explicitly mark foobar and bar and
thus prevent foo from getting autoinstalled.

In update scenarios it is hard to tell whether a user wants a package to be
installed or updated. In particular whether or not a package is meant to
be installed according to the user would prevent them from being
autoremoved. A user may however run `pkcon update foo` to update foo, not
because she actually thinks foo is needed or useful. To prevent
unintentionally losing the auto flag on updates, the backend now
always preserves the flag when updates are requested. Conversely on
installs we default to not having the auto flag.

fixes #153
fixes #151
fixes #145

### DIFF
--- a/backends/aptcc/apt-intf.cpp
+++ b/backends/aptcc/apt-intf.cpp
@@ -359,12 +359,14 @@ PkgList AptIntf::filterPackages(const PkgList &packages, PkBitfield filters)
             pkgProblemResolver Fix(*m_cache);
             {
                 pkgDepCache::ActionGroup group(*m_cache);
-                for (const pkgCache::VerIterator &ver : ret) {
-                    if (m_cancel) {
-                        break;
-                    }
+                for (auto autoInst : { true, false }) {
+                    for (const pkgCache::VerIterator &ver : ret) {
+                        if (m_cancel) {
+                            break;
+                        }
 
-                    m_cache->tryToInstall(Fix, ver, false);
+                        m_cache->tryToInstall(Fix, ver, false, autoInst, false);
+                    }
                 }
             }
 
@@ -2176,26 +2178,28 @@ bool AptIntf::runTransaction(const PkgList &install, const PkgList &remove, cons
 
     pkgProblemResolver Fix(*m_cache);
 
+    // TODO: could use std::bind an have a generic operation array iff toRemove had the same
+    //       signature
+
+    struct Operation {
+        const PkgList &list;
+        const bool preserveAuto;
+    };
+
     // new scope for the ActionGroup
     {
         pkgDepCache::ActionGroup group(*m_cache);
-        for (const pkgCache::VerIterator &verIt : install) {
-            if (m_cancel) {
-                break;
-            }
 
-            if (!m_cache->tryToInstall(Fix, verIt, BrokenFix)) {
-                return false;
-            }
-        }
-
-        for (const pkgCache::VerIterator &verIt : update) {
-            if (m_cancel) {
-                break;
-            }
-
-            if (!m_cache->tryToInstall(Fix, verIt, BrokenFix)) {
-                return false;
+        for (auto op : { Operation { install, false }, Operation { update, true } }) {
+            for (auto autoInst : { false, true }) {
+                for (const pkgCache::VerIterator &verIt : op.list) {
+                    if (m_cancel) {
+                        break;
+                    }
+                    if (!m_cache->tryToInstall(Fix, verIt, BrokenFix, autoInst, op.preserveAuto)) {
+                        return false;
+                    }
+                }
             }
         }
 
--- a/backends/aptcc/apt-intf.cpp
+++ b/backends/aptcc/apt-intf.cpp
@@ -352,7 +352,7 @@ PkgList AptIntf::filterPackages(const PkgList &packages, PkBitfield filters)
             }
         }
 
-        // This filter is more complex so we filter it after the list has shrink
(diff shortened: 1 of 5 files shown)


backends/aptcc/apt-intf.cpp
```diff
@@ -2157,7 +2160,8 @@ PkgList AptIntf::resolveLocalFiles(gchar **localDebs)
     return ret;
 }
 
-bool AptIntf::runTransaction(const PkgList &install, const PkgList &remove, bool fixBroken, PkBitfield flags, bool autoremove)
+bool AptIntf::runTransaction(const PkgList &install, const PkgList &remove, const PkgList &update,
+                             bool fixBroken, PkBitfield flags, bool autoremove)
 {
     //cout << "runTransaction" << simulate << remove << endl;
```
ximion: Urg, debugging by commenting out...
Not an issue in this PR, it just caught my attention now...

ximion:
Wow :) Zero objections from me.

Just one hint: If you make trivial commits like removing a type, prefix them with "trivial", e.g. `trivial: typo-- in aptcc` - that way, they will not show up as clutter in the release news.

ximion:
I will add this patch to today's revision of the `packagekit` Debian package.

PR #174, opened 2016-11-03
Get backends warnings on the syslog(), makes it visible on journalctl

aleixpol:
Should allow more efficient issue reporting.
Should help detecting regressions when the backend behaviors change.
Otherwise, the output we generate goes to /dev/null, which is possibly
fine for g_debug but in this case we're not being warned.

### DIFF
--- a/backends/alpm/pk-alpm-config.c
+++ b/backends/alpm/pk-alpm-config.c
@@ -26,6 +26,7 @@
 #include <string.h>
 #include <sys/utsname.h>
 #include <glib/gstdio.h>
+#include <syslog.h>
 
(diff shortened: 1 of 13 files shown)


aleixpol:
Hi, if you don't want it this way on all backends, I can change it to only happen on alpm, would that be better?

aleixpol:
Will push the change for alpm, as there doesn't seem to be interest in others.

ximion:
This will solve itself anyway with the new logging facilities which are scheduled to be landing in GLib and make logging to the journal really easy.
As soon as that stuff is ready, we don't need special support for logging into PK anymore.

PR #176, opened 2016-11-21
Fix redeclaration of polkit autocleanup functions

pwithnall:
ifdef cannot be used to determine if a C symbol is defined, as it’s
evaluated by the preprocessor, before symbols are parsed. Instead of
trying to detect whether polkit.h is suitably recent enough to define
its own auto-cleanup functions, just define our own in a different
namespace which is guaranteed not to collide with upstream.


hughsie:
I'd much rather a configure.ac check TBH.

pwithnall:
Updated to do that.

Issue #159, opened 2016-08-23
[aptcc] Update transaction progress regresses over time


aleixpol:
Just as described in issue #70 for ALPM.

See /3847_dadeacce goes from 96 -> -1 -> 0.

```
$ pkmon | grep percentage
/3846_baddbcee	percentage   50
/3846_baddbcee	percentage   100
/3847_dadeacce	percentage   -1
/3847_dadeacce	percentage   0
/3847_dadeacce	percentage   100
/3847_dadeacce	percentage   0
/3847_dadeacce	percentage   50
/3847_dadeacce	percentage   100
/3847_dadeacce	percentage   -1
/3847_dadeacce	percentage   0
/3847_dadeacce	percentage   1
/3847_dadeacce	percentage   5
/3847_dadeacce	percentage   11
/3847_dadeacce	percentage   16
/3847_dadeacce	percentage   22
/3847_dadeacce	percentage   27
/3847_dadeacce	percentage   29
/3847_dadeacce	percentage   34
/3847_dadeacce	percentage   39
/3847_dadeacce	percentage   46
/3847_dadeacce	percentage   53
/3847_dadeacce	percentage   55
/3847_dadeacce	percentage   56
/3847_dadeacce	percentage   57
/3847_dadeacce	percentage   58
/3847_dadeacce	percentage   59
/3847_dadeacce	percentage   60
/3847_dadeacce	percentage   61
/3847_dadeacce	percentage   62
/3847_dadeacce	percentage   63
/3847_dadeacce	percentage   68
/3847_dadeacce	percentage   74
/3847_dadeacce	percentage   81
/3847_dadeacce	percentage   89
/3847_dadeacce	percentage   96
/3847_dadeacce	percentage   -1
/3847_dadeacce	percentage   0
/3847_dadeacce	percentage   1
/3847_dadeacce	percentage   2
/3847_dadeacce	percentage   3
/3847_dadeacce	percentage   4
/3847_dadeacce	percentage   5
/3847_dadeacce	percentage   6
/3847_dadeacce	percentage   7
/3847_dadeacce	percentage   8
/3847_dadeacce	percentage   9
/3847_dadeacce	percentage   10
/3847_dadeacce	percentage   11
/3847_dadeacce	percentage   12
/3847_dadeacce	percentage   13
/3847_dadeacce	percentage   14
/3847_dadeacce	percentage   15
/3847_dadeacce	percentage   16
/3847_dadeacce	percentage   17
[...]

PR #181, opened 2017-03-14
[aptcc] Specify status when calling pk_backend_job_set_item_progress

aleixpol:
At the moment we always returned PK_STATUS_ENUM_UNKNOWN and it was hard
to keep track of what was happening.

Should improve issue #159.


aleixpol:
Before merging, please test. I made this change on a non-debian system.

dantti:
Oh I see backend API changed to have a per item status, lgtm, though I didn't tested.

ximion:
Change looks good to me, +1

PR #182, opened 2017-03-14
First approach to a CI

aleixpol:
Compiles the project on debian stretch and fedora

See results: https://travis-ci.org/aleixpol/PackageKit (will have to enable it on hugsie after merging)

### DIFF
--- a/tests/ci/Dockerfile-fedora
+++ b/tests/ci/Dockerfile-fedora
@@ -1,0 +1,8 @@
+FROM fedora:25
+
+RUN dnf -y update
+RUN dnf -y install dnf-plugins-core libdnf-devel redhat-rpm-config autoconf-archive
+RUN dnf -y builddep PackageKit
+
+RUN mkdir /build
+WORKDIR /build
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,14 @@
+language: c
+sudo: required
+dist: trusty
+
+services:
+  - docker
+
+before_script:
+  - docker build -t packagekit-debian -f tests/ci/Dockerfile-debian .
+  - docker build -t packagekit-fedora -f tests/ci/Dockerfile-fedora .
+
+script:
+  - docker run -t -v `pwd`:/build packagekit-debian ./tests/ci/build_and_test.sh --enable-aptcc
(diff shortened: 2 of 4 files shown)


tests/ci/Dockerfile-fedora
```diff
@@ -0,0 +1,9 @@
+FROM fedora
+
+RUN dnf -y update
+RUN dnf -y install dnf-plugins-core
+RUN dnf -y builddep PackageKit
+RUN dnf -y install libdnf-devel redhat-rpm-config autoconf-archive
+
```
Conan-Kudo: I'd merge lines 4 and 6 together, and also add `autoconf automake libtool` to that line.

tests/ci/Dockerfile-fedora
```diff
@@ -0,0 +1,9 @@
+FROM fedora
```
Conan-Kudo: `FROM fedora:latest`

tests/ci/Dockerfile-fedora
Conan-Kudo: Also, you can make them all quiet by adding `-q` to `-y`.

tests/ci/Dockerfile-fedora
aleixpol: I wonder if we want to specify the version like on debian...

Conan-Kudo: Id you'd want to do that, maybe set `25`.

hughsie:
Looks great! How do we set up the "hook" to make travis re-run the tests on new checkins?

ximion:
@hughsie You need to select the Travis module (and register at Travis) in Github's repo configuration. The rest is done automatically by GH and Travis.

aleixpol:
Ping

hughsie:
Sorry about this, currently at a conference. I like the idea, we just have to make it work! :)

ximion:
This is also just building stuff at time, and not yet running tests - running the unit-tests would be pretty awesome, but in case of PK it's also an incredibly difficult task (esp. for the performance tests).

hughsie:
Okay, merge done. I guess setting up travis is next, right?

PR #184, opened 2017-04-13
pk-progress-bar: checks NULL strings inside pk_progress_bar_start ().

jcfaracco:
This commit adds a conditional to check if the string of progress bar
and text input are NULL. The comparison needs to be executed with two
valid strings and not two NULL pointes. That's why the conditional has
an AND operator. Both needs to be an empty string or a valid string, not
a NULL pointer. This missing check is causing an error during the pkcon
execution. See the command output:

    $ sudo pkcon repair
    FIXME: need to call pk_progress_bar_start()
    earlier![=========================]
                                  [=========================]

After patching:

    $ sudo ./pkcon repair
                                  [=========================]
                                  [=========================]

### DIFF
--- a/lib/packagekit-glib2/pk-progress-bar.c
+++ b/lib/packagekit-glib2/pk-progress-bar.c
@@ -268,8 +268,9 @@ pk_progress_bar_start (PkProgressBar *self, const gchar *text)
 	g_return_val_if_fail (PK_IS_PROGRESS_BAR (self), FALSE);
 
 	/* same as last time */
-	if (g_strcmp0 (self->priv->old_start_text, text) == 0)
-		return TRUE;
+	if (self->priv->old_start_text != NULL && text != NULL)
+		if (g_strcmp0 (self->priv->old_start_text, text) == 0)
+			return TRUE;
 	g_free (self->priv->old_start_text);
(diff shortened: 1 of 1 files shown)


lib/packagekit-glib2/pk-progress-bar.c
```diff
@@ -268,8 +268,9 @@ pk_progress_bar_start (PkProgressBar *self, const gchar *text)
 	g_return_val_if_fail (PK_IS_PROGRESS_BAR (self), FALSE);
 
 	/* same as last time */
-	if (g_strcmp0 (self->priv->old_start_text, text) == 0)
-		return TRUE;
+	if (self->priv->old_start_text != NULL && text != NULL)
+		if (g_strcmp0 (self->priv->old_start_text, text) == 0)
```
hughsie: Can the "outer" `if` please have `{}` to make it clear please. The inner one is fine as it is.

jcfaracco: Great! I will close the PR and resend a V2 PR with the fixes. Thanks!

jcfaracco:
Closing the PR to apply the suggested changes.

PR #186, opened 2017-04-19
pk-main: segfault when config file is not found.

jcfaracco:

See for more details:

conf_filename is NULL and inside that function the `assert()` fails:

    $ sudo packagekitd

    Segmentation fault

This commits fixes this problem with a NULL check and change some 
`g_print()` functions to `g_error()`.

### DIFF
--- a/src/pk-main.c
+++ b/src/pk-main.c
@@ -173,10 +173,14 @@ main (int argc, char *argv[])
 	/* get values from the config file */
 	conf = g_key_file_new ();
 	conf_filename = pk_util_get_config_filename ();
+	if (conf_filename == NULL) {
+		g_error ("Cannot found config file.");
+		goto out;
+	}
 	ret = g_key_file_load_from_file (conf, conf_filename,
 					 G_KEY_FILE_NONE, &error);
 	if (!ret) {
-		g_print ("Failed to load config file: %s", error->message);
+		g_error ("Failed to load config file: %s", error->message);
 		goto out;
 	}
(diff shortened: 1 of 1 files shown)


src/pk-main.c
```diff
@@ -173,10 +173,14 @@ main (int argc, char *argv[])
 	/* get values from the config file */
 	conf = g_key_file_new ();
 	conf_filename = pk_util_get_config_filename ();
+	if (conf_filename == NULL) {
+		g_error ("Cannot found config file.");
```
jcfaracco: I don't know if we need this g_error() here since we have the g_warning() inside the pk_util_get_config_filename() function, but the text is not appearing to me by default.

hughsie: g_error is fatal and won't do any of the cleanup. It just needs to be a `g_print` and then `goto out` kind of cleanup.

ximion: ...or a `g_printerr` :P

jcfaracco: Yes. Well, we all agree that `g_error()` is not right.
Basically, `g_printerr()` -> stderr and `g_print()` -> stdout.
In practice, I agree. Maybe if someone would like to redirect part of the output (only errors).

hughsie: I agree, `g_printerr` is what we want here.

jcfaracco:
Closing this PR and sending V2 without `g_print()` changes.
Fix the segfault is the main objective here.

PR #188, opened 2017-04-21
pk-main: segfault when config file is not found [v2].

jcfaracco:

See:

`conf_filename` is NULL and inside that function the `assert()` fails:

    $ sudo packagekitd

    Segmentation fault

This commit fixes the problem.

### DIFF
--- a/src/pk-main.c
+++ b/src/pk-main.c
@@ -173,6 +173,10 @@ main (int argc, char *argv[])
 	/* get values from the config file */
 	conf = g_key_file_new ();
 	conf_filename = pk_util_get_config_filename ();
+	if (conf_filename == NULL) {
+		g_print ("Cannot found config file.");
+		goto out;
+	}
 	ret = g_key_file_load_from_file (conf, conf_filename,
 					 G_KEY_FILE_NONE, &error);
 	if (!ret) {


src/pk-main.c
```diff
@@ -173,6 +173,10 @@ main (int argc, char *argv[])
 	/* get values from the config file */
 	conf = g_key_file_new ();
 	conf_filename = pk_util_get_config_filename ();
+	if (conf_filename == NULL) {
+		g_print ("Cannot found config file.");
```
ximion: Really really minor, but `Cannot found config file.` is weird English - maybe `Could not find config file.` or `Config file not found.` would be better here.
Or, IMHO even better and by using `g_printerr`:
```C
g_printerr ("Config file '%s' not found.", conf_filename);
```

Additionally, but unrelated to this PR: @hughsie, shouldn't we exit with a non-zero exit status if we encounter such initialization issues? At time, PK seems to always terminate with an exit code of zero.

ximion:
Btw, I think you can just force-push to your repo, the bug that Github forgets the comments made on PRs is gone, AFAIK, and the existing PR will just update in that case.

src/pk-main.c
jcfaracco: Ok, let me change the message. ;-)
I didn't change to `g_printerr()` now to follow the current pattern inside pk-main.c.
I totally agree, but maybe it can be introduced by another patch. There are many occurrences of others `g_print()` for errors in other files either.

jcfaracco:
I create a v3 PR with the recommendations. For reference #190 
Probably, other files will need to be checked if they are using `g_print()` for errors too.

Conan-Kudo:
Can we close this PR, since I think v3 was merged?

ximion:
Jep

PR #190, opened 2017-05-02
pk-main: segfault when config file is not found [v3].

jcfaracco:

See:

`conf_filename` is NULL and inside that function the `assert()` fails:

    $ sudo packagekitd
    Segmentation fault

This PR introduces `g_printerr()` to redirect properly errors from pk-main.c to 
stderr.

### DIFF
--- a/src/pk-main.c
+++ b/src/pk-main.c
@@ -173,6 +173,10 @@ main (int argc, char *argv[])
 	/* get values from the config file */
 	conf = g_key_file_new ();
 	conf_filename = pk_util_get_config_filename ();
+	if (conf_filename == NULL) {
+		g_print ("Config file was not found.");
+		goto out;
+	}
 	ret = g_key_file_load_from_file (conf, conf_filename,
 					 G_KEY_FILE_NONE, &error);
 	if (!ret) {
--- a/src/pk-main.c
+++ b/src/pk-main.c
@@ -174,13 +174,13 @@ main (int argc, char *argv[])
 	conf = g_key_file_new ();
 	conf_filename = pk_util_get_config_filename ();
 	if (conf_filename == NULL) {
-		g_print ("Config file was not found.");
+		g_printerr ("Config file was not found.");
 		goto out;
 	}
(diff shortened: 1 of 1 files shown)


ximion:
Looks good to me, +1

I do wonder though why all of these print statements lack a terminating newline character - I guess this codepath is triggered to rarely (chance to hit it is pretty much zero on all Linux distros unless one purposefully broke it) that nobody noticed.
Anyway, this looks good to go, IMO.

jcfaracco:
Hi guys,
For documentation, here is the steps using the PackageKit installed by my Linux repos (not compiled from sources):
**1. `$ systemctl status packagekit.service`**
```
   Active: active (running) since Tue 2017-05-02 09:34:23 -03; 31min ago
 Main PID: 2224 (packagekitd)
  CGroup: /system.slice/packagekit.service
           └─2224 /usr/libexec/packagekitd
```
**2. `$ sudo mv /etc/PackageKit/PackageKit.conf /etc/PackageKit/PackageKit.conf.old`**
**3. `$ sudo systemctl restart packagekit.service`**
```
Job for packagekit.service failed because a fatal signal was delivered to the control process. See "systemctl status packagekit.service" and "journalctl -xe" for details.
```
**4. `$ sudo systemctl status packagekit.service`**
```
   Active: failed (Result: signal) since Tue 2017-05-02 10:06:12 -03; 5min ago
  Process: 21205 ExecStart=/usr/libexec/packagekitd (code=killed, signal=SEGV)
 Main PID: 21205 (code=killed, signal=SEGV)

systemd[1]: Starting PackageKit Daemon...
packagekitd[21205]: (packagekitd:21205): GLib-CRITICAL **: g_key_file_load_from_file: assertion 'file != NULL' failed
systemd[1]: packagekit.service: main process exited, code=killed, status=11/SEGV
systemd[1]: Failed to start PackageKit Daemon.
systemd[1]: Unit packagekit.service entered failed state.
systemd[1]: packagekit.service failed.
```

PR #191, opened 2017-05-15
Alpm hooks

eworm-de:


eworm-de:
Here are some infos about pacman hooks:
https://www.archlinux.org/pacman/alpm-hooks.5.html

This is used to regenerate kernel initial ramdisk, update mime database, write font cache, ... and a lot more.

Some of these hooks are time consuming, thus user may think packagekit hangs after 100% package installation. With these changes packagekit changes state to "Running hooks".

hughsie:
I think if it's hanging at 100% then you need to rework how the progress is updated to include the time spent running the hooks. It's quite bad UX to have a progress bar rush to 99% and then sit there for minutes before it's installed. I don't think "Running Hooks" tells me anything at all, sorry.

eworm-de:
Currently the packagekit action looks like this:
```
$ pkcon update
Getting updates               [=========================]         
Downloading list of packages  [=========================]         
Updating packages             [=========================]         
Downloading packages          [=========================]
Checking signatures           [=========================]
Starting                      [=========================]
Installing updates            [=========================]
```
... where
```
Installing updates            [=========================]
```
shows for a long time with 100% if hooks are run.

Instead I would expect it to show information about hooks:
```
[...]
Installing updates            [=========================]
Running hooks                 [=========================]
```

Hooks do not have any progress, that is given by libalpm/pacman. But `Running hooks` at 0% is better than `Installing updates` at 100%. Possibly there is still room for improvement on libalpm side. ;)

Probably every Arch user should know what hooks do. So I do not see a real problem here.

hughsie:
> Hooks do not have any progress

I think that's the bit that needs fixing, especially if they are going to take more than 0.5s.

eworm-de:
Well, every hook runs a single command, so you can not tell any percentage there.

Running several hooks I could calculate a progress myself... So for four hooks the progress would increase by 25% for every hook finished. Would you accept that?

eworm-de:
Updated my pull request. How about this?

hughsie:
I think progress like that is certainly better than nothing. I think what the "install" phase should probably do is go from 0% to 90% (i.e. you scale the existing percentage by a factor of 0.9) and then running hooks you want to scale it by 0.1 to make 0% to 10% with an offset of 90%. That way you get an "installing" progressbar that doesn't jump back to 0%, which isn't going to look great on the command line or in gnome-software.

eworm-de:
There is updating package list, downloading signature, downloading packages, verifying sources, testing changes, ... and more. The percentage jumps back to 0% for each of them. So I do not see the point in merging installation and running hooks.

The log for my current code looks like this:
```
May 23 21:24:36 nepu PackageKit[3250]: Hook gdk-pixbuf-query-loaders.hook (Probing GDK-Pixbuf loader modules...) complete (1 of 16)
May 23 21:24:36 nepu PackageKit[3250]: Hook gio-querymodules.hook (Updating GIO module cache...) complete (2 of 16)
May 23 21:24:36 nepu PackageKit[3250]: Hook glib-compile-schemas.hook (Compiling GSettings XML schema files...) complete (3 of 16)
May 23 21:24:36 nepu PackageKit[3250]: Hook gtk-query-immodules-2.0.hook (Probing GTK2 input method modules...) complete (4 of 16)
May 23 21:24:36 nepu PackageKit[3250]: Hook gtk-query-immodules-3.0.hook (Probing GTK3 input method modules...) complete (5 of 16)
May 23 21:24:36 nepu PackageKit[3250]: Hook gtk-update-icon-cache.hook (Updating icon theme caches...) complete (6 of 16)
May 23 21:24:36 nepu PackageKit[3250]: Hook gvfsd.hook (Reloading GVFS config...) complete (7 of 16)
May 23 21:24:37 nepu PackageKit[3250]: Hook linux.hook (Updating linux initcpios) complete (8 of 16)
May 23 21:24:37 nepu PackageKit[3250]: Hook systemd-hwdb.hook (Updating udev hardware database...) complete (9 of 16)
May 23 21:24:37 nepu PackageKit[3250]: Hook systemd-sysusers.hook (Updating system user accounts...) complete (10 of 16)
May 23 21:24:37 nepu PackageKit[3250]: Hook systemd-tmpfiles.hook (Creating temporary files...) complete (11 of 16)
May 23 21:24:37 nepu PackageKit[3250]: Hook systemd-update.hook (Arming ConditionNeedsUpdate...) complete (12 of 16)
May 23 21:24:38 nepu PackageKit[3250]: Hook texinfo-install.hook (Updating the info directory file...) complete (13 of 16)
May 23 21:24:38 nepu PackageKit[3250]: Hook update-ca-trust.hook (Rebuilding certificate stores...) complete (14 of 16)
May 23 21:24:38 nepu PackageKit[3250]: Hook update-desktop-database.hook (Updating the desktop file MIME type cache...) complete (15 of 16)
May 23 21:24:39 nepu PackageKit[3250]: Hook update-mime-database.hook (Updating the MIME type database...) complete (16 of 16)
```
This is a pretty fast system, though. Only slower systems with large initial ramdisk running the linux hook first will have 0% for a longer time.

PR #193, opened 2017-05-18
aptcc: Fix installing gst codecs for gstreamer 1.0

### LINKED ISSUE #192: aptcc: WhatProvides gstreamer always returns 0 results for 1.0 codecs
Try

    pkcon what-provides 'gstreamer1.0(decoder-audio/mpeg)(mpegversion=(int)1)(mpegaudioversion=(int)1)(layer=(int)3)'

It always returns no results on Debian and Ubuntu. I've got https://paste.debian.net/932844/ so far, but it needs some refining to be PRable. Just filing this for the record.

iainlane:
The regex didn't work with gstreamer 1.0 input.

Try

  pkcon what-provides 'gstreamer1.0(decoder-audio/mpeg)(mpegaudioversion=1)(mpegversion=1)(layer=3)'

to test.

Also, the second commit fixes a critical and no results that happen for the non-first queries. You can't deinit and then init again - it breaks gstreamer.

@ximion Please test this. I don't know much about this stuff. I think there's a way to request multiple codecs at once - how do you do that? I don't know and so I didn't test it. :)

Fixes #192.

### DIFF
--- a/backends/aptcc/gst-matcher.cpp
+++ b/backends/aptcc/gst-matcher.cpp
@@ -23,9 +23,14 @@
 #include <regex.h>
 #include <gst/gst.h>
 
+static bool inited = false;
+
 GstMatcher::GstMatcher(gchar **values)
 {
-    gst_init(NULL, NULL);
+    if (!inited) {
+        gst_init(NULL, NULL);
+        inited = true;
+    }
 
     // The search term from PackageKit daemon:
     // gstreamer0.10(urisource-foobar)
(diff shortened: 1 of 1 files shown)


dantti:
LGTM

iainlane:
@ximion want to review/merge?

ximion:
The awful regex still makes me cringe, but in general this patch looks good to me :-)
In the long run we might want to rework this a bit more, for example by using GRegex and only compiling the regex once, etc.
But for now, +1 from me too, and since @dantti is fine with it as well, I'll just merge this now :-)

Thanks Iain for fixing this, and I am still puzzled that nobody noticed this was broken for such a long time.

PR #200, opened 2017-06-30
aptcc: Don't return installed packages for codec search

iainlane:
It doesn't seem possible to perform this search properly with the
information we have. In particular we almost always say that '-good' is
a candidate when it might not be. '-good' is almost always installed, so
it is good enough to filter out installed packages.

The argument as to why this is okay to do is that if the user is being
shown the dialog then it's because they don't have a codec installed, so
it's not that helpful to show installed packages.

@ximion another for you - is this a good idea? I looked for a while and I can't see how to filter out -good using caps.

### DIFF
--- a/backends/aptcc/apt-intf.cpp
+++ b/backends/aptcc/apt-intf.cpp
@@ -537,6 +537,14 @@ void AptIntf::providesCodec(PkgList &output, gchar **values)
             continue;
         }
 
+        // Ignore installed packages - we're being told that we don't have the
+        // codec, so installed packages can't satisfy what we're after. In
+        // particular, the caps syntax currently doesn't allow us to filter out
+        // -good when requesting packages for MP3s.
+        if (pkg->CurrentState == pkgCache::State::Installed) {
+            continue;
+        }
+
         // TODO search in updates packages
         // Ignore virtual packages
         pkgCache::VerIterator ver = m_cache->findVer(pkg);


iainlane:
oh and it's what sessioninstaller does

ximion:
I think this is a good idea for the reasons you stated: If we are missing a codec, there is no point in trying to list installed codecs, since these can be safely assumed to not be the ones we are looking for.
The question is whether putting this into PK itself makes sense...

Do we ever want providesCodec to be used to list *all* (installed/available) codecs? Before AppStream existed, I would have said yes to that question for sure.
Also, is there any option to filter the results in the client for only available packages?

Conan-Kudo:
@iainlane Can you rebase onto master, please?

dantti:
hmm should session-installer just send not-installed filter?

hughsie:
I wondered that too....

iainlane:
On 5 July 2017 20:41:27 BST, Daniel Nicoletti <notifications@github.com> wrote:
>hmm should session-installer just send not-installed filter?

I'm away this week so can't check the code, but this is about removing sessioninstaller. If *gnome-software* can do this then that would be better I think. :)

Cheers for the comments!

ximion:
I thought GNOME Software does that filtering?

iainlane:
OK, I'm back!

We can fix this in gnome-software by doing something like this:

```
diff --git a/plugins/packagekit/gs-plugin-packagekit.c b/plugins/packagekit/gs-plugin-packagekit.c
index 2b96b6c7..f91112a6 100644
--- a/plugins/packagekit/gs-plugin-packagekit.c
+++ b/plugins/packagekit/gs-plugin-packagekit.c
@@ -619,6 +619,7 @@ gs_plugin_add_search_what_provides (GsPlugin *plugin,
 	gs_plugin_status_update (plugin, NULL, GS_PLUGIN_STATUS_WAITING);
 	filter = pk_bitfield_from_enums (PK_FILTER_ENUM_NEWEST,
 					 PK_FILTER_ENUM_ARCH,
+					 PK_FILTER_ENUM_NOT_INSTALLED,
 					 -1);
 	results = pk_client_what_provides (PK_CLIENT (priv->task),
 	                                   filter,
```

@hughsie - any opinion? I wonder if there are cases where you would want to show installed software when searching WhatProvides, or is it always to install something you don't have & so anything installed is going to be a false positive? If it's the former then this patch will hide things you want to show.

hughsie:
I think I'd rather the fix went into gnome software.

iainlane:
OK, then I'll push that diff ^-, cheers. I like solving it in gnome-software more than a workaround in PackageKit.

PR #204, opened 2017-07-21
Add the ability to install updates on reboot in PackageKit-cron

tjikkun:


contrib/cron/packagekit-background
```diff
@@ -15,6 +15,13 @@ ENABLED=no
 #
 CHECK_ONLY=no
 
+## Type:	yesno
+## Default:	no
+#
+# Trigger updates, so they will be installed on reboot.
+#
+UPDATE_ON_REBOOT=no
```
hughsie: I'm not sure I like that name, wouldn't "UPDATE_OFFLINE" be clearer?

tjikkun: Agreed, updated

PR #210, opened 2017-08-07
zypp: Return error if invalid package IDs are detected

JonathanKang:


backends/zypp/pk-backend-zypp.cpp
```diff
@@ -1737,7 +1737,11 @@ pk_backend_destroy (PkBackend *backend)
 static bool
 zypp_is_no_solvable (const sat::Solvable &solv)
 {
-	return solv.id() == sat::detail::noSolvableId;
+	if (!solv) {
+		return FALSE;
+	} else {
+		return solv == sat::detail::noSolvable;
+	}
 }
```
mlandres: No! 
```
-    return solv.id() == sat::detail::noSolvableId;
+    return solv == sat::Solvable::noSolvable;
```
That's all. (note that it's  `sat::Solvable::noSolvable` and not `..::detail::...`)

The `if (!solv) return FALSE; ` is wrong. `if (!solv)` you have no solvable, so you'd want to return `true`. 

`!solv` and `solv == sat::Solvable::noSolvable` are equivalent.

JonathanKang:
@mlandres PR updated according to you review. Thanks.

mlandres:
LGTM

PR #213, opened 2017-08-12
Make proxy settings work

sfanxiang:
These patches include several fixes for session proxy settings, and some updates to outdated docs.


hughsie:
Looks great, thanks!

PR #222, opened 2017-11-10
aptcc: fix offline updates by sanitizing origin more aggressively

apachelogger:
origin may contain any number of nonesense, strip everything but words and
digits out!
specifically comma is breaking offline updates as the list of 'prepared'
packages is joined using comma, so having lingering commas in our package
ids utterly breaks splitting of the list and thus the offline update as a
whole.
seeing as origin is entirely free form it is probably wise to be very
aggressive in what we remove.

I've also added a bunch more details to the formats involved here for
prosperity and future reference.

### DIFF
--- a/backends/aptcc/apt-utils.cpp
+++ b/backends/aptcc/apt-utils.cpp
@@ -26,6 +26,7 @@
 #include <glib/gstdio.h>
 
 #include <fstream>
+#include <regex>
 
 PkGroupEnum get_enum_group(string group)
 {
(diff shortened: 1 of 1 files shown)


ximion:
Aha! That explains why I was never able to reproduce this issue.
I dislike another regex in the code, but this is probably fast enough, so thumbs up.

aleixpol:
LGTM +1

@ximion FWIW, usually the concern about regex is not speed but readability...

ximion:
@aleixpol That depends on the regex - if you have a loop with thousands of entries and recompile the regex in the loop every time, it can slow down the routine significantly.
Also, the overhead of an regex in itself is sometimes not worth using one if string operations for what you want to achieve are simple enough.
(in this case it's fine though)

PR #223, opened 2017-11-13
Minor 2 line change. Removed unnecessary check

loa-in-:
Removed unnecessary check with unclear reasoning behind it. There is …
…an outstanding bug request caused by this at: https://bugzilla.redhat.com/show_bug.cgi?id=1393168

### DIFF
--- a/contrib/command-not-found/pk-command-not-found.c
+++ b/contrib/command-not-found/pk-command-not-found.c
@@ -803,6 +803,11 @@ main (int argc, char *argv[])
 	len = strlen (argv[1]);
 	if (len < 1)
 		goto out;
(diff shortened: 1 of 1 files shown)


hughsie:
This was added in 0e85001b247c2ed933e5eb028cd167a8e3e40589 to fix https://bugzilla.redhat.com/show_bug.cgi?id=1151185 -- can you come up with a fix that addresses the original bug too please.

loa-in-:
Should be ok now, because there are no other special sequences in this context than "." and ".." that should be taken into consideration. Guaranteed by the filesystem.
Still will ignore "." and ".." inside root directory even though they're not there.

hughsie:
I've fixed this a slightly different way, thanks for the patches.

PR #225, opened 2017-11-29
service: Use ConditionPathExists=!/run/ostree-booted

cgwalters:
Today gnome-shell unconditionally tries to talk to PK, which doesn't
really do anything useful for ostree-managed systems (including rpm-ostree),
for example Fedora Atomic Workstation.

This is similar to https://github.com/rpm-software-management/dnf/pull/723

### DIFF
--- a/data/packagekit-offline-update.service.in
+++ b/data/packagekit-offline-update.service.in
@@ -5,6 +5,8 @@ DefaultDependencies=no
 Requires=sysinit.target dbus.socket
 After=sysinit.target dbus.socket systemd-journald.socket
 Before=shutdown.target system-update.target
+# See packagekit.service
+ConditionPathExists=!/run/ostree-booted
 
 [Service]
 Type=oneshot
--- a/data/packagekit.service.in
+++ b/data/packagekit.service.in
@@ -1,5 +1,9 @@
 [Unit]
 Description=PackageKit Daemon
+# PK doesn't know how to do anything on ostree-managed systems;
+# currently the design is to have dedicated daemons like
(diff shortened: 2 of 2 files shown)


pwithnall:
Looks good to me. This doesn’t affect us on Endless since we don’t even package PackageKit, but if it’s useful for rpm-ostree, then go for it as far as I’m concerned. 👍

cgwalters:
Yeah I was pretty sure you didn't use PK, just wanted to double check.

ximion:
Does an rpm-ostree backend for PackageKit make sense in future?

cgwalters:
The thing is that rpm-ostree is *already* a DBus daemon that has its own concept of transactions, history and state.  (Though honestly some of what PK does is better but it's a complex topic)

ximion:
Ah, very interesting! I need to learn more about this, as soon as I find some spare time (having a dpkg/apt-ostree is on my "remote future projects" list for a while now)
Having an own daemon that can provide API specific to the usecases of rpm-ostree doesn't sound like a bad idea.

cgwalters:
@ximion Yep, supporting such a thing is the big reason why libostree is separate from rpm-ostree!  (Also non-host-management systems like flatpak and build cache systems like BuildStream).  It's a *lot* of work though.  Briefly one major issue is that in order to support e.g. running `apt` inside Docker, you still need to have a "traditional" package manager which mostly works the same way it always has, and that's where we are today in Fedora with `dnf` versus `rpm-ostree`...but we're trying to share code in libdnf, etc.  Anyways if you decide to embark on it, or even just experiment, feel free to join https://mail.gnome.org/archives/ostree-list/ !

PR #228, opened 2017-12-15
aptcc: do not crash on double free when updating a config file

apachelogger:
filename is being inserted into the argv array, this results in a double
free when we later free filename (first) AND then the argv array (second)


ximion:
Yes, indeed! That's an ugly one, especially since the memory allocation and assignment to the array happen right next to each other. Bah!
Thanks for spotting this and other issues - remind me to invite you to a beer at the next conference :-)

PR #231, opened 2018-01-16
Fix getting user session ID with D-Bus "user bus" model

kalev:
In case a PackageKit requesting process is not part of the user's login
session, we ask systemd-logind for the the user's "display" session
instead.

This fixes gnome-software "failed to set proxy" warnings.

https://bugs.freedesktop.org/show_bug.cgi?id=101281


ximion:
Oooh! So that was the cause for that issue!
The patch looks good to me!

hughsie:
@kalev another one that had me scratching my head, nice one!

PR #238, opened 2018-02-08
Include whether it's a Source repository

aleixpol:
Most repositories will have a "deb" and a "deb-src" variant and
currently the description they offer is the same.

### DIFF
--- a/backends/aptcc/apt-sourceslist.cpp
+++ b/backends/aptcc/apt-sourceslist.cpp
@@ -468,6 +468,10 @@ string SourcesList::SourceRecord::niceName()
         ret += " (" + joinedSections() + ")";
     }
 
+    if(Type & SourcesList::DebSrc) {
+        ret += " Sources";
+    }
+
     return ret;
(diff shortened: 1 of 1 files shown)


dantti:
The string should not be translated (and will not anyway), can you give an example of before/after strings?

aleixpol:
Now:
`/etc/apt/sources.list.d/neon.list:deb http://archive.neon.kde.org/dev/unstable/ xenial main` is `Xenial (main)`
`/etc/apt/sources.list.d/neon.list:deb-src http://archive.neon.kde.org/dev/unstable/ xenial main` is also  `Xenial (main)`

After my patch, the second will be `Xenial (main) Sources`.

aleixpol:
screenshot_20180209_184058

dantti:
Ok, I see, the only issue is that this won't be properly translated so IMO it's better to not even try to translate this.

aleixpol:
Merging as it seems there's some agreement.

PR #239, opened 2018-02-09
zypp: removing memory leak from zypp backend [v2].

jcfaracco:
There is a memory leak related to variable 'items' when remove package
thread function fails. This commit moved the pool 'items' to the stack
inside all functions where 'items' is used with a dynamic allocation.
This approach avoids clean up/free structures inside the code.


kalev:
Looks good to me! Thanks.

PR #240, opened 2018-02-12
aptcc: Return multiple packages when using '|' operator

davidmhewitt:


davidmhewitt:
Apologies, didn't add a message for this.

I've noticed that when passing an array of packages names to the aptcc backend via the `search_names` method, only the first package result was returned.

Looking at https://github.com/hughsie/PackageKit/blob/master/backends/aptcc/pk-backend-aptcc.cpp#L748 , it joins the array with the `|` character. But then the `Matcher` class has some issues where there separate search strings are not parsed out.

I couldn't find enough documentation to know how this Matcher is _supposed_ to work or what the syntax is supposed to be. But this works more like how I'd expect. This may be totally incorrect, but let me know how this could be improved if that is the case.

Thank you

cassidyjames:
Following up on this. It sounds like this enables a significant performance improvement for app store contexts that need to display a lot of results, like a home page. In our tests, this lets us query in parallel and improves the load time from ~25 secs to ~5 secs.

@hughsie Was the previous behavior in PackageKit a bug, and if so, does this solve it in a satisfying way?

ximion:
@cassidyjames You can already query in parallel - don't you use the async interfaces of PK? (you really should, using PK with sync interfaces is really and likely always will be very slow)

davidmhewitt:
The async methods are used, however, we don't typically display a section of packages to the user on the homepage until all the results have come back to prevent too many changes in the interface while it's loading and the user is trying to browse.

However, every call to the `search_names` or `search_details` method, async or not, iterates through the whole apt package list to run the matcher against it. This PR allows multiple search queries to be passed, and only one iteration needs to occur to match all.

davidmhewitt:
I've been looking into this a bit more by looking into the code of the `dnf` backend.

From the code, it looks as though passing an array of search terms to `search_names` or `search_details` will return an array of packages that match any of the search terms in the array (i.e. an OR statement across all of the search terms). I'll confirm this in practice on Fedora when I get chance.

The current behaviour of the `aptcc` backend is that an array of packages that match *only the first* search term in the array. This doesn't look intentional as there is code in there to join all of the search terms from the array together with `|` characters. So I assume this is a bug in the `aptcc` backend.

This PR adds a fix to make the behaviour the same as the `dnf` backend.

However, there is also a lot of code in this `Matcher` class for parsing brackets and quotes etc... which adds a lot of complication. I'm happy to re-write this class without the complication as there is nothing in `search_names` or `search_details` that requires that.

But, before I start work on that, I'd like some guidance or confirmation that the above is the expected behaviour and if possible, that there isn't some reason why the `aptcc` backend is keep around some code to try and do some clever parsing of the search query, which doesn't look like it's fully implemented anyway.

davidmhewitt:
I've just done some testing on fedora and can confirm that the behaviour of the `aptcc` backend is different to the `dnf` backend in this regard.

Using the following python code for testing (note the use of passing two search queries in an array to `search_names`):
```
import gi
gi.require_version('PackageKitGlib', '1.0')

from gi.repository import PackageKitGlib

pk = PackageKitGlib.Client()

res = pk.search_names(PackageKitGlib.FilterEnum.NONE, ["d-feet","calculator"], None, lambda p, t, d: True, None)
package_ids = res.get_package_array()

for package_id in package_ids:
    print package_id.get_name ()
```

On `aptcc`:
```
# python pktest.py 
d-feet
```

On `dnf`:
```
$ python pktest.py 
d-feet
gnome-calculator
d-feet
lumina-calculator
pantheon-calculator
sugar-calculator
texlive-calculator-doc
texlive-calculator
xfce4-calculator-plugin
gnome-calculator
pantheon-calculator
```

Should the behaviour of the apt backend not be the same?

ximion:
> But, before I start work on that, I'd like some guidance or confirmation that the above is the expected behaviour and if possible

@hughsie Should all backends support the | operator, so frontends can rely on it? Or is this some magic specific to dnf?
(generally, I think this is a useful addition to have and we should add it to aptcc)

> that there isn't some reason why the aptcc backend is keep around some code to try and do some clever parsing of the search query, which doesn't look like it's fully implemented anyway

I would need to look into this in details, but it's very likely that we worked on improving search a while back and the work just wasn't finished. Could also be that the code is a leftover from Synaptic/apt-get where some of aptcc's code comes from.

hughsie:
@ximion I forger the details for "|" but I think it's a good idea to add.

ximion:
@hughsie Is this documented somewhere? I couldn't find it in a quick search, and for frontends to rely on it we should at least somewhat make sure most backends can support this feature.

davidmhewitt:
The "|" operator seems to be specific to the internals of the `aptcc` backend. The array of packages passed to the packagekit-glib search methods are joined into a single string by the `aptcc` backend with "|" characters in between each entry.  The `Matcher` class then uses that to iterate through the apt packages. There's no need for documenting a "|" character for frontends, since you already pass an array of package name/detail queries.

For example, the `dnf` backend just passes that array of package search queries to the package manager's search method without doing any prior processing.

So I don't think there's particularly a standard syntax here that should be used across all backends and documented for frontends. But it does sound like there's some agreement that the `aptcc` backend should at least return packages that match any of the search queries, rather than packages that only match the first query.

This PR already does that, but if nobody disagrees, I'll look to simplify this `Matcher` class and remove the extra parsing for characters other than "|", because even if they were used in the search query, they would have undefined effects on the outcome of the search due to the logic behind them not being implemented.

dantti:
What the check for start == '|' does?
The code above of that does indeed looks like it's trying to get the cmd line options of apt for package matching, I don't even recall if it was I that added that, and yes, the first part does looks like a fix as the array should be ORed and it's not.

davidmhewitt:
@dantti It advances the iterator past the "|" character, so that the next string can be parsed.

There's this line that causes the string to stop being parsed after the first "|" character. So, if we increment the iterator after we see that, we can continue with the other strings:
`while (start != end && *start != '|' && *start != ')') {`

dantti:
Isn't it better then to just remove the  ***start != '|'**  from the while?

davidmhewitt:
Maybe, but the iterator still needs to be advanced somewhere, so that we don't include the "|" in the search string.

It doesn't matter anyway, I plan to try and re-write this class to make the parsing of the search query much simpler as the other metacharacters aren't used, so it doesn't need to be so complex.

dantti:
Ok, I'll merge this, but indeed that piece of code does needs some love :)

PR #244, opened 2018-03-06
Reflect latest changes in libdnf

j-mracek:
Requires: libdnf-0.13.0+

G object DnfPackageSet in libdnf was replaced by c++ class therefore
g_object_unref cannot be used anymore, but the new dnf_packageset_free has to be
used instead.

### DIFF
--- a/backends/dnf/pk-backend-dnf.c
+++ b/backends/dnf/pk-backend-dnf.c
@@ -836,7 +836,7 @@ dnf_utils_run_query_with_newest_filter (DnfSack *sack, HyQuery query)
 	hy_query_free (query_tmp);
 	g_ptr_array_unref (results_tmp);
 
-	g_object_unref (pkgset);
+	dnf_packageset_free(pkgset);
 
 	return results;
 }


hughsie:
We appreciate the patch! @kalev -- what is our plan here with the DNF API-incompatible changes?

kalev:
Thanks Jaroslav!

I guess we have a few options here:

1) start bundling libdnf again and keeping track of libdnf master
2) start bundling libdnf again and keeping track of the libdnf rpm-ostree branch (https://github.com/projectatomic/libdnf/tree/rpmostree-branch)
3) keep building against released libdnf and branch libdnf before the C++ changes and start doing libdnf maintenance releases again (libdnf 0.11.x)

I guess it depends on what the DNF team prefers, I would be fine with either option. It would also be OK I think to change the PK libdnf backend to C++ if that's easier for libdnf maintenance.

My only wish is to keep PackageKit in a releasable state, which means that if we start tracking libdnf master, it should be in a working state :)

@j-mracek Which of the option would you suggest?

j-mracek:
Each of the solution has some advantages.

Conan-Kudo:
I'd personally prefer option 3, as it makes it possible for me to ship updated releases of PackageKit in Mageia and OpenMandriva, both of which are using the DNF backend and I'm the maintainer for the package in both distributions.

If we want to start moving to the new libdnf code, that's fine, but I want a firm timeline on stabilizing for making releases because I can't reasonably deal with this using an unreleased libdnf.

That is, if the next PK release is going to require libdnf master, I want to see a release of libdnf master (and a corresponding release of the whole dnf stack compatible with it...) to go with it.

PR #246, opened 2018-03-21
aptcc: Not all downloads have to be packages

julian-klode:
When the mirror method is used in apt 1.6 or newer, a mirror list is
downloaded prior to any package, causing PackageKit to crash after
casting its item to pkgAcqArchiveSane.

pkgAcqArchiveSane is a local workaround to expose some protected
stuff in pkgAcqArchive, so we first do a dynamic_cast to pkgAcqArchive
which will be nullptr if it is not a pkgAcqArchive and then cast it
statically to pkgAcqArchiveSane.

Bug-Ubuntu: https://launchpad.net/bugs/1754265

CC @ximion


dantti:
LGTM Thanks.

ximion:
Awesome! I was looking at that crash a few days ago, and it didn't make much sense to me - but this is a perfect explanation! :-)

PR #247, opened 2018-03-30
aptcc: Simplify search methods

davidmhewitt:
I promised a code cleanup/simplification in the last PR I made around the aptcc search code, so here we go.

In summary, this PR removes the ability to use regex in search queries for `search_names` and `search_details` which I think makes this more consistent with the other backends. Even when I was testing the regex searches, the outcomes were always pretty unpredictable due to the weird extra parsing of brackets and quotes that this backend did, so I would hope no frontend is using that "feature".

Using regex caused a couple of things that I think were bugs where using regex characters in search queries might cause them to get interpreted as regex sequences, or break the search altogether. For example, doing a `search_details` for `$` would return all packages, rather than packages that had `$` in their description.

But hopefully this code is much more maintainable and readable now.


dantti:
Thanks, this does looks much better, I only have two questions:
- newer C++ has a regex class, I dunno if it's good but since it's in the standard I was planning on using, have you used it?
- regex are always slower than manual string matching, since it's basically a string.contains() that we do can't we get rid of regex completely?

davidmhewitt:
Yes, I suppose it's best to do a `string.contains`. I was thinking that regex might be more efficient for larger arrays of search queries (if the regex implementation has a good compiler and optimizer, which I'm not sure if it does).

But it's more likely that the search methods will mostly be used with single search queries, so it makes more sense to optimize for that case. I'll change this branch to use basic string comparison instead.

davidmhewitt:
@dantti I've removed all of the regex code altogether now, let me know how it looks. Thanks.

dantti:
Looks a lot simpler now, I'd just change the methods to foo(const vector<string> &strings).
Have you tried to measure time?

davidmhewitt:
@dantti I've changed to using `const vector<string> &queries` now, let me know if that's what you meant.

On average, it seems around 12% faster than the regex implementation for a single search query.

davidmhewitt:
Ah, good point! I'll fix that soon, but I have to go out now. Thanks for your help.

ximion:
Thank you for this PR! Looks good to me! (I couldn't find anything more that Daniel didn't already mention).

davidmhewitt:
Thanks both! I've resolved the case sensitivity issue, should be good for review again.

dantti:
Ok, last thing, since statistically I really believe most characters are lower case using tolower would perform less changes, (at least with QString this means if I don't change a char I avoid a deep-copy), anyway, due this I found on the documentation that you should use unsigned char when using tolower or toupper:
http://en.cppreference.com/w/cpp/string/byte/tolower
Thanks

dantti:
Thanks, I have squashed it to avoid the touch on configure.ac

PR #251, opened 2018-05-16
pk-offline-update: shut down services cleanly before rebooting

sourcejedi:
do not use forced reboot.
https://bugzilla.redhat.com/show_bug.cgi?id=1564462

Seconded by Zbigniew Jędrzejewski-Szmek.
"org.freedesktop.systemd1.Manager.Reboot is unnecessarily harsh, and
shutting down normally through StartUnit is much better."

Having pk-offline-update use a different, harsher shutdown path than is
used in normal operation, is a really no-good idea and can trigger problems
which are harder than necessary to test, reproduce, etc.

For example, Fedora is not unmounting the root filesystem cleanly! due to a
plymouth bug.  In *some* circumstances, the graceful systemd shutdown
will avoid this.  Because a graceful shutdown *with an LVM rootfs*, will
transition back into dracut for final shutdown.  And dracut includes a
workaround for plymouth or similar very weird processes, which has recently
been modified and hopefully works more reliably now.

https://unix.stackexchange.com/questions/441511/does-recovering-journal-prove-an-unclean-shutdown-unmount

https://bugzilla.redhat.com/show_bug.cgi?id=1575376#c10

https://github.com/dracutdevs/dracut/commit/df6bb5e959178cba06118493a7c8d019e84d54e7

I was able to reproduce the plymouth bug, by downgrading a package and then
running an offline update. I have a systemd-shutdown hook which pauses
before rebooting/entering dracut; this let me see log messages about
failing to remount the root filesystem as read-only, because it was still
busy (for writing) - i.e. by plymouth.

With this commit applied, the filesystem was clean after the reboot, and
did not require journal recovery.  Thanks to the version of dracut which
makes sure to kill plymouth before unmounting the root filesystem.


hughsie:
@kalev this makes sense to you?

kalev:
It does! Thanks for the patch, @sourcejedi

Issue #245, opened 2018-03-07
Should remove unused dependencies (APT backend)


AsciiWolf:
Separate issue from #201 - aptcc backend has the same issue.

PackageKit aptcc backend does not support uninstallation of unused dependencies. This is a major issue for regular users, especially since GNOME Software does not report when installing a desktop environment or another big dependency with an application and there is no easy way for regular user to remove the unused dependencies after uninstalling the application. Please, consider fixing this.

robert-ancell:
See #253

robert-ancell:
Ubuntu bug https://launchpad.net/bugs/1552792

AsciiWolf:
Autoremove is now supported by the aptcc backend!

PR #253, opened 2018-05-23
aptcc: Only autoremove packages that relate to the current transaction

robert-ancell:


robert-ancell:
This appears to correctly identify the packages that can be removed. However, I'm not an expert in the pkg-apt interface. I suspect after my `tryToRemove()` calls there should be some checks as the above code seems to do?

robert-ancell:
See #245

ximion:
Can you change the patch to only ever attempt autoremoval if the `autoremove` flag is given by the frontend? (The method should have a bool argument for that already).
The backend should never remove stuff on its own without being asked to explicitly.

It is part of an ongoing debate on whether aptcc should actually remove packages that have nothing to do with the current transaction and that may have been obsoleted by something else. But since APT has no way of knowing that information, I think the "remove all the things" option is valid. I know others disagree on this though ;-)

dantti:
I was sure I had implemented this, and yes, there is the bool autoremove that must be honored.
Weird in Apper I even have a define to change the default behavior when calling packagekit, if this is not working I believe it broke somehow

robert-ancell:
The autoremove flag currently removes _all_ packages that aren't depended on. I'm not sure what the intended behaviour is for this flag - is it supposed to mean only packages related to this transaction?

dantti:
Ah! So this patch will remove unused packages that only relates to the package being removed?

robert-ancell:
@dantti yes. I think I misread the purpose of the autoremove flag and thought it was more about doing a general cleanup rather than specifically removing package related to the current transaction. So this patch is (now) really just making it less aggressive.

The logic is copied from aptd, which did this automatically. The motivation is to try and make GNOME Software behave more like the old Ubuntu Software Center and clean up when removing apps.

robert-ancell:
The main case being solved is when you have something like a game that has two binary packages 'game' and 'game-data'. On removal 'game-data' can get left behind and may be the vast majority of the disk space.

dantti:
Ok, so IMO we need a new flag, and this patch is too intrusive as changing the current behavior.

robert-ancell:
From src/org.freedesktop.PackageKit.Transaction.xml:

```
      <arg type="b" name="autoremove" direction="in">
        <doc:doc>
          <doc:summary>
            <doc:para>
              Either <doc:tt>true</doc:tt> or <doc:tt>false</doc:tt>.
              This option is only really interesting on embedded devices with a limited amount of
              flash storage.
              It suggests to the packagekit backend that dependencies installed at the same time as
              the package should also be removed if they are not required by anything else.
              For instance, if you install OpenOffice, it might download libneon as a dependency.
              When <doc:tt>auto_remove</doc:tt> is set to true, and you remove OpenOffice then
              libneon will also get removed automatically.
            </doc:para>
          </doc:summary>
        </doc:doc>
      </arg>
```

This suggests the current behaviour is incorrect - it should not be removing all pacakges, only those related to the transaction.

ximion:
@robert-ancell This sounds pretty good!
@dantti If the result is really less aggressive cleanup behavior, this patch should be fine (I haven't tested this though), and maybe we indeed need a global autoremoval / cleanup-cruft method in PackageKit (would actually make sense to me, it could also clean up cached packages. Although I have a feeling that hughsie would want the backends to take care of this automatically, instead of adding a PK method for it that has to be explicitly called by frontends).

dantti:
Ok, so first please do not remove the doAutoRemove function, and before getting the package deps check if autoremove is set.

robert-ancell:
@dantti you want to keep the function even if nothing is calling it? (Can do, just checking that is your intention)

dantti:
it's also important that these packages to be removed are emitted on simulate methods.

dantti:
Yes, because that's what apt autoremove does, in in the future we want to do that it's already implemented.

robert-ancell:
I can't work out if there's any special code required for simulation but running from the command line:

```
$ pkcon remove alarm-clock-applet
Resolving                     [=========================]         
Finished                      [                         ] (0%)  
The following packages have to be removed:
 alarm-clock-applet-0.3.4-1build1.amd64	alarm clock panel indicator
 gconf-service-3.2.6-4ubuntu1.amd64	GNOME configuration database system (D-Bus service)
 gconf-service-backend-3.2.6-4ubuntu1.amd64	GNOME configuration database system (D-Bus service)
 gconf2-3.2.6-4ubuntu1.amd64	GNOME configuration database system (support tools)
 gconf2-common-3.2.6-4ubuntu1.all	GNOME configuration database system (common files)
 libgconf-2-4-3.2.6-4ubuntu1.amd64	GNOME configuration database system (shared libraries)
 libunique-1.0-0-1.1.6-6.amd64	Library for writing single instance applications - shared libraries
Proceed with changes? [N/y]
```

Is getting the list of packages to remove done using a simulation?

dantti:
yes, so simulate is working as expected

backends/aptcc/apt-intf.cpp
```diff
@@ -2179,6 +2180,8 @@ bool AptIntf::runTransaction(const PkgList &install, const PkgList &remove, cons
             }
 
             m_cache->tryToRemove(Fix, verIt);
+
+            getDepends (removed_deps, verIt, true);
```
dantti: this should only run if autoremove is set

backends/aptcc/apt-intf.cpp
```diff
@@ -2196,6 +2200,14 @@ bool AptIntf::runTransaction(const PkgList &install, const PkgList &remove, cons
         }
     }
 
+    // Remove dependencies that are no longer required
```
julian-klode: Move that block inside the action group to get better performance.

backends/aptcc/apt-intf.cpp
```diff
@@ -2179,6 +2180,9 @@ bool AptIntf::runTransaction(const PkgList &install, const PkgList &remove, cons
             }
 
             m_cache->tryToRemove(Fix, verIt);
+
+            if (autoremove)
+                getDepends (removed_deps, verIt, true);
```
julian-klode: `getDepends()` only considers Depends, not Pre-Depends, Recommends, or Suggests, and is thus incomplete.

robert-ancell:
I've updated the branch to use the delta-autoremove method - seems to work!

dantti:
Cool thanks, it LGTM I just would like to test it before merging

backends/aptcc/apt-intf.cpp
```diff
@@ -2156,6 +2156,16 @@ bool AptIntf::runTransaction(const PkgList &install, const PkgList &remove, cons
         const bool preserveAuto;
     };
 
+    // Calculate existing garbage before the transaction
+    PkgList initial_garbage;
+    if (autoremove) {
+        for (pkgCache::PkgIterator pkg = (*m_cache)->PkgBegin(); ! pkg.end(); ++pkg) {
+            const pkgCache::VerIterator &ver = m_cache->findCandidateVer(pkg);
+            if (!ver.end() && m_cache->isGarbage(pkg))
+                initial_garbage.push_back(ver);
```
julian-klode: Building a list of versions seems rather suboptimal given that we are talking about packages, and don't need the versions - it makes it fragile in case candidate versions change during the transaction - we'd not find the garbage anymore.

It's also a vector, which, for large counts of garbage, would make the code below quite slow. My suggestion would be to use an std::unordered_set<pkgCache::Package*>, that should work reliably. You can then just pass pkgCache::PkgIterators (they behave like Package pointers) to insert() and find().

julian-klode:
Oh, I was wrong, `contains()` actually looks for packages. So this works fine, as long as the installed package has a candidate. I think they do.

robert-ancell:
@julian-klode does your last comment relate to the review comment - i.e. do you still want me to change from a vector to std::unordered_setpkgCache::Package* ?

julian-klode:
You can in fact pin down installed packages so that there is no candidate, and I think that can happen easily. For example:

```
Package: apt
Pin: release *
Pin-Priority: -1
```

leads to

```
$ apt-cache policy  apt
apt:
  Installed: 1.7.0~alpha1+0~201806280812~ubuntu18.10.1
  Candidate: (none)
  Version table:
 *** 1.7.0~alpha1+0~201806280812~ubuntu18.10.1 -1
        500 http://ppa.launchpad.net/deity/sid/ubuntu cosmic/main amd64 Packages
        100 /var/lib/dpkg/status
     1.7.0~alpha1 -1
        500 http://de.archive.ubuntu.com/ubuntu cosmic/main amd64 Packages
```

With the current code, packages pinned like this would never be considered garbage, since they do not have a candidate.

 So, you probably have to use the package set, and when removing the garbage, use 
```diff
-            const pkgCache::VerIterator &ver = m_cache->findCandidateVer(pkg);
+            const pkgCache::VerIterator &ver = m_cache->findVer(pkg);
```

so you always get a version there for an installed package.

julian-klode:
Or, in fact, just use `pkg.CurrentVer()` in the autoremoval loop.

julian-klode:
So, I was thinking just replacing `cache->findCandidateVer(pkg)` with `pkg.CurrentVer()` in both loops probably works in all cases, and also is a bit faster :)

ximion:
I'm confident we get this merged this Debconf (Julian and I are both here), but I want to wait for his improvements first.

julian-klode:
Updated in #260

ximion:
Thanks! Merged the new PR instead of this one, thank you both for fixing the issue!

PR #255, opened 2018-06-12
Allow module filtering

j-mracek:
It should disable packages from disabled modules.

The patch requires libdnf-0.15.0


backends/dnf/pk-backend-dnf.c
```diff
@@ -799,6 +800,8 @@ dnf_utils_create_sack_for_filters (PkBackendJob *job,
 			return NULL;
 	}
 
+    dnf_sack_filter_modules(sack, dnf_context_get_repos(job_data->context), install_root);
```
hughsie: Tab here, not spaces please. Also, a space after both brackets. Thanks.

PR #260, opened 2018-07-26
 aptcc: Only autoremove packages that relate to the current transaction 

julian-klode:
Updated version of #253 that also fixes a corner case.


ximion:
Neat! Thank you!
(Sitting next to the PR author speeds up reviews :P)

PR #261, opened 2018-08-02
Add additional argument to the dnf_sack_filter_modules() call.

dmach:
Libdnf API has changed and the function takes an additional argument: platformModule.
It's set to NULL to trigger platform module auto-detection based on /etc/os-release.


kalev:
Thanks!

PR #272, opened 2018-09-01
aptcc: removing duplicate delete call.

jcfaracco:
This commit removes a duplicate call inside providesCodec() function.
When cancel is active, the loop call tries to delete 'matcher' and after
breaks the loop. But, the function will tries to delete 'matcher' again
at the end.


dantti:
Thanks

PR #273, opened 2018-09-04
aptcc: Fix invalid version dereference in AptInf::providesCodec()

julian-klode:
ver.Arch() was accessed even if ver.End() was true and ver thus
pointed to invalid memory, causing undefined behavior and likely
crashes, such as the Ubuntu bug below.

Bug-Ubuntu: https://bugs.launchpad.net/bugs/1790671


ximion:
Eww...
Thank you for the patch!

PR #278, opened 2018-09-21
Never assert when an interactive TTY is not available

hughsie:
Also, fd==0 is unlikely, but valid. Do the right thing in finalize().


kalev:
Looks good to me! :+1:

hughsie:
Nope, to fix a RHEL bug that popped up in 7.6 testing -- when there was no TTY set up.

AdamWill:
yeah, I found the real fix for that bug a few minutes later. but that's old news now!

PR #279, opened 2018-09-26
aptcc: Always try multi-arch lookups for lists

julian-klode:
aptcc wrongly only tried the multi-arch names with architecture
included if multi-arch was enabled. dpkg however always uses arch-
qualified names - otherwise it would have to rename half its database
after enabling / disabling multi-arch.


dantti:
LGTM, although I recall that when you installed without multi-arch support the default behavior was files without arch, so dpkg changed the default behavior?
In any case this will work since there's still the fallback behavior

PR #282, opened 2018-10-02
offline update: Fix translations to show up

kalev:
Do the setlocale stanza in main() to correctly set up gettext.

Fixes: https://github.com/hughsie/PackageKit/issues/280

(Completely untested, just copied the gettext stuff from pk-main.c)


hughsie:
Looks sane to me. If CI passes, please merge.

PR #285, opened 2018-10-09
dnf: Run subscription-manager sync before refreshing metadata

hughsie:


kalev:
Hm, libdnf has a bunch of bugs in the repo reloading. I think we need to do dnf_repo_loader_get_repos() after the subscription manager call to force it to correctly reload, similar to what pk_backend_yum_repos_changed_cb does and also clear the sack cache etc.

Conan-Kudo:
Why didn't we use librhsm here?

hughsie:
I thought that was dead? The explicit exec was what the subscription manager team wanted us to do.

Conan-Kudo:
That's news to me, since libdnf still uses it, and there's a PR to convert its usage into a passive runtime plugin.

PR #289, opened 2018-11-02
Socket-activated client helper

julian-klode:
Introduce support for a socket-activated client helper for aptcc. This solves a problem where the current client helpers are children of the client starting the transaction and potentially bound to the client's lifetime; so if the client is exited early, while the transaction is still running, no debconf prompts would be shown.

A socket-activated helper solves this problem by making sure we only ever run one helper as a systemd user service.

This change is incompatible with the debconf-kde hacks, I'm not sure what to do about those yet.

### DIFF
--- a/backends/aptcc/pk-debconf-helper.c
+++ b/backends/aptcc/pk-debconf-helper.c
@@ -48,7 +48,7 @@ int main(void)
 		return 1;
 	}
 
-	g_timeout_add_seconds(5, exit_loop, NULL);
+	g_timeout_add_seconds(60, exit_loop, NULL);
 
 	g_main_loop_run(main_loop);
 }
--- a/backends/aptcc/pk-debconf-helper.c
+++ b/backends/aptcc/pk-debconf-helper.c
@@ -1,0 +1,54 @@
+#include "config.h"
+#include <stdio.h>
+#include <packagekit-glib2/pk-client.h>
+#include <packagekit-glib2/pk-client-helper.h>
+#include <systemd/sd-daemon.h>
+
+static GMainLoop *main_loop;
+static PkClientHelper *helper;
+
+static gboolean exit_loop(gpointer user_data)
+{
+	g_debug("Checking for active connections");
+	if (!pk_client_helper_is_active(helper)) {
+		g_message("No active connections, exiting");
+		g_main_loop_quit(main_loop);
+	}
+	return TRUE;
+}
+
+int main(void)
+{
+	char **argv = NULL;
+	char **envp = NULL;
+	GSocket *socket = NULL;
+	GError *error = NULL;
+	int fd = -1;
+
+	main_loop = g_main_loop_new(NULL, FALSE);
+	pk_client_create_helper_argv_envp(&argv, &envp);
+
+
+	if (sd_listen_fds(0) != 1) {
+			g_error("No or too many file descriptors received.\n");
+			exit(1);
+	}
+
+	fd = SD_LISTEN_FDS_START + 0;
+	socket = g_socket_new_from_fd(fd, &error);
+
+	if (error != NULL) {
+		g_error("%s\n", error->message);
+		return 1;
+	}
+
+	helper = pk_client_helper_new();
+	if (!pk_client_helper_start_with_socket(helper, socket, argv, envp, &error)) {
+		g_error("%s\n", error->message);
+		return 1;
+	}
+
+	g_timeout_add_seconds(5, exit_loop, NULL);
+
+	g_main_loop_run(main_loop);
+}
--- a/configure.ac
+++ b/configure.ac
@@ -225,8 +225,15 @@ if test x$enable_systemd = xyes; then
 		    [has_systemdsystemunitdir=$with_systemdsystemunitdir],
 		    [has_systemdsystemunitdir=$($PKG_CONFIG --variable=systemdsystemunitdir systemd)])
 	AC_SUBST([systemdsystemunitdir], [$has_systemdsystemunitdir])
+	AC_ARG_WITH([systemduserunitdir],
+		    AS_HELP_STRING([--with-systemduserunitdir=DIR], [Directory for systemd user service files]),
+		    [has_systemduserunitdir=$with_systemduserunitdir],
+		    [has_systemduserunitdir=$($PKG_CONFIG --variable=systemduserunitdir systemd)])
+	AC_SUBST([systemduserunitdir], [$has_systemduserunitdir])
+	AC_DEFINE(HAVE_SYSTEMD,1,[Build systemd code])
 fi
 AM_CONDITIONAL(HAVE_SYSTEMD, [test -n "$has_systemdsystemunitdir"])
+AM_CONDITIONAL(HAVE_SYSTEMD, [test -n "$has_systemduserunitdir"])
 
 dnl ---------------------------------------------------------------------------
 dnl - Use systemd offline update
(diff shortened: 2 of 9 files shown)


julian-klode:
I think it would be wise to make the .socket unit remove the socket if it is stopped, so we don't end up in a situation where the socket exists but is not usable.

hughsie:
@robert-ancell -- can you review these please? Thanks.

dantti:
> This change is incompatible with the debconf-kde hacks, I'm not sure what to do about those yet.

If this breaks debconf-kde we need a fix before merging it.

julian-klode:
> > This change is incompatible with the debconf-kde hacks, I'm not sure what to do about those yet.
> 
> If this breaks debconf-kde we need a fix before merging it.

This prevents packagekit-glib2 from using debconf-kde. I'm not sure anybody is actually testing this. I think debconf-kde needs to gather the ability to launch with an open socket / socket activation, and then pk-debconf-helper could just exec to it if it does.

dantti:
alright, I thought the change would break it when running like it's done in Apper

julian-klode:
So, looking at this again after the Christmas break:

This change also breaks if we are running in a terminal, because we'd suddenly be talking to a systemd-activated frontend which does not have our tty. That said, IIRC, it did not work for me anyway, but I'll probably have to reverify.

We do have a fallback to the current way currently in case debconf-socket does not exist. We can easily extend that check to also use the current way if we are running on kde or in a terminal (`DEBIAN_FRONTEND={kde,dialog}` set by `pk_client_create_helper_argv_envp()`).

This also avoids any possible new breakage regarding KDE.

configure.ac
```diff
@@ -216,10 +216,16 @@ if test x$enable_systemd = xyes; then
 		    AS_HELP_STRING([--with-systemdsystemunitdir=DIR], [Directory for systemd service files]),
 		    [has_systemdsystemunitdir=$with_systemdsystemunitdir],
 		    [has_systemdsystemunitdir=$($PKG_CONFIG --variable=systemdsystemunitdir systemd)])
-	AC_DEFINE(HAVE_SYSTEMD,1,[Build systemd code])
 	AC_SUBST([systemdsystemunitdir], [$has_systemdsystemunitdir])
+	AC_ARG_WITH([systemduserunitdir],
+		    AS_HELP_STRING([--with-systemduserunitdir=DIR], [Directory for systemd user service files]),
+		    [has_systemduserunitdir=$with_systemduserunitdir],
+		    [has_systemduserunitdir=$($PKG_CONFIG --variable=systemduserunitdir systemd)])
+	AC_SUBST([systemduserunitdir], [$has_systemduserunitdir])
```
robert-ancell: This line is duplicated.

backends/aptcc/pk-debconf-helper.c
```diff
@@ -0,0 +1,54 @@
+#include "config.h"
+#include <stdio.h>
+#include <packagekit-glib2/pk-client.h>
+#include <packagekit-glib2/pk-client-helper.h>
+#include <systemd/sd-daemon.h>
+
+static GMainLoop *main_loop;
+static PkClientHelper *helper;
+
+gboolean exit_loop(gpointer user_data)
```
robert-ancell: Nit: this should probably be static.

lib/packagekit-glib2/pk-client.c
```diff
@@ -1965,15 +1964,20 @@ pk_client_create_helper_socket (PkClientState *state)
...
 
 	/* no supported frontends available */
 	if (!ret)
 		return NULL;
 
+	socket_filename = g_build_filename(g_get_user_runtime_dir(), "debconf-socket", NULL);
+	if (g_file_test(socket_filename, G_FILE_TEST_EXISTS))
+		goto out;
+
+	g_free(socket_filename);
```
robert-ancell: This is a bit dangerous - socket_filename is now pointing to invalid memory. You should make use a new variable name for the debconf-socket string to avoid a bug being introduced in the future. The other option would be to use `g_clear_pointer (&socket_filename, g_free)` to set it to NULL (I still think multiple variables is clearer and safer).

backends/aptcc/pk-debconf-helper.c
```diff
@@ -0,0 +1,54 @@
...
+{
+	g_debug("Checking for active connections");
```
robert-ancell: This file doesn't match the existing code in having space between the function name and parenthesis.

robert-ancell:
I don't know enough detail about the intended behaviour with KDE but the code looks otherwise fine aside from the small notes I made.

hughsie:
@julian-klode can you fix up the nits please, then it LGTM. I'm planning a release for next week so if it gets merged before then it'll be in the tarball. Thanks.

julian-klode:
At least not today, I'm afraid. Probably tomorrow.

backends/aptcc/pk-debconf-helper.c
julian-klode: It does match the rest (or most, I did not check it all) of the aptcc backend style, though.

julian-klode:
Left to do:

- [ ] Rename debconf socket to pk-debconf
- [ ] Add back support for debconf-kde
- [ ] Do not use socket-activated helper when running in terminal or under KDE

julian-klode:
OK, should be fine now. I cleaned things up a bit, and fixed a few style issues in the lib/ code. I did not change the style of the pk-debconf-helper in the aptcc backend directory to match the lib/ style, as it currently matches the aptcc style.

lib/packagekit-glib2/pk-client-helper.c
```diff
@@ -487,6 +483,50 @@ pk_client_helper_start (PkClientHelper *client_helper,
...
+	g_autoptr(GError) error_local = NULL;
+	g_autoptr(GSocketAddress) address = NULL;
+
+	g_return_val_if_fail (PK_IS_CLIENT_HELPER (client_helper), FALSE);
+	g_return_val_if_fail (socket != NULL, FALSE);
+	g_return_val_if_fail (argv != NULL, FALSE);
+	g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
+
+	/* make sure not been started before */
+	g_return_val_if_fail (priv->argv == NULL, FALSE);
```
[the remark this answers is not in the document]
julian-klode: Just copy/paste out of `pk_client_helper_start`. And no, it's not normal to call twice, it's only allowed to be called once, as we can only track state of one running client helper (and only need one).

lib/packagekit-glib2/pk-client-helper.c
```diff
@@ -487,6 +483,50 @@ pk_client_helper_start (PkClientHelper *client_helper,
...
+ * Since: 1.1.13
+ **/
+gboolean
+pk_client_helper_start_with_socket (PkClientHelper *client_helper,
+				    GSocket *socket,
+				    gchar **argv, gchar **envp,
+				    GError **error)
+{
+	gint fd;
+	PkClientHelperPrivate *priv = client_helper->priv;
```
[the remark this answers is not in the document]
julian-klode: Not going to change this. This should be handled in a different pull request as it's done all over the place.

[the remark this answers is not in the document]
julian-klode: I guess I'll just fix up the rest, submit a pull request for that and rebase this branch on top of that one.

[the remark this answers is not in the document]
julian-klode: Rebased on top of #324 with that fixed.

julian-klode:
All reported issues should be fixed now. This depends on (or well, includes) #324 which fixes up dereferencing client_helper before checking it.

julian-klode:
OK, after the refactoring something broke:

Mai 22 12:41:59 jak-t480s pk-debconf-help[8066]: pk_client_helper_start_with_socket: assertion 'priv->argv == NULL' failed

PR #295, opened 2018-11-29
aptcc: Ignore non-fatal warnings on refresh-cache

davidmhewitt:
Because of these few lines of code:

```
if (_error->PendingError() == false && _error->empty() == false) {
        // TODO this shouldn't
        show_errors(m_job, PK_ERROR_ENUM_GPG_FAILURE);
```

If warnings (but no errors) had been emitted by apt, they'd be turned into a GPG failure which causes a couple of issues:

- Any non-fatal issues during `refresh-cache` would cause the task to report failure. For example, misconfigured/unreachable repositories or even repositories without appstream information (when it was expected) were causing the whole task to report failure.
- Emitting GPG_FAILURE in a non-interactive context causes the task to automatically retry with untrusted software allowed. This fails, because PackageKit still inteprets any warnings/errors that come out of apt as GPG failures, so this gets stuck in an infinite loop.

However, since real GPG failures are reported as errors, not warnings by apt, it's not necessary to do it this way.

This PR ignores all of the non-error items from the `_error` list as they were considered non-fatal by the apt refresh. The rest are processed in `show_errors` which already had some provisions for ignoring 404 errors, which I've added another check to about release files, which usually follows a 404.

While this doesn't cover all the possible cases, it covers the common misconfigured repository issues (usually caused by wrong release name after system upgrade) and temporary issues with mirrors and just silently ignores these issues like apt on the command line does. I'll raise a separate issue about the GPG failure infinite loop.

### DIFF
--- a/backends/aptcc/apt-intf.cpp
+++ b/backends/aptcc/apt-intf.cpp
@@ -2073,9 +2073,9 @@ void AptIntf::refreshCache()
     }
 
     // missing repo gpg signature would appear here
-    if (_error->PendingError() == false && _error->empty() == false) {
-        // TODO this shouldn't
-        show_errors(m_job, PK_ERROR_ENUM_GPG_FAILURE);
+    // TODO check what other errors could remain here and ensure the right error code is emitted for each
+    if (_error->empty() == false) {
+        show_errors(m_job, PK_ERROR_ENUM_GPG_FAILURE, true, true);
     }
 }
 
--- a/backends/aptcc/apt-messages.cpp
+++ b/backends/aptcc/apt-messages.cpp
@@ -29,7 +29,7 @@
 
 using namespace std;
 
-bool show_errors(PkBackendJob *job, PkErrorEnum errorCode, bool errModify)
+bool show_errors(PkBackendJob *job, PkErrorEnum errorCode, bool errModify, bool ignoreWarnings)
 {
     stringstream errors;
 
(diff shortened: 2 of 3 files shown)


dantti:
LGTM any thoughts @ximion @julian-klode ?

davidmhewitt:
Is there anything I can clarify or improve with this PR to help move it along? It'd be good to get this in as I believe it fixes quite a few of the causes of PackageKit getting stuck in an infinite retry loop that I've seen in various places over the years.

julian-klode:
I'm not happy with hiding warnings. They are warnings, they should be taken seriously and fixed. Warnings do not happen just for fun.

I think the correct way to solve the issues would be to extend PackageKit to have the ability to show non-fatal warnings, rather than just dropping them.

julian-klode:
We have the same problem in python-apt, where we check for errors on return, and ignore warnings if there is no error (if there is an error, we still include the warning). That's better, but still suboptimal.

hughsie:
> to extend PackageKit to have the ability to show non-fatal warnings, rather than just dropping them

We used to have `Message()` and it was a disaster as backends abused this showing horrible messages that mere mortals couldn't understand. PackageKit just has to make it work without the user caring how.

dantti:
Yes, aptcc used to use Message() for warning, but when it got removed there were no other option...

hughsie:
> Well, we should at least log them to the journal.

Sure, agreed. I think all g_warning() statements go to the journal now, right?

julian-klode:
I believe it would, everything on stdout and stderr goes to the journal.

davidmhewitt:
Thanks all, I've added in a line to put all errors/warnings from apt through `g_warning` before any filtering is done.

The `show_errors` method throws away some safe to ignore error as well as warnings, so if we log everything, at least it can be found in the journal if necessary.

julian-klode:
I missed that before, but what I'd like to see is not ignoring warnings at all. If there is an error, we should also include warnings (and notices, for that matter) - they might be related (and notices certainly are, they provide more information for errors in quite a few cases).


So, what `show_errors` should do is count how many messages are errors. Then at the end, instead of checking `if(!errors.str().empty())` it should check `if(!error_count > 0)`. This also avoids the extra parameter.

julian-klode:
See #321

davidmhewitt:
My reasoning for adding an extra parameter was that `show_errors` is used in a considerable number of places and was unsure whether any of the other places were relying on certain warnings being thrown as errors, so I was making sure the behavior was the same when that method is used elsewhere.

I'm happy that your approach fixes some of what I was trying to fix, but the lines to search for the `"does not have a Release file"` error and ignore it were relevant to my fix too. This is an error that you commonly get out of the backend with misconfigured repositories (IMO dpkg should report it as a warning instead of an error).

However, these error lines will still be counted in your `errorCount` variable and will still end up throwing the GPG failure and get stuck in an infinite loop which is what I was trying to fix here. Is there a reason you ommited that part from your new patch?

Conan-Kudo:
Closing in favor of #321.

PR #299, opened 2018-12-12
Add elogind support

EvaSDK:
One of our downstream maintainers forgot to submit this for inclusion it seems so here it is.

Closes: https://bugs.gentoo.org/620948
Signed-off-by: Ettore Di Giacinto <mudler@gentoo.org>


hughsie:
@kalev looks pretty self contained to me, agree?

kalev:
Yep, the patch is definitely self-contained. Not sure it's a good idea to support systemd forks though, maybe it should stay a downstream patch instead ... @keszybz any suggestions if we should merge this?

keszybz:
The patch looks superficially OK. From systemd side, all I can say that it doesn't bother us. If anything, it's an additional maintanance / testing burden on the PackageKit side, but just a small one.

kalev:
Let's merge then. Thanks @keszybz!

PR #303, opened 2018-12-26
Remove pk_is_thread_default check in pk_backend_is_eula_valid

### LINKED ISSUE #300: Unable to call pk_backend_is_eula_valid() in backend methods
All the real actions are done in a new thread using pk_backend_job_thread_create() inside zypp backend, when we call pk_backend_is_eula_valid to check the EULA id, it fails saying it's not in the default thread, which is caused by this check.
> g_return_val_if_fail (pk_is_thread_default (), NULL);

This is introduced in 44bd3ab6.

I'd like to hear your recommendation on how to make such a pk_backend_is_eula_valid call. Should we simply disable that check in pk_backend_is_eula_valid?

JonathanKang:
pk_is_thread_default() check was introduced in commit 44bd3ab6 to
detect possible threading problems.

Some backends need call pk_backend_is_eula_valid() in a thread which is
not the default one to perform EULA related actions. This way
pk_is_thread_default() check fails.

Fix that by removing the pk_is_thread_default() check in
pk_backend_is_eula_valid().

Fixes #300

### DIFF
--- a/src/pk-backend.c
+++ b/src/pk-backend.c
@@ -189,6 +189,7 @@ struct PkBackendPrivate
 	gchar			*name;
 	gpointer		 file_changed_data;
 	GHashTable		*eulas;
+	GMutex			 eulas_mutex;
 	GModule			*handle;
 	PkBackendDesc		*desc;
 	PkBackendFileChanged	 file_changed_func;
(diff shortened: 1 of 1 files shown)


JonathanKang:
> Is there a mutex protecting priv->eulas?

Emm, right. There is not mutex lock protecting priv->eulas in pk-backend.c. We should use mutex lock to protect priv->eulas then.

hughsie:
> We should use mutex lock to protect priv->eulas then

Right. A new mutex would be fine.

JonathanKang:
Please review this if you have some free time. Thanks.

src/pk-backend.c
```diff
@@ -955,9 +958,10 @@ pk_backend_is_eula_valid (PkBackend *backend, const gchar *eula_id)
 
 	g_return_val_if_fail (PK_IS_BACKEND (backend), FALSE);
 	g_return_val_if_fail (eula_id != NULL, FALSE);
-	g_return_val_if_fail (pk_is_thread_default (), FALSE);
 
+	g_mutex_lock (&backend->priv->eulas_mutex);
```
hughsie: Could we use GMutexLocker for all these cases?

PR #305, opened 2018-12-28
zypp: Switch to doUpgrade solver when required by distribution

hellcp:
Not the prettiest solution, but since zypper checks for `dup` argument the same way when notifying the user about proper update command, this is the most distro-agnostic way I could do this.

### DIFF
--- a/backends/zypp/pk-backend-zypp.cpp
+++ b/backends/zypp/pk-backend-zypp.cpp
@@ -3297,6 +3297,14 @@ backend_update_packages_thread (PkBackendJob *job, GVariant *params, gpointer us
 	ResPool pool = zypp_build_pool (zypp, TRUE);
 	PkRestartEnum restart = PK_RESTART_ENUM_NONE;
 
+	for ( const PoolItem & pi : pool.byKind<Product>() ) {
+		static const Capability indicator( "product-update()", Rel::EQ, "dup" );
+		if ( pi->asKind<Product>()->referencePackage().provides().matches( indicator ) ) {
+			zypp->resolver()->setUpgradeMode(TRUE);
+			break;
+		}
+	}
+
 	PoolStatusSaver saver;
 
 	for (guint i = 0; package_ids[i]; i++) {


kalev:
@DimStar77 @JonathanKang Can you review please?

JonathanKang:
@hellcp Is this meant to use `zypper dup` like action while performing `pkcon update`?

hellcp:
@JonathanKang Yeah, basically

JonathanKang:
looks good to me.

@mlandres do you have any comment on this?

backends/zypp/pk-backend-zypp.cpp
```diff
@@ -1454,6 +1454,20 @@ zypp_perform_execution (PkBackendJob *job, ZYpp::Ptr zypp, PerformType type, gbo
...
 
 		ResPool pool = ResPool::instance ();
+
+		// Products labeled as `dup` for update need to have solver switched to doUpgrade
+		if (type == UPDATE) {
+			for (ResPool::const_iterator it = pool.begin (); it != pool.end (); ++it) {
+				if ( it->isKind<Product>() ) {
+					static const Capability indicator( "product-update()", Rel::EQ, "dup" );
+					if ( it->operator->()->asKind<Product>()->referencePackage().provides().matches( indicator ) ) {
+						zypp->resolver()->setUpgradeMode(TRUE);
```
mlandres: Looks to me like `resolvePool ()` was already called at L1406, so switching the upgrade mode here will affect the next solver run, not the current one. Not sure if this is what you intended.

Conan-Kudo: But the transaction hasn't been run yet at this point, so shouldn't it affect the currently developing transaction?

mlandres: Not immediately.  Libzypp basically allows to manip pool items directly and to commit without running the solver. 

Changing `Resolver` flags or solver jobs does not immediately change the status of the items in the `ResPool`. You need to call `resolvePool()` to run the solver and apply the result to the pool items (i.e. updating their status).

backends/zypp/pk-backend-zypp.cpp
```diff
@@ -3297,6 +3297,15 @@ backend_update_packages_thread (PkBackendJob *job, GVariant *params, gpointer us
 	ResPool pool = zypp_build_pool (zypp, TRUE);
 	PkRestartEnum restart = PK_RESTART_ENUM_NONE;
 
+	for (ResPool::const_iterator it = pool.begin (); it != pool.end (); ++it) {
+		if ( it->isKind<Product>() ) {
```
mlandres: Basically right. As you have some 1000 packages and just a few products you should `break;`  after having found `setUpgradeMode(TRUE)`. You may also use the pools internal filtering to iterate over the products only:
```
  for (ResPool::byKind_iterator it = pool.byKindBegin<Product>(); it != pool.byKindEnd<Product>(); ++it) {
    static const ...
    if ( (*it)->asKind<Product>()...
```
Or range based
```
  for ( const PoolItem & pi : pool.byKind<Product>() ) {
    static const ...
    if ( pi->asKind<Product>()...
```

Conan-Kudo:
@hellcp Could you please squash the commits into a single commit for merging?

sysrich:
+1 from me..it would be really really cool if we could get this merged and finally get this solved in openSUSE :heart:

kalev:
Thanks, let's merge then.

PR #307, opened 2019-01-24
Add meson build system

tintou:
This is a WIP branch of a meson port which might help to maintain PackageKit as it aligns it with the ~GNOME~ GLib libraries standard


tintou:
Also here so that you can stop me if you're reluctant to this change before I continue :smile:

Conan-Kudo:
I personally don't mind the change from autofoo to Meson, but it'd be nice if it was more complete... 👍 

@hughsie What do you think?

kkofler:
Why not CMake? Why are GNOME standards relevant for a shared freedesktop.org project hosted on third-party infrastructure (GitHub)?

Conan-Kudo:
@kkofler While I personally have no problem with CMake, I think @hughsie doesn't like CMake very much.

But if a proposal were made for CMake, I'd at least be open to reviewing it.

tintou:
I should probably have said GLib-based library instead of GNOME, I'm not opening a build-system contest here, I do not intend to port to CMake so that's my proposition.

dantti:
Just for the record I did a port in the past to CMake but Richard didn't like the idea as it was not a tool he was used to, but IMHO anything is better than autohell so I wouldn't mind if this goes into Meson direction, a shame Gnome didn't go with CMake now that even Qt is going to drop qmake and better support CMake

hughsie:
@tintou Great work. This has been on my TODO for so long. How much more is left to complete?

hughsie:
@kkofler no CMake, sorry. It's horrible in my opinion compared to Meson. Lots of other fd.o projects (xorg, wayland, fwupd etc) are converting or have converted to Meson so I don't see any problem with that at all.

tintou:
@hughsie So I've done a big chunk, I need to port `contrib` and `docs` folder. I'll probably need some help to test all the backends… I've probably missed a few things here and there but it's already in a good shape.

meson.build
```diff
@@ -0,0 +1,141 @@
...
+	get_option('systemd'),
+	get_option('offline-update'),
+	get_option('gstreamer-plugin'),
+	get_option('gtk-module'),
+	get_option('gobject-introspection'),
+	get_option('bash-command-not-found'),
+	get_option('cron'),
+	get_option('python-3'),
+	get_option('backend-name'),
+	get_option('python-backend'),
```
Conan-Kudo: `backend-name` and `python-backend` need to be swapped here. Also, there's nothing that will make it print the selected DNF vendor if the DNF backend is chosen?

tintou: @Conan-Kudo Made it adaptive to the backend and added dnf vendor message in case the dnf backend is enabled

tintou:
So I just added the `contrib` and `docs` folders, so I guess that it is ready for review now

tintou:
I made no change to the original files so that Autotools and Meson can coexist during the transition.
After the _presumed_ removal of Autotools support, it might be good to switch to a gettext only backend for instance (and thus remove intltool dependency)

hughsie:
> I made no change to the original files so that Autotools and Meson can coexist during the transition.

I've tried this with other packages and it was a disaster as one build system was always getting out of sync. You have my full blessing to delete all the autotools stuff!

> it might be good to switch to a gettext only backend for instance (and thus remove intltool dependency)

Nuking intltool now is the right thing to do.

hughsie:
@tintou Also, I've made you a collaborator now, so if you can push your branch to this repo (something like wip/tintou/meson) then we can just pile fixup patches on top and squash before merging.

tintou:
Great, continuing in #309 then

PR #310, opened 2019-01-29
dnf: Use system-release(releasever) rpmdb provides for release version

kalev:
This is completely untested beyond that it compiles. @kparal, do you want to give it a try?

https://pagure.io/releng/issue/7445


kparal:
Ooh, cool! I'll try this out.

kparal:
Hmm, it doesn't seem to work. I applied the patch on top of current `PackageKit-1.1.12-2.fc30` in Rawhide, is it necessary to use the actual `master` branch instead?

I build modified `fedora-release` and `PackageKit` in this copr:
https://copr.fedorainfracloud.org/coprs/kparal/rawhide-releasever/packages/

In a fully updated Fedora Workstation Rawhide, I updated to those packages. The new provision is now available to all `fedora-release-VARIANT` subpackages (because they recently split them into subpackages), in my case the workstation one:
```
# rpm -q --provides fedora-release-workstation | grep system-release
system-release
system-release(30)
system-release(releasever) = rawhide
system-release-product
```

I modified `fedora-rawhide.repo` to contain this:
```
[rawhide]
name=Fedora - Rawhide - Developmental packages for the next Fedora release
failovermethod=priority
baseurl=http://download.fedoraproject.org/pub/fedora/linux/development/$releasever/Everything/$basearch/os/
#metalink=https://mirrors.fedoraproject.org/metalink?repo=rawhide&arch=$basearch
enabled=1
metadata_expire=6h
repo_gpgcheck=0
type=rpm
gpgcheck=1
gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-$releasever-$basearch
skip_if_unavailable=False
```
(I forced the use `baseurl` and used `$releasever` inside instead of the hardcoded `rawhide` in the URL).

DNF works fine with this, it replaces `$releasever` with `rawhide`, as expected:
```
# dnf repolist --repo rawhide -v | grep baseurl
Repo-baseurl : http://download.fedoraproject.org/pub/fedora/linux/development/rawhide/Everything/x86_64/os/
```

But PackageKit doesn't. I was trying to find a way to figure out which URL it accesses, and I didn't, but it at least prints an error when trying to access a non-existent URL:
```
# pkcon -v -p refresh force
...
Fatal error: cannot update repo 'rawhide': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried; Last error: Status code: 404 for http://download.fedoraproject.org/pub/fedora/linux/development/30/Everything/x86_64/os/repodata/repomd.xml
```

It replaced `$releasever` with `30` (old behavior). So this doesn't seem to work yet.

kalev:
OK, I tested it as well now and it turned out to be a missing rpm global init: rpm was complaining that it couldn't find its database path.

Anyway, looking at it some closer, libdnf's DnfContext already has all the "system-release(releasever)" logic and we can easily avoid reimplementing it in packagekit. I've pushed a completely different patch now that's just a trivial two line change; can you give it a try please?

kalev:
Also, when I tested it a got a lot of errors about not finding GPG keys: the gpgkey= values in .repo files contain $releasever and the gpg keys shipped in fedora-gpg-keys need updating to have rawhide instead of 30 in the file names (or both names).

kparal:
Great, it works! Thanks for the heads up about the GPG keys, after I fixed that in my COPR repo and used the new patch for PackageKit, it correctly translates `$releasever` to `rawhide`. It also works fine with an unmodified system (translating `$releasever` to `30`).

Can this be merged?

kparal:
Thanks. Once you do a new release and push it to Rawhide in the future, I'll follow up with the releng ticket.

PR #315, opened 2019-02-25
aptcc: also generate a repo detail when a download happened (and finished)

apachelogger:
the previous code paths only covered

- Hit (aka the server responded with not-modified or somesuch code)
- Ign (aka the server 404s or somesuch indicating the repo doesn't exist)

but ignored

- Get (aka code 200 and data was downloaded)

this resulted in repos which actually had updates available (hence the
file change and need for download) not being reported in `pkcon refresh`
output at all.

::Done acquire updates now also cause a repo detail to get added so
pkcon can show the repo as [enabled] when running `pkcon refresh`.

### DIFF
--- a/backends/aptcc/acqpkitstatus.cpp
+++ b/backends/aptcc/acqpkitstatus.cpp
@@ -79,6 +79,13 @@ void AcqPackageKitStatus::Fetch(pkgAcquire::ItemDesc &Itm)
 /* We don't display anything... */
 void AcqPackageKitStatus::Done(pkgAcquire::ItemDesc &Itm)
 {
+    PkRoleEnum role = pk_backend_job_get_role(m_job);
+    if (role == PK_ROLE_ENUM_REFRESH_CACHE) {
+        pk_backend_job_repo_detail(m_job,
+                                   "",
+                                   Itm.Description.c_str(),
+                                   true);
+    }
     // Download completed
     updateStatus(Itm, 100);
 }


apachelogger:
@dantti

dantti:
Does this works fine for GUI too?

apachelogger:
I don't see why it wouldn't. All this does is add additional repo details entries to the result. If a GUI rendered the previous details it would have suffered from the same defect as pkcon and get fixed by this PR (i.e. the previous list was incomplete and the new one is complete).

backends/aptcc/acqpkitstatus.cpp
```diff
@@ -79,6 +79,13 @@ void AcqPackageKitStatus::Fetch(pkgAcquire::ItemDesc &Itm)
 /* We don't display anything... */
 void AcqPackageKitStatus::Done(pkgAcquire::ItemDesc &Itm)
 {
+    PkRoleEnum role = pk_backend_job_get_role(m_job);
+    if (role == PK_ROLE_ENUM_REFRESH_CACHE) {
+        pk_backend_job_repo_detail(m_job,
+                                   "",
```
hughsie: I think you want a repo ID here...

apachelogger: Yeah, technically... unfortunately we don't have easy access to it. So I left that intentionally broken same as in `AcqPackageKitStatus::IMSHit`.

What apt sends there are basically protocol requests succeeding/failing, but in the one minute I spent digging through apt-pkg's API I didn't find a way to get from there to the information we need to build a repo id.

There **may** be a way to infer it (break URI apart, compare to possible URI parts a repo might feasibly have). But it's out of scope for what I actually want to fix: the current detail list isn't even complete.

dantti:
Yeah, repo-id isn't easy to get AFAIR.

PR #316, opened 2019-02-27
offline update: Use new plymouth "system-upgrade" and "reboot" modes

kalev:
These landed in plymouth git yesterday to implement
https://wiki.gnome.org/Design/OS/BootProgress

Using the new API conditionally is tricky as it's command line API;
enterprise distros that don't have new enough plymouth should probably
just revert this commit.

### DIFF
--- a/client/pk-offline-update.c
+++ b/client/pk-offline-update.c
@@ -188,7 +188,11 @@ pk_offline_update_reboot (void)
 
 	/* reboot using systemd */
 	sd_journal_print (LOG_INFO, "rebooting");
+#ifdef PLYMOUTH_0_9_5
+	pk_offline_update_set_plymouth_mode ("reboot");
+#else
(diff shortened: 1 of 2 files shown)


hughsie:
Mrr, that sounds like it would break things for some people, in unexpected ways. Can't we have one more meson option (defaulting to on) and add some #ifdefs?

kalev:
Yeah, that's probably a good idea :) I'll look into the ifdefs next week though, not now.

Conan-Kudo:
@kalev Have you had a chance to look into finishing this? This patch shipped in Fedora, and it'd be nice to get this properly upstreamed.

kalev:
No, I haven't.

kalev:
OK, I've updated the PR now to make this conditional depending on the installed plymouth version.

hughsie:
If CI likes it, LGTM, thanks.

kalev:
CI is green!

PR #317, opened 2019-03-01
aptcc: Implement frontend-locking

julian-klode:
This implements APT frontend locking, which prevents other programs
from stealing the dpkg lock in race conditions where apt(cc) has to
release the lock in order to run dpkg.

Bug-Ubuntu: https://bugs.launchpad.net/bugs/1795614


julian-klode:
We've been running this in Ubuntu stable releases since late October (and in cosmic since August), and it works fine.

ximion:
Does this need a libapt version requirement bump?

julian-klode:
I added a separate check for frontend locking, in case there is a backport for trusty for example.

ximion:
Thank you for the patch!

PR #320, opened 2019-03-14
[zypp] Perform actions disallowed by update in upgrade mode

hellcp:
Some fixes to how upgrades are handled, there really is not much point in removing packages from pool for upgrade action. Also to remain the same as zypper in calculated solution, added check for vendor change configuration.


Conan-Kudo:
@hellcp Could you please squash your changes into a single commit?

PR #323, opened 2019-05-05
dnf: Replace armv5tl with aarch64 for Mageia repo configuration vendor

Conan-Kudo:
With Mageia 7, armv5tl has been removed and aarch64 has been introduced.
Thus, we want to make sure aarch64 repositories are marked as supported
and remove armv5tl from the supported arches list.


kalev:
Looks good to me, thanks.

PR #324, opened 2019-05-20
pk-client-helper: Do not dereference argument before checking it

julian-klode:
Stop dereferencing client_helper before validating it in public
functions.

### DIFF
--- a/lib/packagekit-glib2/pk-client-helper.c
+++ b/lib/packagekit-glib2/pk-client-helper.c
@@ -135,10 +135,12 @@ pk_client_helper_child_free (PkClientHelperChild *child)
 gboolean
 pk_client_helper_stop (PkClientHelper *client_helper, GError **error)
 {
-	PkClientHelperPrivate *priv = client_helper->priv;
+	PkClientHelperPrivate *priv = NULL;
 
 	g_return_val_if_fail (PK_IS_CLIENT_HELPER (client_helper), FALSE);
 	g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
+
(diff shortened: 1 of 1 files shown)


lib/packagekit-glib2/pk-client-helper.c
```diff
@@ -135,11 +135,13 @@ pk_client_helper_child_free (PkClientHelperChild *child)
 gboolean
 pk_client_helper_stop (PkClientHelper *client_helper, GError **error)
 {
-	PkClientHelperPrivate *priv = client_helper->priv;
+	PkClientHelperPrivate *priv = NULL;
```
hughsie: Would pk_client_helper_get_instance_private() work here? i.e. does it ignore NULL?

julian-klode: Seems to be. Then we also don't really need ->priv anymore.

https://developer.gnome.org/gobject/stable/gobject-Type-Information.html says:
> It is safe to call _get_instance_private on NULL or invalid object since it's only adding an offset to the instance pointer. In that case the returned pointer must not be dereferenced.

julian-klode:
So this was a nice idea, but `pk_client_helper_get_instance_private` does not work.Probably needs the private part be regustered differently.

kalev:
Can you fix the conflicts, please? Otherwise looks good to me.

Conan-Kudo:
@julian-klode ping? Could you please address @kalev's request?

julian-klode:
There's nothing to address here, this change has been merged already.

PR #328, opened 2019-06-14
Port to APT 1.9

julian-klode:
Port aptcc backend to APT 1.9


julian-klode:
Note that 1.9.x is somewhat of a moving target, and there might be another break in a 1-1.5 months or so, but hopefully nothing breaks for PackageKit anymore. Well, the CurrentItem stuff might, as we might want to have CurrentItems instead to prepare for HTTP/2.

1.9.0 was mostly a remove deprecated stuff thing.

ximion:
Looks good to me!

PR #329, opened 2019-06-19
zypp: Emit installed package for newest filter

JonathanKang:
Previously, zypp backend only emit newest package in
backend_resolve_thread(). The newest filter processes installed
and available package lists separately, so the installed package should
be emitted as well.

https://bugzilla.suse.com/show_bug.cgi?id=1137019


JonathanKang:
moved to #345.

PR #330, opened 2019-07-19
Do not trigger an inotity event when the AppStream XML data is unchanged

hughsie:
We already call dnf_utils_refresh_repo_appstream() on startup, but only as
required. Remove the unconditional copy to speed up gnome-software startup.

### DIFF
--- a/backends/dnf/pk-backend-dnf.c
+++ b/backends/dnf/pk-backend-dnf.c
@@ -565,13 +565,6 @@ dnf_utils_add_remote (PkBackendJob *job,
 	if (!ret)
 		return FALSE;
 
-	/* update the AppStream copies in /var */
-	for (guint i = 0; i < repos->len; i++) {
-		DnfRepo *repo = g_ptr_array_index (repos, i);
-		if (!dnf_utils_refresh_repo_appstream (repo, error))
-			return FALSE;
-	}
-
 	/* done */
 	if (!dnf_state_done (state, error))
 		return FALSE;


hughsie:
@kalev I'd appreciate it if you could put this in a F30 build and see if it fixes the slow startup issue in gnome-software for you too. Thanks.

kalev:
> @kalev I'd appreciate it if you could put this in a F30 build and see if it fixes the slow startup issue in gnome-software for you too. Thanks.

Sorry, no, I'll be gone and on vacation for the next three weeks. Also, I'm not sure what the slow startup issue is, haven't run into it here so not the best person to test.

backends/dnf/pk-backend-dnf.c
```diff
@@ -565,13 +565,6 @@ dnf_utils_add_remote (PkBackendJob *job,
 	if (!ret)
 		return FALSE;
 
-	/* update the AppStream copies in /var */
-	for (guint i = 0; i < repos->len; i++) {
-		DnfRepo *repo = g_ptr_array_index (repos, i);
-		if (!dnf_utils_refresh_repo_appstream (repo, error))
-			return FALSE;
-	}
-
```
kalev: So this deletes the `dnf_utils_refresh_repo_appstream()` calls from add_remote(). The only other place where it's called from is `pk_backend_refresh_cache_thread` -> `pk_backend_refresh_repo` ->  `dnf_utils_refresh_repo_appstream`. Not sure if we are hitting that path all the time? I think gnome-software may explicitly call `pk_client_refresh_cache`, but not sure if all other frontends do.

hughsie: >  dnf_utils_refresh_repo_appstream

This is only "hit" when we download new metadata, which of course is fine to write a new appstream file.

> I think gnome-software may explicitly call pk_client_refresh_cache, but not sure if all other frontends do.

I guess maybe we could test if the file already exists? It does seem a bit of a layering violation in that case as libappstream-glib would need to leak out the appstream system path. I'm also not sure how much I care about other front-ends...

hughsie:
> I'll be gone and on vacation for the next three weeks

Aha, then I'm just being lazy! Enjoy your vacation, and I'll do a Fedora build in rawhide and see if it fixes the issues various people were seeing.

PR #331, opened 2019-07-28
Refresh Appdata cache in zypp backend

pontaoski:
The zypp backend will now attempt to refresh the appdata cache when refreshing repositories.

### DIFF
--- a/backends/zypp/pk-backend-zypp.cpp
+++ b/backends/zypp/pk-backend-zypp.cpp
@@ -36,6 +36,7 @@
 #include <sys/vfs.h>
 #include <unistd.h>
 #include <vector>
+#include <appstream-glib.h>
 
 #include <glib.h>
 #include <glib/gi18n.h>
@@ -79,6 +80,7 @@
 #include <zypp/parser/IniDict.h>
 #include <zypp/parser/ParseException.h>
 #include <zypp/parser/ProductFileReader.h>
+#include <zypp/parser/xml/Reader.h>
 #include <zypp/repo/PackageProvider.h>
 #include <zypp/repo/RepoException.h>
 #include <zypp/repo/SrcPackageProvider.h>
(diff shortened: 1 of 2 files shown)


backends/zypp/pk-backend-zypp.cpp
```diff
@@ -154,10 +156,46 @@ guint _dl_count = 0;
...
+	std::string metadataPathString = metadataPath.c_str();
+	xml::Reader reader(metadataPathString + "/repomd.xml");
+
+	while (!reader.atEnd())
+	{
+		reader.nextNode();
+		std::string location = reader->getAttribute("href").c_str();
+		if (location.find("appdata.xml.gz") != std::string::npos ||
+		    location.find("appdata-icons.xml.gz") != std::string::npos ||
+		    location.find("primary.xml.gz") != std::string::npos)
```
Conan-Kudo: Doesn't this logic mean that it will also attempt this if `primary.xml.gz` exists but not the other two? That obviously would be quite broken...

hughsie:
It'd be easier to review if it was squashed down into one commit and force pushed too. Thanks.

backends/zypp/pk-backend-zypp.cpp
Appadeia: I didn't notice any errors in testing with this logic. The repositories with appdata cached theirs just fine and the other repositories didn't cause any errors.

hellcp: I would suggest going for that anyway, get the hrefs into an array and iterate over it to install, checking if all the required files are there.

We also probably should change the names of those files in our repos so the backend works with Fedora and other dnf based distros, as well as dnf work on openSUSE and other zypp based distros.

hellcp:
Paging @DimStar77 who is responsible for the zypp appdata plugin. It would be cool to get distributions in line with naming of appstream files, so we can use dnf and zypp on respective distros. As you can see in [the dnf backend source](https://github.com/hughsie/PackageKit/blob/master/backends/dnf/pk-backend-dnf.c#L508), dnf uses `appstream` and `appstream-icons` names, we use `appdata` and `appdata-icons`, how would we go about changing that so both are consistent?

DimStar77:
IMHO, the logic to refresh appdata metadata belongs into libzypp - which PK relies already upon to do the right thing.

Conan-Kudo:
@DimStar77 IIRC, libzypp used to do this, but then removed that facility a few years ago. At this point, the AppStream fetching is broken in the zypp stack.

DimStar77:
there is libzypp-plugin-appdata that is called by libzypp on repo refresh - and this is still working (short of an error in appstream-glib, which installed the .xml files with wrong permissions)

hellcp:
Does it belong in zypp though? There is no particular reason to install appdata without packagekit present, there are no tools that rely on appdata without packagekit afaik.

DimStar77:
It's repository meta data - and as such is best refreshed / updated together whenever a repo is updated. And it must not matter if the refresh was done using pkcon/PackageKit or any other zypp-frontend (e.g. zypper). In any case, frontends MUST get the up-do-date information. Even things like ```appstream-util search``` - which is NOT a PK frontend, but interacts with the appstream metadata directly. So yes, it DOES belong to the lowest common denominator - which is libzypp

PR #332, opened 2019-08-08
Port to meson build system #2

t-chaik:
https://github.com/hughsie/PackageKit/pull/309 follow-up.


hughsie:
Wow, do you want to squash all these down and then I can do some proper review? Thanks.

t-chaik:
> Wow, do you want to squash all these down and then I can do some proper review? Thanks.

Do you want me to squash everything in let say 3 commits (meson port, autotools removal and CI update) or just my new ones in jurst one?

t-chaik:
Tests from master (autotools) are failing on my machine too (`make check`). Correct me if I'm wrong, but it seems like current CI is not running them but only perform build + install. For what I can tell, they are not failing for environment/build reasons and fixing them would require jumping in PackageKit's code and understand that's going wrong. I've drop them from the meson CI for now, please advice on what you'd prefer here.

hughsie:
I think disable the failing self tests for now, I think the meson patch is more important. When it's all passing CI I can fix up the tests as required. Thanks again!

docs/api/dbus/meson.build
```diff
@@ -0,0 +1,25 @@
...
+packagekit_ref_xml = custom_target(
+	'org.freedesktop.PackageKit.ref.xml',
+	input: join_paths(meson.source_root(), 'src', 'org.freedesktop.PackageKit.xml'),
+	output: 'org.freedesktop.PackageKit.ref.xml',
+	command: [
+		spec_to_docbook,
+		files('spec-to-docbook.xsl'),
+		'@INPUT@',
+		'@OUTPUT@',
+	],
```
crawfxrd: Adding `build_by_default: true` to these custom targets will allow `ninja dist` to work.

t-chaik:
Shall I squash all this in 2 or 3 commits (+meson, -autotools, +CI)?

hughsie:
Ya, squash please.

hughsie:
I'm just fixing up some of the failing tests this morning.

hughsie:
I've fixed up a lot of the issues, so lets continue in https://github.com/hughsie/PackageKit/pull/365 -- all comments there very welcome, thanks!

PR #341, opened 2019-10-03
aptcc: improve support of obsoletes during updates

darktemplarbasealt:
Output of 'pkcon get-updates' command is changed to reflect more closely what would actually be updated, installed, removed or obsoleted during update, but it may require to modify users of this command to not just install everything from output of this command, including removed and obsoleted packages, similarly to how 'pkcon update' is updated in these changes.

### DIFF
--- a/client/pk-console.c
+++ b/client/pk-console.c
@@ -1112,6 +1112,13 @@ pk_console_update_packages (PkConsoleCtx *ctx, gchar **packages, GError **error)
 }
 
 static gboolean
+pk_console_update_system_filter_helper (PkPackage *package, gpointer user_data)
+{
+	PkInfoEnum package_enum = pk_package_get_info (package);
+	return (package_enum != PK_INFO_ENUM_OBSOLETING && package_enum != PK_INFO_ENUM_REMOVING);
+}
+
+static gboolean
 pk_console_update_system (PkConsoleCtx *ctx, GError **error)
 {
 	g_autoptr(PkPackageSack) sack = NULL;
@@ -1130,6 +1137,7 @@ pk_console_update_system (PkConsoleCtx *ctx, GError **error)
 
 	/* do the async action */
 	sack = pk_results_get_package_sack (results);
+	pk_package_sack_remove_by_filter (sack, &pk_console_update_system_filter_helper, NULL);
 	package_ids = pk_package_sack_get_ids (sack);
 	if (g_strv_length (package_ids) == 0) {
 		pk_progress_bar_end (ctx->progressbar);
--- a/backends/aptcc/apt-intf.cpp
+++ b/backends/aptcc/apt-intf.cpp
@@ -852,10 +852,23 @@ void AptIntf::emitUpdateDetail(const pkgCache::VerIterator &candver)
     bugzilla_urls = getBugzillaUrls(changelog);
     cve_urls = getCVEUrls(changelog);
 
+    GPtrArray *obsoletes = g_ptr_array_new();
+
+    for (auto deps = candver.DependsList(); not deps.end(); ++deps)
+    {
+        if (deps->Type == pkgCache::Dep::Obsoletes)
+        {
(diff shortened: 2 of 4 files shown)


dantti:
This is something I wanted to do a long time ago, I didn't test but looks fine, but it would be interesting is @hughsie could review the changes made outside the backend.

lib/packagekit-glib2/pk-package-sack.h
```diff
@@ -106,6 +106,7 @@ typedef gboolean (*PkPackageSackFilterFunc)		(PkPackage		*package,
 /* managing the array */
 void		 pk_package_sack_clear			(PkPackageSack		*sack);
 gchar		**pk_package_sack_get_ids		(PkPackageSack		*sack);
+gchar		**pk_package_sack_get_ids_without_removals		(PkPackageSack		*sack);
```
hughsie: I don;t think we want to export this function

client/pk-console.c
```diff
@@ -1130,7 +1130,7 @@ pk_console_update_system (PkConsoleCtx *ctx, GError **error)
 
 	/* do the async action */
 	sack = pk_results_get_package_sack (results);
-	package_ids = pk_package_sack_get_ids (sack);
+	package_ids = pk_package_sack_get_ids_without_removals (sack);
```
hughsie: I think the filtering needs to be done here

darktemplarbasealt: I've found 'pk_package_sack_remove_by_filter' function. I'll rework it using this function.

darktemplarbasealt:
I've tested changes similar to these with apt-rpm (fork of apt, based on pretty old version of apt) from ALT:

http://git.altlinux.org/gears/p/packagekit.git?p=packagekit.git;a=shortlog;h=refs/heads/sisyphus

But I didn't test it with apt from Debian or Ubuntu.

darktemplarbasealt:
Any feedback on this change?

client/pk-console.c
```diff
@@ -1111,6 +1111,16 @@ pk_console_update_packages (PkConsoleCtx *ctx, gchar **packages, GError **error)
...
 }
 
+static gboolean
+pk_console_update_system_filter_helper(PkPackage *package, gpointer user_data)
+{
+	PkInfoEnum package_enum;
+
+	package_enum = pk_package_get_info(package);
+
+	return ((package_enum != PK_INFO_ENUM_OBSOLETING) && (package_enum != PK_INFO_ENUM_REMOVING));
```
hughsie: I think you can remove the inner bracket here.

client/pk-console.c
```diff
@@ -1111,6 +1111,16 @@ pk_console_update_packages (PkConsoleCtx *ctx, gchar **packages, GError **error)
 	return TRUE;
 }
 
+static gboolean
+pk_console_update_system_filter_helper(PkPackage *package, gpointer user_data)
```
hughsie: space after bracket

client/pk-console.c
```diff
@@ -1111,6 +1111,16 @@ pk_console_update_packages (PkConsoleCtx *ctx, gchar **packages, GError **error)
...
+{
+	PkInfoEnum package_enum;
+
+	package_enum = pk_package_get_info(package);
```
hughsie: space after braket, and can probably be moved up to the `PkInfoEnum` line making the entire function only two lines.

client/pk-console.c
```diff
@@ -1130,6 +1140,7 @@ pk_console_update_system (PkConsoleCtx *ctx, GError **error)
 
 	/* do the async action */
 	sack = pk_results_get_package_sack (results);
+	pk_package_sack_remove_by_filter(sack, &pk_console_update_system_filter_helper, NULL);
```
hughsie: space before bracket pls

darktemplarbasealt:
Uploaded updated commits with implemented mentioned changes

PR #343, opened 2019-10-09
zypp: Fix get-updates in Tumbleweed

JonathanKang:
Previously, whether you're using Tumbleweed or Leap, get-updates
always uses doUpdate() to fetch available updates. It's the same output
as "zypper lu". While in Tumbleweed, you won't get the desired updates
list like "zypper dup".

Fix that by adding a is_tumbleweed() function to detect whether you're
using Tumbleweed and call correct resolver function accordingly.


JonathanKang:
moved to #345

Issue #347, opened 2019-10-11
pkcon does not play nice in small windows; looks bad on terminal widths below 66 columns


deftdawg:
pkcon is unfortunately somewhat awful to use within the confines of a termux window on an Android device. 

List of packages to be updated is a big jumble, in part because full descriptions of every package are displayed.

When waiting for user to enter sudo credentials to authorize package upgrades, the prompt is quickly lost by all the drawing of ascii knight rider scanner characters.

When downloading packages status scrolls everywhere... 

Is there a way to simplify the output to produce something more readable like apt used to do?

**EDIT** It seems any terminal with a width below 66 width, bars redraw a mess, the narrower the worse it gets.

![image](https://user-images.githubusercontent.com/4991612/66692978-d75a7500-ec71-11e9-85cd-c0fc275f78a7.png)

![2019-10-11](https://user-images.githubusercontent.com/4991612/66686412-63a67100-ec4d-11e9-8b40-6a6a205a369a.jpg)

![2019-10-11 (1)](https://user-images.githubusercontent.com/4991612/66686407-6012ea00-ec4d-11e9-956d-0c340b112693.jpg)

hughsie:
I think we need to look at the $COLUMNS value in the code. If you did that I'd merge a PR. :)

deftdawg:
I suspect it's way more complicated than that to do it properly... 

The most basic fix (which I may take a hack at) is to replace the static values assigned to the progress bar creation in these two lines:
https://github.com/hughsie/PackageKit/blob/master/client/pk-console.c#L1656-L1657
with something calculated based on the available terminal columns.

The proper way to fix this is probably to rewrite how stuff is rendered using ncurses or something that can handle terminal resizes, as just doing what I will attempted will probably still look awful should the terminal get resized.

hughsie:
> to the progress bar creation in these two lines

I think that would fix the bug here, indeed.

> The proper way to fix this is probably to rewrite how stuff is rendered using ncurses

ENOTIME for that I'm afraid.

ximion:
I added some code that should address the progress bar side of things, but this may need a bit more testing than I have given it so far (also, the text display may still not be ideal for terminals with small widths).

PR #348, opened 2019-10-18
* Shrink the progress bar to fit when run in small spaces

deftdawg:
For #347

Work pretty well in most cases, except when termux prompts for authentication which still get the staircase in a high rise from progress bar scanning.

### DIFF
--- a/client/pk-console.c
+++ b/client/pk-console.c
@@ -32,6 +32,8 @@
 #include <sys/types.h>
 #include <pwd.h>
 #include <locale.h>
+#include <sys/ioctl.h>
+#include <sys/param.h>
 
 #define PK_EXIT_CODE_SYNTAX_INVALID	3
 #define PK_EXIT_CODE_FILE_NOT_FOUND	4
@@ -1601,6 +1603,9 @@ main (int argc, char *argv[])
 	g_autofree gchar *filter = NULL;
 	g_autofree gchar *options_help = NULL;
 	g_autofree gchar *summary = NULL;
+	guint bar_padding = 30;
+	guint bar_size = 25;
+	struct winsize w;
 
 	const GOptionEntry options[] = {
 		{ "version", '\0', 0, G_OPTION_ARG_NONE, &program_version,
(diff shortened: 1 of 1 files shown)


client/pk-console.c
```diff
@@ -1652,9 +1654,17 @@ main (int argc, char *argv[])
 				ctx,
 				NULL);
 
+	// Shrink the progresbar to fit in small spaces i.e. termux, small tmux panes, large font terminals
+	struct winsize w;
+	ioctl(STDOUT_FILENO, TIOCGWINSZ, &w);
```
hughsie: space before bracket, and I think we need to initialize the `w` object with NULL. I think we should only query `ws_col` if the ioctl succeeded too.

deftdawg: Fixed the formatting and ioctl invocation logic.   As far as init-ing with NULL, I have no idea what the correct way to do it is (none of the examples I found did it), alas I'm much not a C guy...

client/pk-console.c
```diff
@@ -1652,9 +1654,21 @@ main (int argc, char *argv[])
 				ctx,
 				NULL);
 
+	// default progress bar padding and bar size to old values of 30 and 25 respectively
+	int bar_padding = 30;
+	int bar_size = 25;
```
hughsie: These need to be `guint` rather than `int` and also declared at the top of the function in C99 style.

client/pk-console.c
```diff
@@ -1652,9 +1654,21 @@ main (int argc, char *argv[])
 				ctx,
 				NULL);
 
+	// default progress bar padding and bar size to old values of 30 and 25 respectively
```
hughsie: no comment reqd.

client/pk-console.c
```diff
@@ -1652,9 +1654,21 @@ main (int argc, char *argv[])
...
+	int bar_padding = 30;
+	int bar_size = 25;
+
+	// Shrink the progresbar to fit in small spaces i.e. termux, small tmux panes, large font terminals
+	struct winsize w;
```
hughsie: same, at the top please.

client/pk-console.c
```diff
@@ -1652,9 +1654,21 @@ main (int argc, char *argv[])
...
+	// If ioctl reports back and the terminal is small, shrink to fit as best we can
```
hughsie: C style comments please, i.e. `/* foo */`

hughsie:
Getting there, thanks! A few more style issues -- then please rebase against master and squash into one commit and force push to your branch to update the PR.

PR #350, opened 2019-10-23
dnf: Reset libgit2 module when upgrading to F31

kalev:
Work around a modularity upgrade bug by resetting the libgit2 module
when upgrading to F31. This depends on new libdnf API from
https://github.com/rpm-software-management/libdnf/pull/822

https://bugzilla.redhat.com/show_bug.cgi?id=1762751

### DIFF
--- a/backends/dnf/pk-backend-dnf.c
+++ b/backends/dnf/pk-backend-dnf.c
@@ -3435,6 +3435,17 @@ pk_backend_upgrade_system_thread (PkBackendJob *job, GVariant *params, gpointer 
 	/* set the installonly limit one higher than usual to avoid removing any kernels during system upgrades */
 	dnf_sack_set_installonly_limit (sack, dnf_context_get_installonly_limit (job_data->context) + 1);
 
+	/* reset libgit2 module when upgrading to F31: https://bugzilla.redhat.com/show_bug.cgi?id=1762751 */
+	if (g_strcmp0 (release_ver, "31") == 0) {
+		const gchar *reset_modules[] = { "libgit2", NULL };
+		g_autoptr(GError) error_local = NULL;
+
+		g_debug ("resetting libgit2 module");
+		if (!dnf_context_reset_modules (job_data->context, sack, reset_modules, &error_local)) {
+			g_warning ("failed to reset libgit2 module: %s", error_local->message);
+		}
+	}
+
 	/* done */
 	if (!dnf_state_done (job_data->state, &error)) {
 		pk_backend_job_error_code (job, error->code, "%s", error->message);


hughsie:
Makes sense, although it's a shame to hardcode it and not read it from some config file. We ought to get a new enough libdnf in CI too.

backends/dnf/pk-backend-dnf.c
```diff
@@ -3435,6 +3435,17 @@ pk_backend_upgrade_system_thread (PkBackendJob *job, GVariant *params, gpointer
 	/* set the installonly limit one higher than usual to avoid removing any kernels during system upgrades */
 	dnf_sack_set_installonly_limit (sack, dnf_context_get_installonly_limit (job_data->context) + 1);
 
+	/* reset libgit2 module when upgrading to F31: https://bugzilla.redhat.com/show_bug.cgi?id=1762751 */
+	if (g_strcmp0 (release_ver, "31") == 0) {
+		const gchar *reset_modules[] = { "libgit2", NULL };
+		g_debug ("resetting libgit2 module");
+		if (!dnf_context_reset_modules (job_data->context, sack, reset_modules, &error)) {
```
hroncok: can this happen if the libgit2 module isn't available? should ti be ignored instead?

kalev: Ah yes, that's probably a good idea

kalev: OK, updated the PR to not error out in case the reset fails

backends/dnf/pk-backend-dnf.c
```diff
@@ -3435,6 +3435,17 @@ pk_backend_upgrade_system_thread (PkBackendJob *job, GVariant *params, gpointer
 	/* set the installonly limit one higher than usual to avoid removing any kernels during system upgrades */
 	dnf_sack_set_installonly_limit (sack, dnf_context_get_installonly_limit (job_data->context) + 1);
 
+	/* reset libgit2 module when upgrading to F31: https://bugzilla.redhat.com/show_bug.cgi?id=1762751 */
+	if (g_strcmp0 (release_ver, "31") == 0) {
```
Conan-Kudo: Uhh, you guys? This patch will kick in for *any* distribution with the version 31. Why isn't this guarded behind a check on it being Fedora too?

kalev: Sorry, not interested in supporting any other distros than Fedora and RHEL in this backend. Feel free to submit patches if you need.

Conan-Kudo: You shouldn't have merged this then. You *know* multiple distros use this backend, and we've historically not included distro-specific patches in *any* backend...

Conan-Kudo: Well, I suppose that's not completely true. We do have the distro-specific C files...

PR #352, opened 2019-11-09
PkControl: Connect to DBus on construct

### LINKED ISSUE #207: packagekit-glib2 signals not fired after using synchronous method
Hi,

I hope this is a suitable place to report this as it's quite possible something I don't understand rather than an issue, but it's been causing a fair amount of confusion. I've written the following Vala sample that demonstrates the issue:

```
class SignalTest : Application {
    Pk.Control control;
    Pk.Task client;
    bool broken;

    public SignalTest (bool broken) {
        this.broken = broken;
    }

    public async void update_cache () {
        try {
            yield client.refresh_cache_async (false, null, (t, p) => { });
        } catch (Error e) {
            stderr.printf ("%s\n",e.message);
        }
    }

    public override void activate () {
        client = new Pk.Task ();

        if (broken) {
            Pk.Bitfield filter = Pk.Bitfield.from_enums (Pk.Filter.INSTALLED, Pk.Filter.NEWEST);
            try {
                stdout.printf ("running sync method\n");
                client.get_packages_sync (filter, null, (prog, type) => {});
                stdout.printf ("sync method finished\n");
            } catch (Error e) {
                stderr.printf ("%s\n", e.message);
            }
        }

        control = new Pk.Control ();
        control.updates_changed.connect (() => {
            stdout.printf ("!!! updates changed !!!\n");
            release ();
        });
        stdout.printf ("checking for updates\n");
        update_cache.begin ();

        hold ();
    }
}

int main (string[] args) {
    stdout.printf ("First run, get signal\n");
    new SignalTest (false).run (args);
    stdout.printf ("Second run, don't get signal\n");
    new SignalTest (true).run (args);

    return 0;
}
```

What this tries to demonstrate is that the `Pk.Control.updates_changed` signal is never received after running seemingly any synchronous method on the `Pk.Client`. The signal is still fired and can be seen on D-BUS and in `pkmon`.

Interestingly, it seems that wrapping the async version of the method in a MainLoop works OK, but as this is seemingly what the sync methods do internally, I'm struggling to understand why using the sync methods causes the above problem.

This is reproducible 100% of the time on Ubuntu 16.04 with Vala 0.34. Can anyone shed any light on this issue?

Thank you for any help in advance,
David

davidmhewitt:
Fixes #207 

Currently `PkControl` only constructs its `GDBusProxy` object when a method is called. I.e. you have to construct a `PkControl` and then call something like `pk_control_get_properties_async` for a connection to the bus to be made. This causes two issues:

- You can't create a `PkControl` object, then only connect up to a signal (`updates-changed` for example) without calling any methods and expect to receive the signal as a DBus connection will never be made.
- The second, more problematic issue is that all of the synchronous methods within `packagekit-glib2` wrap the async versions of their method in a `GMainLoop` with a temporary thread-default main context. Usage of these methods can cause signals to never be received from the `GDBusProxy` even after the `PkControl` object has connected to the bus, explanation follows:

The `GDBusProxy` documentation states:
> A GDBusProxy instance can be used from multiple threads but note that all signals (e.g. “g-signal”, “g-properties-changed” and “notify”) are emitted in the thread-default main context of the thread where the instance was constructed.

Given that the `GDBusProxy` object is currently only constructed in the the getter/setter style methods in `PkControl`, if the first thing we call in the `packagekit-glib2` API is a synchronous public method on `PkClient` for example, it will create a new temporary thread-default context, and within that context call `pk_control_get_tid_async`, which in turn will construct the `GDBusProxy` object in the wrong thread context. This causes the `PkControl` to never receive any signals from the `GDBusProxy`.

I propose this PR which tries to connect to the bus synchronously in the constructor. It also does the same in the `name_appeared` callback, however, this alone wasn't enough. It was still possible to have a race condition where a synchronous method elsewhere in the library was able to cause the construction of the `GDBusProxy` object before the `name_appeared` callback and still break the signals.

### DIFF
--- a/lib/packagekit-glib2/pk-control.c
+++ b/lib/packagekit-glib2/pk-control.c
@@ -2343,6 +2343,60 @@ pk_control_class_init (PkControlClass *klass)
 }
 
 /*
+ * pk_control_construct_proxy_finish:
+ **/
+static void
+pk_control_construct_proxy_finish (PkControlState *state)
+{
+	g_object_unref (state->control);
+	if (state->proxy != NULL)
+		g_object_unref (state->proxy);
+	g_slice_free (PkControlState, state);
+}
+
+/*
+ * pk_control_construct_proxy:
+ **/
+static void
+pk_control_construct_proxy (PkControl *control)
+{
+	PkControlState *state;
+
+	/* already got a proxy, don't need to continue */
+	if (control->priv->proxy != NULL) {
+		return;
+	}
+
+	state = g_slice_new0 (PkControlState);
+	state->control = g_object_ref (control);
+
+	/* get a connection to the main interface */
+	state->proxy = g_dbus_proxy_new_for_bus_sync (G_BUS_TYPE_SYSTEM,
+						      G_DBUS_PROXY_FLAGS_NONE,
+						      NULL,
+						      PK_DBUS_SERVICE,
+						      PK_DBUS_PATH,
+						      PK_DBUS_INTERFACE,
+						      control->priv->cancellable,
+						      NULL);
+
+	if (state->proxy == NULL) {
+		pk_control_construct_proxy_finish (state);
+		return;
+	}
+
+	/* connect up proxy */
+	pk_control_proxy_connect (state);
+
+	control->priv->connected = TRUE;
+	g_debug ("notify::connected");
+	g_object_notify (G_OBJECT(control), "connected");
+
+	/* we're done */
+	pk_control_construct_proxy_finish (state);
+}
+
+/*
  * pk_control_name_appeared_cb:
  **/
 static void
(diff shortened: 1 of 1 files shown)


hughsie:
> You can't create a PkControl object, then only connect up to a signal (updates-changed for example) without calling any methods

Can't you just call pk_control_get_properties() here? I really don't like connecting to the proxy without a proper GCancellable and GError.

davidmhewitt:
> Can't you just call pk_control_get_properties() here? I really don't like connecting to the proxy without a proper GCancellable

Yes, and that's exactly the workaround I've proposed in AppCenter as the stopgap for fixing the library. As it's somewhat of a bug for all the signals on an object to never fire just because you happened to call a synchronous method in the library's public API, wouldn't you say?

hughsie:
> to call a synchronous method

There's an async version too, right?

davidmhewitt:
> There's an async version too, right?

There is, but given the amount of time I've wasted on a bug caused by daring to use a method in the public API of a library is such that I thought it would be worth trying to help out with this.

If this isn't an acceptable solution, can you suggest improvements or help me to understand the usage of temporary `GMainContext`s used in all of the synchronous methods? Would these methods not work without pushing a new context and just using the already existing default context?

hughsie:
We can't do anything remotely synchronous in the constructor. Most other apis either make the constructor failable (with a GCancellsble) or require the user to 'load' or 'setup' the instance post instantiation.

PR #356, opened 2019-11-20
Zypp: Make name/description search case insensitive

Appadeia:
This addresses https://github.com/hughsie/PackageKit/issues/354 and matches the behavior of zypper command line.

Example searches to test: `mozillafirefox` (MozillaFirefox), `playonlinux` (PlayOnLinux).


ximion:
That looks very straighforward to me!
I'd merge it if there is no Zypper backend maintainer who can do so (or has objections).

PR #358, opened 2019-11-30
command-not-found: Remove spurious ellipsis...

andersk:
`bash` does not output an ellipsis here...so there’s no reason for `command-not-found` to do so...it was actually kind of annoying...


hughsie:
The ellipsis was deliberate, in that something was pending and processing, rather than just hung.

PR #361, opened 2020-01-07
dnf: Condition the libgit2 module reset hack to the Fedora vendor

Conan-Kudo:
As the libgit2 module reset hack introduced in 6dc4d503cfab4c0270924ddf8acfefb4f06d47fa
is specific to Fedora and the DNF backend is used by other distributions,
it makes sense to condition the code behind the fedora vendor flag.


Conan-Kudo:
This is "compromise" version to avoid quirks with other distributions using the DNF backend. My preferred way to fix this is in #364.

hughsie:
We merged #364 instead. Thanks.

Conan-Kudo:
Thanks @hughsie.

PR #363, opened 2020-01-07
dnf: Add vendor configuration for ROSA Linux

Conan-Kudo:
Starting with the ROSA Linux 2019.1 platform, ROSA Linux will use
DNF as the PackageKit backend.

### DIFF
--- a/backends/dnf/Makefile.am
+++ b/backends/dnf/Makefile.am
@@ -3,7 +3,8 @@ plugin_LTLIBRARIES = libpk_backend_dnf.la
 EXTRA_DIST =								\
 	dnf-backend-vendor-fedora.c					\
 	dnf-backend-vendor-mageia.c					\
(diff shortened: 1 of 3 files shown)


Conan-Kudo:
cc: @mikhailnov

mikhailnov:
Rosa does not have ports to neither armv8 nor armv7

mikhailnov:
Repo main seems to be missing. Also there is main/testing which is ised to publish packages being at QA testing

Conan-Kudo:
@mikhailnov main is the base case. These are repo IDs from rosa-repos. If there are no plans for ARM, I'll drop them.

Conan-Kudo:
@mikhailnov I've dropped the ARM architectures for now, so this should match what is available in ROSA 2019.1-dev.

PR #364, opened 2020-01-07
Revert "dnf: Reset libgit2 module when upgrading to F31"

Conan-Kudo:
This change has the potential to break other distributions
using the DNF backend, and exclusively applies to Fedora.

This is better maintained as a downstream-only patch in Fedora.

This reverts commit 6dc4d503cfab4c0270924ddf8acfefb4f06d47fa.

This is an alternative to #361. I'd personally prefer this to be merged.


hughsie:
I agree this should be a downstream hack.

PR #369, opened 2020-01-19
dnf: Load all the repos and vars directories

Conan-Kudo:
Historically, the backend has internally determined its setup with
static values. However, we generally want PackageKit to load all
repositories defined in all repository directories that DNF normally
searches, since it is not guaranteed to be in `/etc/yum.repos.d` and
DNF supports multiple repository configuration paths.

We also need the vars to be loaded so that repository definitions
that rely on more than the built-in vars will work.

This bumps our dependency for libdnf to 0.43.1, as we're using APIs
introduced in this release.

### DIFF
--- a/backends/dnf/pk-backend-dnf.c
+++ b/backends/dnf/pk-backend-dnf.c
@@ -135,11 +135,12 @@ pk_backend_context_invalidate_cb (DnfContext *context,
 static gboolean
 pk_backend_setup_dnf_context (DnfContext *context, GKeyFile *conf, const gchar *release_ver, GError **error)
 {
+	const gchar * const *repo_dirs;
+	const gchar * const *var_dirs;
 	gboolean keep_cache;
 	g_autofree gchar *cache_dir = NULL;
 	g_autofree gchar *destdir = NULL;
 	g_autofree gchar *lock_dir = NULL;
-	g_autofree gchar *repo_dir = NULL;
 	g_autofree gchar *solv_dir = NULL;
 
 	destdir = g_key_file_get_string (conf, "Daemon", "DestDir", NULL);
@@ -150,11 +151,31 @@ pk_backend_setup_dnf_context (DnfContext *context, GKeyFile *conf, const gchar *
 	dnf_context_set_cache_dir (context, cache_dir);
 	solv_dir = g_build_filename (destdir, "/var/cache/PackageKit", release_ver, "hawkey", NULL);
 	dnf_context_set_solv_dir (context, solv_dir);
-	repo_dir = g_build_filename (destdir, "/etc/yum.repos.d", NULL);
-	dnf_context_set_repo_dir (context, repo_dir);
 	lock_dir = g_build_filename (destdir, "/var/run", NULL);
 	dnf_context_set_lock_dir (context, lock_dir);
 	dnf_context_set_rpm_verbosity (context, "info");
+
+	/* Add prefix to repo directories */
+	repo_dirs = dnf_context_get_repos_dir (context);
+	if (repo_dirs != NULL && repo_dirs[0] != NULL) {
+		g_auto(GStrv) full_repo_dirs = NULL;
+		guint len = g_strv_length ((gchar **)repo_dirs);
+		full_repo_dirs = g_new0 (gchar*, len + 1);
+		for (guint i = 0; i < len; i++)
+			full_repo_dirs[i] = g_build_filename (destdir, repo_dirs[i], NULL);
+		dnf_context_set_repos_dir (context, (const gchar * const*)full_repo_dirs);
+	}
+
+	/* Add prefix to var directories */
+	var_dirs = dnf_context_get_vars_dir (context);
+	if (var_dirs != NULL && var_dirs[0] != NULL) {
+		g_auto(GStrv) full_var_dirs = NULL;
+		guint len = g_strv_length ((gchar **)var_dirs);
+		full_var_dirs = g_new0 (gchar*, len + 1);
+		for (guint i = 0; i < len; i++)
+			full_var_dirs[i] = g_build_filename (destdir, var_dirs[i], NULL);
+		dnf_context_set_vars_dir (context, (const gchar * const*)full_var_dirs);
+	}
 
 	/* use this initial data if repos are not present */
 	dnf_context_set_vendor_cache_dir (context, "/usr/share/PackageKit/metadata");
--- a/backends/dnf/meson.build
+++ b/backends/dnf/meson.build
@@ -1,5 +1,5 @@
 appstream_dep = dependency('appstream-glib')
-dnf_dep = dependency('libdnf', version: '>=0.22.0')
+dnf_dep = dependency('libdnf', version: '>=0.43.1')
 rpm_dep = dependency('rpm')
 
 shared_module(
--- a/tests/ci/Dockerfile-fedora
+++ b/tests/ci/Dockerfile-fedora
@@ -1,4 +1,4 @@
-FROM fedora:29
+FROM fedora:31
 
 RUN dnf -y update
 RUN dnf -y install dnf-plugins-core libdnf-devel redhat-rpm-config meson gcc ninja-build


Conan-Kudo:
I'm not entirely sure I did this quite right... (cc: @j-mracek, @jrohel)

Conan-Kudo:
@hughsie I don't know if we want to keep the commit to enable updates-testing, but at least it makes CI pass for now...

backends/dnf/pk-backend-dnf.c
```diff
@@ -150,10 +152,16 @@ pk_backend_setup_dnf_context (DnfContext *context, GKeyFile *conf, const gchar *
 	dnf_context_set_cache_dir (context, cache_dir);
 	solv_dir = g_build_filename (destdir, "/var/cache/PackageKit", release_ver, "hawkey", NULL);
 	dnf_context_set_solv_dir (context, solv_dir);
-	repo_dir = g_build_filename (destdir, "/etc/yum.repos.d", NULL);
-	dnf_context_set_repo_dir (context, repo_dir);
 	lock_dir = g_build_filename (destdir, "/var/run", NULL);
 	dnf_context_set_lock_dir (context, lock_dir);
+	repo_dirs = dnf_context_get_repos_dir (context);
+	for (diriter = 0; repo_dirs[diriter] != NULL; diriter++)
```
hughsie: `for (guint i = 0...` please

hughsie:
Looks sane enough to me, but comments welcome from @kalev too.

backends/dnf/pk-backend-dnf.c
Conan-Kudo: I prefer that myself, but I saw other code in here doing otherwise, so...

kalev:
Have you actually tried running the code? To me it looks like the memory management is all wonky. I'd expect crashes.

e.g. both `dnf_context_get_repos_dir` and `dnf_context_get_vars_dir` are returning const const pointers to internal data, but in the caller they are both marked as g_autofree, which likely leads to double frees.

Then, you are replacing DnfContext owned array members with newly allocated strings, without freeing old strings -- likely causing a leak there?

Also, can you add a comment to the code why all of this is necessary? It looks like it's just adding a destdir prefix to all of the paths; why can't DnfContext do that automatically?

jrohel:
@Conan-Kudo 
The work with memory is bugy. Code overwrites the internal array. Double frees, memory leaks.
Example code. The code is not nice but does what is needed (temporary array is needed):
```
    const gchar * const *repo_dirs;
    const gchar * const *var_dirs;

    repo_dirs = dnf_context_get_repos_dir (context);
    if (repo_dirs[0] != NULL) {
        g_auto(GStrv) tmp_repo_dirs = NULL;
        guint len = 1;
        while (repo_dirs[len] != NULL)
            ++len;
        tmp_repo_dirs = g_new0(gchar*, len + 1);
        for (guint diriter = 0; diriter < len; diriter++)
            tmp_repo_dirs[diriter] = g_build_filename (destdir, repo_dirs[diriter]);
        dnf_context_set_repos_dir (context, tmp_repo_dirs);
    }
    var_dirs = dnf_context_get_vars_dir (context);
    if (var_dirs[0] != NULL) {
        g_auto(GStrv) tmp_var_dirs = NULL;
        guint len = 1;
        while (var_dirs[len] != NULL)
            ++len;
        tmp_var_dirs = g_new0(gchar*, len + 1);
        for (guint diriter = 0; var_dirs[diriter] != NULL; diriter++)
            tmp_var_dirs[diriter] = g_build_filename (destdir, var_dirs[diriter]);
        dnf_context_set_vars_dir (context, tmp_var_dirs);
    }
```
Anyway, I have not know that PackageKit supports  installroot. I never tried it.

jrohel:
@kalev
> Also, can you add a comment to the code why all of this is necessary? It looks like it's just adding a destdir prefix to all of the paths; why can't DnfContext do that automatically?

Automatically adding a prefix assumes that the paths in configuration file are relative to the destination directory (installation root). It assumes that repository configuration and variables configuration are present in the destination directory.

If new empty installation root is processed there is nothing inside. So, it is about use case. Sometime we want to use configuration from the host and sometime from the destination directory. That is why libdnf does not adding prefixes. But application can do it by libdnf API.

I see, that PackageKit was adding prefix before this PR:
```
repo_dir = g_build_filename (destdir, "/etc/yum.repos.d", NULL);
dnf_context_set_repo_dir (context, repo_dir);
```
Thus, the PR does not change  behavior of PackageKit. But adds support for more repository directories and for variables.

Conan-Kudo:
@jrohel Thanks for the suggestion on how to fix the code. 👍

Conan-Kudo:
@kalev This should fix some quirks I see with PackageKit on CentOS 8, since they use custom vars in their repository definitions.

kalev:
Can you explain please what is your use case for needing to add destdir (e.g. not using /etc/yum.repos.d, but instead adding a custom prefix e.g. /tmp/random.dir/etc/yum.repos.d)?

kalev:
> @kalev This should fix some quirks I see with PackageKit on CentOS 8, since they use custom vars in their repository definitions.

But you are not doing anything with custom vars here, all this code does is add destdir everywhere?

Conan-Kudo:
@kalev I'm preserving the capability that already existed prior to this change. I presume this functionality is used by tools that use PackageKit in this manner (I know of a few desktop and mobile things that do this). I don't particularly feel like breaking that capability.

backends/dnf/pk-backend-dnf.c
Conan-Kudo: Fixed.

Conan-Kudo:
> Can you explain please what is your use case for needing to add destdir (e.g. not using /etc/yum.repos.d, but instead adding a custom prefix e.g. /tmp/random.dir/etc/yum.repos.d)?

There are tools out there that do use PackageKit this way on desktop and mobile platforms. It was already supported before, and I just kept supporting it.

Conan-Kudo:
> > @kalev This should fix some quirks I see with PackageKit on CentOS 8, since they use custom vars in their repository definitions.
> 
> But you are not doing anything with custom vars here, all this code does is add destdir everywhere?

I am loading the vars directories now, which we did not do before. The dnf context API does not have a lot of implicit behaviors, and if we don't set up the context to do this, it won't happen at all.

Conan-Kudo:
@kalev Also to note, DNF itself supports loading from multiple directories, whereas the DNF context API as PackageKit previously used it only would load from _one_ directory. This makes it consistent with DNF in how repos and variables are loaded.

kalev:
> > > @kalev This should fix some quirks I see with PackageKit on CentOS 8, since they use custom vars in their repository definitions.
> > 
> > 
> > But you are not doing anything with custom vars here, all this code does is add destdir everywhere?
> 
> I am loading the vars directories now, which we did not do before. The dnf context API does not have a lot of implicit behaviors, and if we don't set up the context to do this, it won't happen at all.

You are just reading back vars directories that are already set on the DnfContext and adding prefix everywhere. This is all the code does, from what I can tell.

I don't think you are setting up any extra vars directories, just editing the existing ones.

Conan-Kudo:
@kalev Perhaps you're right and I don't need this if we decided to break destdir support, but I haven't found any particular evidence that all the repos and all the vars are being loaded automatically. At *least*, with the current git master, PackageKit is forcing that only `/etc/yum.repos.d` will work, and that's inconsistent with what DNF itself does.

kalev:
Ahh, I see now -- sorry, I totally missed earlier that you are removing a `dnf_context_set_repo_dir ("/etc/yum.repos.d") call`. That's really the core change here, right?

backends/dnf/pk-backend-dnf.c
```diff
@@ -150,12 +151,34 @@ pk_backend_setup_dnf_context (DnfContext *context, GKeyFile *conf, const gchar *
 	dnf_context_set_cache_dir (context, cache_dir);
 	solv_dir = g_build_filename (destdir, "/var/cache/PackageKit", release_ver, "hawkey", NULL);
 	dnf_context_set_solv_dir (context, solv_dir);
-	repo_dir = g_build_filename (destdir, "/etc/yum.repos.d", NULL);
-	dnf_context_set_repo_dir (context, repo_dir);
 	lock_dir = g_build_filename (destdir, "/var/run", NULL);
 	dnf_context_set_lock_dir (context, lock_dir);
 	dnf_context_set_rpm_verbosity (context, "info");
 
+	/* Set repos and vars dirs */
```
kalev: I think a better comment here would be "add prefix to repo directories"

backends/dnf/pk-backend-dnf.c
```diff
@@ -150,12 +151,34 @@ pk_backend_setup_dnf_context (DnfContext *context, GKeyFile *conf, const gchar *
...
+	repo_dirs = dnf_context_get_repos_dir (context);
+	if (repo_dirs[0] != NULL) {
+		g_auto(GStrv) full_repo_dirs = NULL;
+		guint len = 1;
+		while (repo_dirs[len] != NULL)
+			len++;
```
kalev: Maybe just call `g_strv_length`?

backends/dnf/pk-backend-dnf.c
```diff
@@ -150,12 +151,34 @@ pk_backend_setup_dnf_context (DnfContext *context, GKeyFile *conf, const gchar *
...
+		full_repo_dirs = g_new0(gchar*, len + 1);
+		for (guint diriter = 0; diriter < len; diriter++)
```
kalev: We usually use `i` as a for loop counter

backends/dnf/pk-backend-dnf.c
```diff
@@ -150,12 +151,34 @@ pk_backend_setup_dnf_context (DnfContext *context, GKeyFile *conf, const gchar *
...
 	dnf_context_set_rpm_verbosity (context, "info");
 
+	/* Set repos and vars dirs */
+	repo_dirs = dnf_context_get_repos_dir (context);
+	if (repo_dirs[0] != NULL) {
+		g_auto(GStrv) full_repo_dirs = NULL;
+		guint len = 1;
+		while (repo_dirs[len] != NULL)
+			len++;
+		full_repo_dirs = g_new0(gchar*, len + 1);
```
kalev: Space before opening bracket please

backends/dnf/pk-backend-dnf.c
```diff
@@ -150,12 +151,34 @@ pk_backend_setup_dnf_context (DnfContext *context, GKeyFile *conf, const gchar *
...
+		for (guint diriter = 0; diriter < len; diriter++)
+			full_repo_dirs[diriter] = g_build_filename (destdir, repo_dirs[diriter]);
+		dnf_context_set_repos_dir (context, full_repo_dirs);
+	}
+	var_dirs = dnf_context_get_vars_dir (context);
```
kalev: I would add a newline and a `/* add prefix to vars directories */` comment here

backends/dnf/pk-backend-dnf.c
```diff
@@ -150,12 +151,34 @@ pk_backend_setup_dnf_context (DnfContext *context, GKeyFile *conf, const gchar *
...
 	dnf_context_set_solv_dir (context, solv_dir);
-	repo_dir = g_build_filename (destdir, "/etc/yum.repos.d", NULL);
-	dnf_context_set_repo_dir (context, repo_dir);
 	lock_dir = g_build_filename (destdir, "/var/run", NULL);
 	dnf_context_set_lock_dir (context, lock_dir);
 	dnf_context_set_rpm_verbosity (context, "info");
 
+	/* Set repos and vars dirs */
+	repo_dirs = dnf_context_get_repos_dir (context);
+	if (repo_dirs[0] != NULL) {
```
kalev: Can repo_dirs be NULL?

backends/dnf/pk-backend-dnf.c
```diff
@@ -150,12 +151,34 @@ pk_backend_setup_dnf_context (DnfContext *context, GKeyFile *conf, const gchar *
...
+		for (guint diriter = 0; diriter < len; diriter++)
+			full_repo_dirs[diriter] = g_build_filename (destdir, repo_dirs[diriter]);
+		dnf_context_set_repos_dir (context, full_repo_dirs);
+	}
+	var_dirs = dnf_context_get_vars_dir (context);
+	if (var_dirs[0] != NULL) {
+		g_auto(GStrv) full_var_dirs = NULL;
+		guint len = 1;
+		while (var_dirs[len] != NULL)
+			len++;
```
kalev: `g_strv_length` should be easier here as well

Conan-Kudo:
@kalev Yep!

backends/dnf/pk-backend-dnf.c
```diff
@@ -150,12 +151,34 @@ pk_backend_setup_dnf_context (DnfContext *context, GKeyFile *conf, const gchar *
...
+		full_var_dirs = g_new0(gchar*, len + 1);
+		for (guint diriter = 0; diriter < len; diriter++)
```
kalev: `i` instead of `diriter` please

backends/dnf/pk-backend-dnf.c
```diff
@@ -150,12 +151,34 @@ pk_backend_setup_dnf_context (DnfContext *context, GKeyFile *conf, const gchar *
...
+			full_repo_dirs[diriter] = g_build_filename (destdir, repo_dirs[diriter]);
+		dnf_context_set_repos_dir (context, full_repo_dirs);
+	}
+	var_dirs = dnf_context_get_vars_dir (context);
+	if (var_dirs[0] != NULL) {
+		g_auto(GStrv) full_var_dirs = NULL;
+		guint len = 1;
+		while (var_dirs[len] != NULL)
+			len++;
+		full_var_dirs = g_new0(gchar*, len + 1);
```
kalev: Space before opening bracket please

backends/dnf/pk-backend-dnf.c
Conan-Kudo: In theory, it's possible. I don't know if I'd see it in practice, but I didn't think it'd hurt to have that conditional.

Conan-Kudo:
@kalev I incorporated your suggestions, thanks for the feedback! 💯

backends/dnf/pk-backend-dnf.c
Conan-Kudo: Done.

backends/dnf/pk-backend-dnf.c
kalev: No, I mean, you are checking that the first element is non-NULL, but can the whole repo_dirs be NULL? You are not checking for that.

Conan-Kudo: Hmm, that's a good question. I think it _might_ be possible, though rare. I should fix that for both `repo_dirs` and `var_dirs` checks...

Conan-Kudo: This should be fixed now.

backends/dnf/pk-backend-dnf.c
Conan-Kudo: Fixed.

backends/dnf/pk-backend-dnf.c
Conan-Kudo: Fixed.

backends/dnf/pk-backend-dnf.c
Conan-Kudo: Fixed.

backends/dnf/pk-backend-dnf.c
Conan-Kudo: Done.

backends/dnf/pk-backend-dnf.c
Conan-Kudo: Fixed.

backends/dnf/pk-backend-dnf.c
Conan-Kudo: Done. Thanks!

backends/dnf/pk-backend-dnf.c
Conan-Kudo: Done, thanks!

kalev:
CI shows a bunch of compiler warnings from the new code. Can you fix those please?

Conan-Kudo:
@hughsie @kalev I'm not really sure how to resolve the issue with the warnings, I'd appreciate some assistance here?

backends/dnf/pk-backend-dnf.c
```diff
@@ -150,12 +151,32 @@ pk_backend_setup_dnf_context (DnfContext *context, GKeyFile *conf, const gchar *
 	dnf_context_set_cache_dir (context, cache_dir);
 	solv_dir = g_build_filename (destdir, "/var/cache/PackageKit", release_ver, "hawkey", NULL);
 	dnf_context_set_solv_dir (context, solv_dir);
-	repo_dir = g_build_filename (destdir, "/etc/yum.repos.d", NULL);
-	dnf_context_set_repo_dir (context, repo_dir);
```
hughsie, line 154: can we do this as one commit, and the destdir stuff in another?

Conan-Kudo: it's an atomic change as we're switching from one directory to multiple directories. destdir is already there...

hughsie: Ahh okay. If it's warning free I guess it LGTM.

backends/dnf/pk-backend-dnf.c
```diff
@@ -150,12 +151,32 @@ pk_backend_setup_dnf_context (DnfContext *context, GKeyFile *conf, const gchar *
...
+	}
+
+	/* Add prefix to var directories */
+	var_dirs = dnf_context_get_vars_dir (context);
+	if (var_dirs != NULL && var_dirs[0] != NULL) {
+		g_auto(GStrv) full_var_dirs = NULL;
+		guint len = g_strv_length ((gchar **)var_dirs);
+		full_var_dirs = g_new0 (gchar*, len + 1);
+		for (guint i = 0; i < len; i++)
+			full_var_dirs[i] = g_build_filename (destdir, var_dirs[i], NULL);
```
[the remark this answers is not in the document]
Conan-Kudo: It's a somewhat blind copy of the repos logic. I tested it in the context to fix PackageKit on CentOS, which had been broken for years because it didn't load vars at all.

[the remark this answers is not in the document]
Conan-Kudo: To the best of my knowledge, this works as intended, I didn't know that it would do auto-prefixing that way...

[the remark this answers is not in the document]
lucab: Yes, well, I don't know.
Perhaps this get used most of the times with a `destdir` set to `/`? That would result in a final lookup to `///etc/dnf/vars`, which would go unnoticed.
I admit I'm just randomly guessing at this point, but if you have a process at hand to strace it may shed some light.

PR #380, opened 2020-03-07
ci: Parallelize Fedora and Debian PackageKit CI jobs

Conan-Kudo:
Rather than having them run one after the other, it makes sense
to have them run independently in parallel. This also makes it
clearer in the job logs, since now each build is an independent
CI job.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,15 @@
-language: c
-sudo: required
-dist: trusty
+language: minimal
 
 services:
   - docker
 
+env:
+  - DIST=fedora PKGMGR=dnf
+  - DIST=debian PKGMGR=aptcc
+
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,14 +1,15 @@
...
-sudo: required
-dist: trusty
+language: minimal
 
 services:
   - docker
 
+env:
+  - DIST=fedora PKGMGR=dnf
+  - DIST=debian PKGMGR=aptcc
```
kalev: Maybe BACKEND would be a better variable name here to match with -Dpackaging_backend?

Conan-Kudo: I suppose. I named it as such to make clear when glancing at the job overview page.

Issue #169, opened 2016-10-02
UpdatesChanged is not emitted when available updates list changes


antonio-rojas:
When the available updates list changes (eg. by running 'pkcon get-updates' when new updates are available) the UpdatesChanged signal is not emitted. This breaks update notifiers that rely on this signal (such as Plasma Discover).
This affects the alpm backend, I haven't tested other backends.

aleixpol:
I can reproduce this. Also with aptcc as well

Appadeia:
This issue should be closed since the fix is merged. Wonder why GitHub didn't pick up the mention as a "will close this issue" mention.

aleixpol:
I'd say it's fixed too for aptcc now, at least our users are getting the updates notifications. I'd say let's just close it for now.

PR #386, opened 2020-03-16
[backends/alpm] Emit UpdatesChanged when amount of available updates changes

Appadeia:
Fixes issue #169.


hughsie:
Can you squash your commits down please, and also prefix the commit subject with `alpm: ` please.

Appadeia:
Ping? Commits are squashed and commit subject is prefixed with `alpm: `.

PR #387, opened 2020-03-24
alpm: use separate alpm_handle_t for updates

Appadeia:
This prevents updating the system syncdbs without updating packages,
which often leads to undesirable behaviour on most alpm-based distros.

### DIFF
--- a/backends/alpm/pk-alpm-config.c
+++ b/backends/alpm/pk-alpm-config.c
@@ -24,6 +24,7 @@
 #include <alpm.h>
 #include <glob.h>
 #include <string.h>
+#include <sys/wait.h>
 #include <sys/utsname.h>
 #include <glib/gstdio.h>
 #include <syslog.h>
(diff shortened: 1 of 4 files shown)


backends/alpm/pk-alpm-config.c
```diff
@@ -922,8 +946,9 @@ pk_alpm_config_configure_alpm (PkBackend *backend, PkAlpmConfig *config, GError
 	g_return_val_if_fail (config != NULL, FALSE);
 
 	handle = pk_alpm_config_initialize_alpm (config, error);
-	if (handle == NULL)
+	if (handle == NULL) {
 		return NULL;
+	}
```
Conan-Kudo: Why is this style change being done? We don't typically do curly braces on one-line conditionals.

backends/alpm/pk-alpm-update.c
```diff
@@ -428,17 +429,35 @@ pk_backend_get_updates_thread (PkBackendJob *job, GVariant* params, gpointer p)
...
+			i = NULL;
+			break;
+		}
 
-	if (!pk_alpm_update_databases (job, 0, &error)) {
-		return pk_alpm_error_emit (job, error);
+		ret = pk_alpm_update_database (job, TRUE, i->data, error);
+		if (!ret) {
+			break;
+		}
```
Conan-Kudo: We don't typically use curly braces around one-liner conditionals.

PR #388, opened 2020-03-25
alpm: handle blank target name for progress cb

### LINKED ISSUE #173: [ALPM] itemProgress reported for unknown packages
As reported by `pkmon`:
```
/1211_bddeabcc  item-progress ,80 [unknown]
```

On PackageKitQt::Transaction::itemChanged I get a "" empty string itemID.

Appadeia:
When the target name is blank, the callback is generally for the
transaction in general and not for an unknown item. This handles
all instances where libalpm currently emits a progress callback with a blank
name.

Fixes #173.

### DIFF
--- a/backends/alpm/pk-alpm-transaction.c
+++ b/backends/alpm/pk-alpm-transaction.c
@@ -194,6 +194,34 @@ pk_alpm_transaction_progress_cb (alpm_progress_t type, const gchar *target,
 	g_assert (pkalpm_current_job);
 	job = pkalpm_current_job;
 
+	if (g_strcmp0(target, "") == 0) {
+		switch (type) {
+			case ALPM_PROGRESS_KEYRING_START:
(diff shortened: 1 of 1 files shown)


backends/alpm/pk-alpm-transaction.c
```diff
@@ -194,6 +194,34 @@ pk_alpm_transaction_progress_cb (alpm_progress_t type, const gchar *target,
 	g_assert (pkalpm_current_job);
 	job = pkalpm_current_job;
 
+	if (target == "") {
```
hughsie: does that work? Surely you mean NULL? Otherwise it probably needs to be `g_strcmp0`

Appadeia: The string isn't a null pointer in this case, but a `""` string.

aleixpol: Then you'll have to change it to `if (g_strcmp0(target, "") == 0) {` as @hughsie  said.

PR #405, opened 2020-06-11
pkcon: exit with retval 5 if no packages needed be installed

DimStar77:
In case a user asks pkcon to install an already installed package,
so far we returned with:

  Fatal error: The packages are already all installed
  (detailed string provided by the backend)

and retval 7.

Taking pkcon's manpage into account, retval 5 is more appropriate:

RETURN VALUES
       0
           Success

       5
           Nothing useful was done.

       7
           The transaction failed, see the detailed error for more information.


DimStar77:
The only thing here is that we still print

> Fatal error: The packages are already all installed

Maybe we should rather change that to something like

> Info: The packages are already all installed

The 2nd case used for retval 5 seems to be REPO_NOT_AVALAIBLE (where fatal: makes sense)

PR #406, opened 2020-06-24
Use statements for SQL queries with input.

jusa:


src/pk-transaction-db.c
```diff
@@ -288,83 +288,207 @@ pk_transaction_db_get_list (PkTransactionDb *tdb, guint limit)
...
 }
 
-gboolean
-pk_transaction_db_add (PkTransactionDb *tdb, const gchar *tid)
+static sqlite3_stmt*
+pk_transaction_db_prepare (PkTransactionDb *tdb, const gchar *sql)
 {
-	g_autofree gchar *timespec = NULL;
-	g_autofree gchar *statement = NULL;
+	sqlite3_stmt *statement = NULL;
```
hughsie, line 294: can we define an g_autoptr() here please; all the `goto out`'s look kinda bad.

src/pk-transaction-db.c
```diff
@@ -288,83 +288,207 @@ pk_transaction_db_get_list (PkTransactionDb *tdb, guint limit)
...
+out:
+	sqlite3_finalize (statement);
+
+	return rc == SQLITE_OK;
 }
 
-gboolean
-pk_transaction_db_set_cmdline (PkTransactionDb *tdb, const gchar *tid, const gchar *cmdline)
+static gboolean
+pk_transaction_db_set_ints2_strings1 (PkTransactionDb *tdb, const gchar *sql, gint first, gint second, const gchar *third)
```
hughsie, line 396: I really don't like this template style, sorry; I think sqlite3_bind_int just has to go in the caller.

jusa: That's fine, I just went a bit overboard with the functions. Other than those where two strings are applied these are not used elsewhere so no point in having separate functions.

PR #407, opened 2020-06-26
WIP: start porting alpm backend to adjust to breaking changes in alpm

pontaoski:


hughsie:
Are the alpm changes upstream already? Do you want to support both APIs with #ifdefs or just rely on the new version?

pontaoski:
> Are the alpm changes upstream already?

Yes; PackageKit alpm backend fails to build from source with current libalpm master due to API changes.

> Do you want to support both APIs with #ifdefs or just rely on the new version?

Given that Arch is rolling, I don't think it makes much sense to keep the old code path around.

eworm-de:
The alpm header file do not contain version information, so you can not detect at built time. It just builds or fails.

eworm-de:
Now that pacman 6.0.0 has been released... Do you want to finish your work?

progandy:
> The alpm header file do not contain version information, so you can not detect at built time. It just builds or fails.

If it is necessary to support the old and new library (e.g. some other distributions stay with old pacman), then it could be done with some meson configuration that converts the version number from pkg-config (```alpm_dep.version()```) to a c define. Either make it numeric like ```-DALPM_VER=XXXYYYZZZ``` or ifdef/undef with ```-DALPM_V13=1```

sidt4:
Guess this was addressed as part of https://github.com/PackageKit/PackageKit/pull/487, and can be closed.

PR #408, opened 2020-06-27
Fix #342 by ignoring the HookDir directive and updating meson version to prevent a minor warning

### LINKED ISSUE #342: PackageKit service fails with a core-dump due to unsupported directive HookDir in pacman.conf.
**PackageKit version**

> 1.1.12-1

**Used distribution**

 > Manjaro Stable

**Expected behaviour you didn't see**

 > PackageKit service should be able to parse a standard pacman.conf file without core dumping. Being able to deal with unknown directives could be the best solution for parameters that are not relevant to PackageKit working.

**Unexpected behaviour you saw**

 > The PackageKit service fails with a core dump when it encounters the unknown directive `HookDir`.

```
   Active: failed (Result: core-dump) since Thu 2019-09-26 16:52:37 -03; 2min 38s ago
  Process: 31837 ExecStart=/usr/lib/PackageKit/packagekitd (code=dumped, signal=TRAP)
 Main PID: 31837 (code=dumped, signal=TRAP)

set 26 16:52:37 erico-xps systemd[1]: Starting PackageKit Daemon...
set 26 16:52:37 erico-xps PackageKit[31837]: daemon start
set 26 16:52:37 erico-xps packagekitd[31837]: Failed to initialize alpm: using /etc/PackageKit/alpm.d/pacman.conf: /etc/PackageKit/alpm.d/pacman.conf:8/etc/pacman.conf:17unrecognised directive 'HookDir'
set 26 16:52:37 erico-xps systemd[1]: packagekit.service: Main process exited, code=dumped, status=5/TRAP
set 26 16:52:37 erico-xps systemd[1]: packagekit.service: Failed with result 'core-dump'.
set 26 16:52:37 erico-xps systemd[1]: Failed to start PackageKit Daemon.
```

**Steps to reproduce the problem**

You just need a `HookDir` directive in the `[options]` header in `/etc/pacman.conf`.

**Observation**

* This issue has been present since 2016, at least: https://bbs.archlinux.org/viewtopic.php?id=213009
* https://bugs.freedesktop.org/enter_bug.cgi?product=PackageKit doesn't allow reporting bugs for PackageKit, nor is it available on freedesktop's GitLab instance.
* I am interested in implementing the necessary fix, but I would need to know whether it's best to ignore all unknown directives or just make the `HookDir` directive "known".

MarcJose:
This provides one possible solution for the crashes when using pacman by ignoring the "HookDir" directive. Another possible solution would be to ignore all unknown directives but having atleast some control over it is probably preferred.
Also the minimum required meson version is updated to 0.50.0 to prevent a warning thrown during the build.

### DIFF
--- a/backends/alpm/pk-alpm-config.c
+++ b/backends/alpm/pk-alpm-config.c
@@ -598,6 +598,11 @@ pk_alpm_config_parse (PkAlpmConfig *config, const gchar *filename,
 			/* Ignore "Usage" key instead of crashing */
 			continue;
 		}
+		
+		if (g_strcmp0 (key, "HookDir") == 0 && str != NULL) {
+			/* Ignore "HookDir" key instead of crashing */
+			continue;
+		}
 
 		/* report errors from above */
 		g_set_error (&e, PK_ALPM_ERROR, PK_ALPM_ERR_CONFIG_INVALID,
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,7 @@
 project('PackageKit', 'c',
   version : '1.2.1',
   license : 'LGPL-2.1+',
-  meson_version : '>=0.47.0',
+  meson_version : '>=0.50.0',
   default_options : ['warning_level=2', 'c_std=c99'],
 )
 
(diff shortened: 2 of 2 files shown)


hughsie:
Not the meson fix, the warning is benign and bumping meson causes CI and stable distro issues.

PR #414, opened 2020-07-28
Update Nix backend to use Nix Flakes

matthewbauer:
This is a large overhaul of the Nix backend. It now is much faster
since it can make use of the eval caching that is available in Nix 2.4.


hughsie:
I have no problem with this, but do we know who is the current "maintainer" of the Nix backend for PK?

matthewbauer:
> I have no problem with this, but do we know who is the current "maintainer" of the Nix backend for PK?

I think it's probably me! I was the last to modify it at least.

hughsie (approved):
Looks good to me, but please squash before merging

PR #424, opened 2020-08-20
Fix setting libexecdir for command-not-found helper.

QuLogic:
Otherwise you end up with `libexec/...` in `PackageKit.sh` instead of `/usr/local/libexec/...`.

### DIFF
--- a/contrib/command-not-found/meson.build
+++ b/contrib/command-not-found/meson.build
@@ -13,7 +13,7 @@ executable(
 )
 
 bashprofile_config_data = configuration_data()
-bashprofile_config_data.set('LIBEXECDIR', get_option('libexecdir'))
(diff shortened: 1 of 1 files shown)


hughsie:
What distro? Isn't the other problem that you would get /usr/usr/libexec for building rpms for instance?

QuLogic:
No, `join_paths` will drop any paths before an absolute path. This method is already used all over your `meson.build` files for `*DIR` variable substitutions that need to be absolute.

QuLogic:
And [Fedora 33](https://bugzilla.redhat.com/1870438).

hughsie:
> No, join_paths will drop any paths before an absolute path

Ahh gotcha, thanks!

PR #432, opened 2020-09-18
Dependency handling by user.

nintyfan:
Some things worked at last, but you must install servicesUI (daemonUI) and Bonsole. There's still many things to do. For example, I prefer to find a way do not block packagekit daemon. You probably think rendering UI by daemon at all is bad idea, but it have many advantages. I am not sure it at all less secure. Take in mind we talking about dependency handling. What if discover or some other software ran by user take decision about how solve dependencies instead of user? It also allows to repair older software and delivering new possibilities to it. It could be better if I don't block packagekit, so I will search a way to achieve this. Another, but related thing is that currently my solution is only for single user. Bonsole library could be initialized only once, so I must add support for fork child process, which will use Bonsole library and communicate with Packagekit. I imagine, that If I do this, I could replace current code with non-blocking one. For example remember pipe and pid of child in job structure, do not send response just after zypp_perform_execution ends, but when task is really done. Cache code must be better - packagekit don't clean cache related to selection (it create cache file to user selection, so if user handle dependency it won't be asked second time for handling after password type). Cache also made easier to do non-blocking packagekit, because we could perform lock zypper, release lock when dependency error and user interaction is needed, render UI, wait for user interaction, lock package manager, perform the same test, release lock, render UI, ..., perform operation. I also look how to modify servicesUI to made possible non-blocking behavior. For example render UI by service UI and launch packagekit if user send form. There's many possible solutions.

I must admit I'm beginner C programmer. C++ is not my best skill. I'm waiting for any suggestions. That's not final merge request, but request for code review. I known my code look like a rubbish.


nintyfan:
I modify code, so it's now more behaves more correctly. I correct cache behavior, so user don't be forced to handle dependency twice and there's no wrong assumption cache file associate to current change. We cache user responses, so now my code behaves much better.

I still have many things to do. For example I must add code to remove cache files on exit and split UI code to different process, so we will have multiuser support and we could reduces privileges of UI by passing restrictions to UI process' rights.

README
```diff
@@ -5,3 +5,8 @@ PackageKit is a DBUS abstraction layer that allows the session user to manage
 packages in a secure way using a cross-distro, cross-architecture API.
 
 For more information, please see http://www.packagekit.org
+
```
AsciiWolf, line 8: This was probably intended to be for your personal fork, not the upstream README.

nintyfan:
Zypp backend needs only lifting up, I think. I need some advice about anothers backend - CI don't pass. And of course, mark my work, do any suggestions.

ximion:
One thing that should actually be mentioned: PackageKit has a policy of never asking questions once a transaction is triggered (previously also known as "hughsie's law"), and we go through great lengths to prevent that. The only exception is probably certain high-priority Debconf questions getting passed through on Debian, but everything else adheres to that rule, and it is also an essential requirement for offline updates.
I don't know what Bonsole is (yet), but if this change allows asking users to run through manual dialogs before or while a transaction is executed, it's very unlikely to get merged.
In the case of dependency resolution there is also the question of why a package manager would be unable to find a good solution on its own and apply that, instead of nagging the user about it. In other words, the package manager should be improved to find better solutions and "just do the right thing" instead of asking the user to do dependency resolution for it.

ximion:
As for the Ci failures: You are building the `dependency-solving-helper` binary unconditionally and therefore have an unconditional dependency on `libbonsole`, so it will always be built & linked (and therefore fail on the CI system, as `libbonsole` doesn't exist there).

nintyfan:
@ximion 
About hughsie's law and my series of patches. My work is indeed to only asks for select of solution before start download/install. Currently, I use a hack. PackageKit create a job object. In zypp backend I asks solver to solve dependency and if it's not satisfied, I start helper program, read all solve options and pass to this program and next set special flag to not done. We store job object in special field of zypp backend's struct and pass it as callback argument of some gio related function. In next step, I return from zypp_backend_perform_execution. The trick is related to checking in some previous function in stack our flag is set and do not end transaction. When user tell dependency solving helper, what to do, our gio function call zypper_perform_execution again with stored object. If all test passed, we set done flag to true, start transaction and exit. Now we can delete job object.

I saw PackageKit have function to test transaction is interactive. I suppose I must check value of this flag.

Bonsole's source are on source forge. No packages yet - only sources. Bonsole is a project, which are assemblies by client library and web browser with UI and etc. to support Bonsole apps. Bonsole means browser console, but it's support html form, buttons, etc. When UI isn't accessible, client library will load console backend and translates Bonsole's markup language to console tui app. Bonsole uses special markup language and xslt to translate it output for console/web browser backend (depending which backend are loaded).

About package manager. There could be many cases like updating some software could causing dependency problem. I often got problem with updating software in OpenSUSE Tumbleweed. Situation could be similar: package A request older version of libA, but package manager tries to update package B, which needs newer version of libA. When using of tray applet, in many cases I was unable to update my system.

nintyfan:
About the CI system. I learn meson/ninja currently, but I thought helper was disabled on my system, I added this line:
`option('dependency_handling_by_user', type : 'boolean', value : false)
`
To meson_options.txt
And these:
`if get_option('dependency_handling_by_user')
  subdir('helpers')
endif
`
At bottom on meson.build.
I will learn meson and try to correct this, but if you suggest me solution, I will be happy.

nintyfan:
@ximion:
Ok. Two tests passed, but there is some warning in compilation process. Should I remove it before further processing?

nintyfan:
@ximion:
I prepared Bonsole packages for OpenSUSE Tumbleweed. Go to software.opensuse.org, type BonsoleRTA and install each of three packages (of course, on VM). DaemonUI packages and Packages for other distro (especially for Fedora) are pending. There still some issues (no devel packages and you must type sudo ldconfig after installing), but I will solve each issue in end of next week, I hope.

PR #435, opened 2020-10-23
main: notify the service manager when it's beginning to shutdown

JonathanKang:
This makes sure that the main process won't get SIGTERM on shutdown,
for example timed exit after idling for a while. This fix the problem
that libzypp fails to clean up some temporary files when packagekitd
quits.

### DIFF
--- a/src/pk-main.c
+++ b/src/pk-main.c
@@ -32,6 +32,9 @@
 #include <glib-unix.h>
 #include <glib/gi18n.h>
 #include <packagekit-glib2/pk-debug.h>
+#ifdef HAVE_SYSTEMD_SD_DAEMON_H
+#include <systemd/sd-daemon.h>
+#endif
 
 #include "pk-engine.h"
 #include "pk-shared.h"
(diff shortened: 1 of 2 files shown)


src/pk-main.c
```diff
@@ -245,6 +246,8 @@ main (int argc, char *argv[])
 	syslog (LOG_DAEMON | LOG_DEBUG, "daemon quit");
 	closelog ();
 
+	sd_notify(0, "STOPPING=1");
```
hughsie, line 249: space before bracket please.

src/pk-main.c
```diff
@@ -32,6 +32,7 @@
 #include <glib-unix.h>
 #include <glib/gi18n.h>
 #include <packagekit-glib2/pk-debug.h>
+#include <systemd/sd-daemon.h>
```
hughsie, line 35: I think you'll need to add something like HAVE_SYSTEMD_SD_DAEMON_H around this otherwise you'll upset the usual crowd.

PR #437, opened 2020-10-29
Do not crash when calling pk_dbus_get_uid() before D-Bus is setup

hughsie:
Fixes https://github.com/hughsie/PackageKit/issues/436

### DIFF
--- a/src/pk-dbus.c
+++ b/src/pk-dbus.c
@@ -64,6 +64,10 @@ pk_dbus_get_uid (PkDbus *dbus, const gchar *sender)
 	g_return_val_if_fail (PK_IS_DBUS (dbus), G_MAXUINT);
 	g_return_val_if_fail (sender != NULL, G_MAXUINT);
 
+	/* no connection to DBus */
+	if (dbus->priv->proxy_uid == NULL)
+		return G_MAXUINT;
+
 	/* set in the test suite */
 	if (g_strcmp0 (sender, ":org.freedesktop.PackageKit") == 0) {
 		g_debug ("using self-check shortcut");


voxik:
Fedora report: https://bugzilla.redhat.com/show_bug.cgi?id=1836304

AdamWill:
From Rawhide testing, it seems like this maybe broke checking for updates on system start in GNOME. We have a test that runs an install, then boots to runlevel 3, downgrades python3-kickstart to a lower versioned dummy to ensure an update will be available, goes to the desktop, waits ten minutes, then opens the notifications area and expects to see an 'updates available' notification.

Up till Fedora-Rawhide-20201109.n.0 this test was passing fine. With Fedora-Rawhide-20201110.n.0, which included PackageKit-1.2.2-2.fc34 with this patch backported, the test [fails, because no notification appears](https://openqa.fedoraproject.org/tests/719599#step/desktop_notifications/31). I'll attach the tarball of /var/log from the failed test; it seems to show no update check running at all, AFAICT.

AsciiWolf:
@AdamWill Oops. :-( I did not notice this during my testing. Could you please also test this on F33 and write a comment [here](https://bugzilla.redhat.com/show_bug.cgi?id=1836304)?

AdamWill:
Filed a new bug and commented. Testing on F33 is a bit tricky, though, as this test is not set up to work on updates...

AsciiWolf:
@AdamWill By the way, what was the last PackageKit package version that was successfully tested on Rawhide before this (regressed) one? Was the regression really caused by this fix? :)

AsciiWolf:
I have reopened the [Fedora bz](https://bugzilla.redhat.com/show_bug.cgi?id=1836304) since the fix was unpushed from stable and the original issue was not fixed yet.

AdamWill:
Previous compose had 1.2.2-1.fc34.

AsciiWolf:
Hmm. I just tried the older PackageKit-1.2.2-1.fc33.x86_64 build on a F33 VM that I have here (that previously had 1.2.2-2 installed) and have the the same issue as with the 1.2.2-2 build: I don't seem to get any new updates notification. When I manually open GNOME Software, new updates are displayed correctly there, but that was also the case with 1.2.2-2 build. Are you really sure that it worked fine with the 1.2.2-1 build? Or am I doing anything wrong?

AdamWill:
I'm about 99% sure. Notifications are very tricky to test, because of all the 'clever' heuristics GNOME uses to decide when to show you one. The two main constraints are mentioned in [the code in the test that tries to make sure they're all satisfied](https://pagure.io/fedora-qa/os-autoinst-distri-fedora/blob/master/f/tests/desktop_notifications.pm#_28), but there may be some other I don't know or have forgotten about...

AsciiWolf:
Hm, it looks like that you are right. I tried downgrading PackageKit even more, to PackageKit-1.2.1-1, and still don't get any new update notifications. So either something else is broken on that system or it is the 'clever' GNOME heuristics in action. :)

AdamWill:
You're probably running into the second rule there: if you've had a notification in the last 24 hours, you won't get another. You can use the same gsettings dodge the test uses to try and counteract that.

AsciiWolf:
Hmm, I have tried reverting the VM to a clean after-install (Fedora 33 Beta) state and got new updates notification. Then I have fully updated it to latest F33 state and tried manually downgrading some packages (gnome-boxes etc.), changing the `check-timestamp` gsettings and rebooting/relogging. But at this point, I was not able to get any new update notification at all although I have tried this multiple times. Available updates are displayed fine when I open GNOME Software, but no update notifications appear. So I am beginning to think that this **is not** a PackageKit issue (older PackageKit-1.2.1-1 is being used) and not an issue caused by this PR.

@AdamWill Is it possible to run the desktop notification test again, this time on a fully updated Rawhide system using latest packages, but with a slightly older PackageKit build (1.2.2-1 or 1.2.1-1)? If so, could you please do it? Thanks!

AdamWill:
[robots are on the job](https://openqa.stg.fedoraproject.org/tests/overview?groupid=1&build=Fedora-Rawhide-20201118.n.0-notetest-NOREPORT&version=Rawhide&distri=fedora)

I noticed the test *is* passing occasionally, annoyingly. So let's throw numbers at the problem. That's an openQA flavor which is basically running the same test 20 times; 10 times as usual, and 10 times with an extra step downgrading PackageKit to PackageKit-1.2.2-1.fc34 . It's running on today's Rawhide. Results coming soon...

AdamWill:
Hmm - so for that test we get some passes and some failures both with and without the mod. Here's [another check](https://openqa.stg.fedoraproject.org/tests/overview?distri=fedora&version=Rawhide&build=Fedora-Rawhide-20201109.n.0-notetest-NOREPORT&groupid=1) - same thing with an older Rawhide, the day before I thought the problem first happened. Let's see how that goes.

AdamWill:
Hmm, similar result. Trying again with the oldest successful compose still around (20201107.n.1) [here](https://openqa.stg.fedoraproject.org/tests/overview?distri=fedora&version=Rawhide&build=Fedora-Rawhide-20201107.n.1-notetest-NOREPORT&groupid=1).

AdamWill:
Same result there. So I'd say we at least have no clear indication 1.2.2-2 is the issue here, so I'm gonna say we should go ahead and push the update for F33 since the crash is clearly annoying people. I'll try and look further into why this test might suddenly have started having trouble tomorrow.

hughsie:
I do wonder if libdnf is causing a regression somewhere.

AsciiWolf:
Hmm, so I did more testing again and not sure what I did wrong before, but I now seem to get new update notifications without any issue in my F33 testing VM with latest updates (and gnome-boxes/podman/etc. manually downgraded to test the update notifications). Even when using latest libdnf-0.54.2-3.fc33.x86_64 and PackageKit-1.2.2-2.fc33.x86_64.

So, it seems that there is probably no regression (although I am still not 100% sure), just some issue with the Rawhide test.

AsciiWolf:
The only difference when compared to my previous testing is that I also downgraded podman which is considered as an OS Update by GNOME Software, unlike gnome-boxes/gnome-maps/other GNOME apps that I tried downgrading before. But I think that this should not have any effect on notifications.

Anyway, tl;dr: Both the latest libdnf and PackageKit builds seem to be fine, at least on F33.

AdamWill:
There definitely is *some kind* of regression somewhere, as the test ran reliably for years: https://openqa.fedoraproject.org/tests/725227#next_previous - note the two other recent failure blocks were not due to update notifications not appearing, but other issues. I'll try and figure out what's really going on today.

AdamWill:
@AsciiWolf can you see how downgrading fwupd affects you? See https://bugzilla.redhat.com/show_bug.cgi?id=1896540#c6 ...

AdamWill:
So some tests I ran in openQA make me about 75% confident that backporting https://github.com/fwupd/fwupd/pull/2605 helps, here. At least it doesn't seem like it could hurt, so I built it for Rawhide and submitted updates for F32 and F33. I hacked the disk image that openQA update tests run from to include the update, and will re-run the recent failed tests; I'll keep an eye on things over the weekend to see if this really does improve the problems.

hughsie:
> that backporting fwupd/fwupd#2605 helps

@hughsie looks guiltily at the floor; thanks @AdamWill

AdamWill:
So it does look like backporting that fix was the thing to do, but we may now have broken fwupd for Secure Boot-enabled systems because I'm not on the SB signing list, whoops. hughsie has just built an fwupd 1.5.2 so we'll want to get those updates pushed stable ASAP then hopefully everything should be OK.

PR #438, opened 2020-11-03
zypp: Implement import\remove public key feature

etaishev:
Implementation of the interface for calling zypp import\remove public key


hughsie:
Back a bit; why is this something we want users to handle?

etaishev:
Only for the convenience of key management

Conan-Kudo:
I didn't even know you can do this in PackageKit, but even though you can, none of the PK frontends would be able to do anything with this, so I'm not sure it's worth implementing...

PR #447, opened 2021-01-11
Fix all compiler warnings

ximion:
I did a bit of housekeeping.
Pretty much all of these changes are trivial, but it's better to have someone else just have a quick look at this patch, just in case I missed something :-)


hughsie (approved):
No idea on the C++ stuff, rest looks fine, thanks.

ximion:
I am very confident about the C++ stuff - mostly symbol reordering and dead code removal :-)

PR #448, opened 2021-01-14
zypp: Fix package installation using undefined data

pvuorela:
Code introduced by commit c45e0f22df682 is strange.
Depending partly on uninitialized array values when an array
is not even needed, and also using the values both as enum
and as some kind of existence check.

cc: @JonathanKang

### DIFF
--- a/backends/zypp/pk-backend-zypp.cpp
+++ b/backends/zypp/pk-backend-zypp.cpp
@@ -2740,7 +2740,6 @@ backend_install_packages_thread (PkBackendJob *job, GVariant *params, gpointer u
 		PoolStatusSaver saver;
 		pk_backend_job_set_percentage (job, 10);
 		vector<PoolItem> items;
-		VersionRelation relations[g_strv_length (package_ids)];
 		guint to_install = 0;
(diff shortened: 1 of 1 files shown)


pvuorela:
Ping, long time, no feedback. Without this I wouldn't trust the zypp backend to install anything.

pvuorela:
Ping again @JonathanKang @hughsie

This is relatively simple and doesn't even need much knowledge on the surrounding code. The array created and read as first thing is getting uninitialized data -> start by having the initial status explicitly. Array used always with [i] -> no need for array at all. 

Been patched in SailfishOS now for nearly a year.

hughsie:
I have no C++ knowledge I'm afraid. @dantti ?

ximion:
I can't comment on the API usage of Zypper (I know nothing about it), but C++-wise this is fine, and the previous code is indeed odd - depending on uninitialized array values will absolutely blow up (if it doesn't already do that), so this PR makes sense to me in that regard.

hughsie:
Good enough for me, thanks.

PR #449, opened 2021-01-19
dnf: Add support for coercing upgrade to distupgrade

Conan-Kudo:
Some distributions (particularly rolling release distributions)
prefer that normal "upgrade" actions actually trigger the distupgrade
method in the package manager. This change enables that if the
installed  distribution release package specifies that this
behavior must occur.

### DIFF
--- a/backends/dnf/pk-backend-dnf.c
+++ b/backends/dnf/pk-backend-dnf.c
@@ -823,6 +823,28 @@ dnf_utils_run_query_with_newest_filter (DnfSack *sack, HyQuery query)
 	return results;
 }
 
+static gboolean
+dnf_utils_force_distupgrade_on_upgrade (DnfSack *sack)
+{
+	g_autoptr(GPtrArray) plist = NULL;
+	gint candidates;
+	const gchar *distroverpkg_names[] = { "system-release", "distribution-release", NULL };
+	const gchar *distupgrade_provides[] = { "system-upgrade(dsync)", "product-upgrade() = dup", NULL };
+	HyQuery query_tmp = hy_query_create (sack);
+
+	hy_query_filter (query_tmp, HY_PKG_REPONAME, HY_EQ, HY_SYSTEM_REPO_NAME);
+	hy_query_filter_provides_in (query_tmp, (gchar**) distroverpkg_names);
+	hy_query_filter_provides_in (query_tmp, (gchar**) distupgrade_provides);
+
+	plist = hy_query_run (query_tmp);
+	candidates = plist->len;
+	hy_query_free (query_tmp);
+
+	if (candidates > 0)
+		return TRUE;
+	return FALSE;
+}
+
 static GPtrArray *
 dnf_utils_run_query_with_filters (PkBackendJob *job, DnfSack *sack,
 				  HyQuery query, PkBitfield filters)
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
@hellcp, @j-mracek: Could you please review this?

backends/dnf/pk-backend-dnf.c
```diff
@@ -823,6 +823,25 @@ dnf_utils_run_query_with_newest_filter (DnfSack *sack, HyQuery query)
...
+	const gchar* distupgrade_provides[] = { "system-upgrade(dsync)", "product-upgrade() = dup", NULL };
+	hy_query_filter_provides_in(query, (gchar**) distupgrade_provides);
+	GPtrArray *plist = hy_query_run(query);
+	gint candidates = plist->len;
+	g_ptr_array_unref(plist);
+
+	if (candidates > 0) {
+		return TRUE;
+	}
+	return FALSE;
```
pontaoski, line 843:
```suggestion
	return candidates > 0;
```

no need to do `if (X) return TRUE; return FALSE;`

backends/dnf/pk-backend-dnf.c
```diff
@@ -823,6 +823,26 @@ dnf_utils_run_query_with_newest_filter (DnfSack *sack, HyQuery query)
...
+	HyQuery query_tmp = hy_query_create (sack);
+	hy_query_filter (query_tmp, HY_PKG_REPONAME, HY_EQ, HY_SYSTEM_REPO_NAME);
+	const gchar* distroverpkg_names[] = { "system-release", "distribution-release", NULL };
+	hy_query_filter_provides_in (query_tmp, (gchar**) distroverpkg_names);
+	const gchar* distupgrade_provides[] = { "system-upgrade(dsync)", "product-upgrade() = dup", NULL };
+	hy_query_filter_provides_in (query_tmp, (gchar**) distupgrade_provides);
+	GPtrArray *plist = hy_query_run (query_tmp);
+	gint candidates = plist->len;
+	g_ptr_array_unref (plist);
+	hy_query_free (query_tmp);
```
pontaoski, line 838:
```suggestion

	const gchar* distroverpkg_names[] = { "system-release", "distribution-release", NULL };
	hy_query_filter_provides_in (query_tmp, (gchar**) distroverpkg_names);
	
	const gchar* distupgrade_provides[] = { "system-upgrade(dsync)", "product-upgrade() = dup", NULL };
	hy_query_filter_provides_in (query_tmp, (gchar**) distupgrade_provides);

	GPtrArray *plist = hy_query_run (query_tmp);
	gint candidates = plist->len;
	g_ptr_array_unref (plist);
	hy_query_free (query_tmp);
```

nit: readability

pontaoski (changes requested):

backends/dnf/pk-backend-dnf.c
hughsie, line 838: can we use g_autoptr() here?

hellcp (approved):

dmach:
I think this change in behavior makes sense.
I just want to point out several things to think about:
* dnf_utils_force_distupgrade_on_upgrade should be possibly moved to libdnf
* it seems the query detecting the upgrade method reads all packages (both installed and available); it would be good to use data from an installed package and if there's no such package installed (an empty chroot), use a package from the available as a fallback; we don't want enabling e.g. Rawhide repo to change the upgrade policy
* it would be also good to have a PackageKit or dnf.conf configuration option to override this behavior if needed

backends/dnf/pk-backend-dnf.c
Conan-Kudo: Does that work with non-GObject data types?

hughsie: Well, `plist` is :)

Conan-Kudo:
@dmach Yeah, it probably makes sense to move some of this into libdnf later, and when we do that, we should make this behavior configurable from `dnf.conf`.

backends/dnf/pk-backend-dnf.c
Conan-Kudo: Do we do this normally anywhere else in the codebase? I didn't see this style of implicit conditional return being used elsewhere.

backends/dnf/pk-backend-dnf.c
Conan-Kudo: Done!

backends/dnf/pk-backend-dnf.c
Conan-Kudo: Other parts of this file don't do that, so I'm keeping this as is...

hughsie (approved):

PR #453, opened 2021-02-04
Rename errno to alpm_err to avoid name clashes

conqp:
The current variable name `errno` results in a name clash when building the ALPM backend.
Renaming it to `alpm_err` resolved the issue.


hughsie (approved):

pontaoski (approved):
LGTM

PR #464, opened 2021-03-19
Fix hangs in packagekit-glib2 client if daemon crashes

pwithnall:
See the commit messages for details.

This may fix https://gitlab.gnome.org/GNOME/gnome-software/-/issues/1118 (but I’m not entirely certain about it).

### DIFF
--- a/lib/packagekit-glib2/pk-client.c
+++ b/lib/packagekit-glib2/pk-client.c
@@ -123,6 +123,9 @@ typedef struct {
 } PkClientState;
 
 static void
+pk_client_state_finish (PkClientState *state,
+                        const GError  *error);
+static void
 pk_client_properties_changed_cb (GDBusProxy *proxy,
 				 GVariant *changed_properties,
 				 const gchar* const  *invalidated_properties,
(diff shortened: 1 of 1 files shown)


pwithnall:
Currently only compile-tested. I’m testing it now.

pwithnall:
Seems to fix the case of multiple hung calls to `pk_client_refresh_cache()` queueing up in multiple worker threads in gnome-software if I call `sudo systemctl kill packagekit.service` while gnome-software is searching for updates (thanks to Milan for the reproducer instructions).

ximion:
The patch overall looks good to me! (I haven't tested it locally though)

I wonder whether there is any value in returning a dedicated "internal error"/"daemon crashed" error type here, so the GUI can show an appropriate message telling the user the failure wasn't their fault and they should try again or report a bug (as "daemon crash" is rather cryptic). On the other hand though, any such message would also be kind of horrible in a GUI to have in general, so I'm not sure if we need this or if it's even a good idea.

hughsie:
I think we should get a nag from something like abrt in this case async from what PK does. Good work @pwithnall !

hughsie:
> Seems to fix the case of multiple hung calls

I had the suspicion this was the cause a few weeks ago, but never got around to debugging it fully. I really appreciate the help! I'll do a tarball next week and we can put it in Fedora.

PR #468, opened 2021-05-01
Fix use of POLKIT_CHECK_AUTHORIZATION_FLAGS_ALLOW_USER_INTERACTION

pwithnall:
PackageKit methods may be called in all sorts of contexts — as a result of an explicit user action, or as a background operation. Sometimes PackageKit will need to check a polkit action, and sometimes this will result in a polkit password prompt. Showing those prompts unexpectedly for background operations is quite surprising for users; plumb through the right D-Bus call flags and `PkTransaction` hints to try and ensure that they are only shown for PackageKit methods which are called as a result of explicit user action.

I have tested that this compiles, but no further than that.

See also:
 * https://gitlab.gnome.org/GNOME/gnome-software/-/issues/582
 * https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/726 (this PR is complementary, but independent)

### DIFF
--- a/lib/packagekit-glib2/pk-client.c
+++ b/lib/packagekit-glib2/pk-client.c
@@ -4744,6 +4744,10 @@ void
 pk_client_set_locale (PkClient *client, const gchar *locale)
 {
 	g_return_if_fail (PK_IS_CLIENT (client));
+
+	if (g_strcmp0 (client->priv->locale, locale) == 0)
+		return;
+
 	client->priv->locale = g_strdup (locale);
 	g_object_notify (G_OBJECT (client), "locale");
 }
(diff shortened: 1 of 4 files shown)


ximion:
I also didn't test this yet, but purely from reading the code and PolKit docs this looks good to me - and the behavior of "a random permission dialog pops up" is actually super annoying and confusing for users (I have seen this multiple times in the wild, and it even confused me...).
One problem may be that if, for example, cache refreshes require user interaction via PolKit policy, they may silently never happen now. For example, Debian only permits them for users with a local login, and if you connect remotely to the system's GUI, you get a permission dialog. Admittedly this is probably not a frequent configuration, and the system may have other measures in place to update the caches.

So, in general +1, we should probably test this though ;-)
Also, nice maintenance commits - thanks for finding & fixing typos and a bit of dead code!

hughsie:
Looks good to me, thanks!

PR #471, opened 2021-05-13
Wip/hughsie/client crashes

hughsie:


pwithnall:
The fix for the use-after-free (which does look like a correct fix) completely undoes #464, but that’s not mentioned in the commit message so I think it might be an oversight.

Perhaps a combination fix would be something like this (untested):
```
/* get the result */
value = g_dbus_proxy_call_finish (proxy, res, &error);
if (value == NULL) {
	/* Instructing the daemon to cancel failed, so just return an
	 * error to the client so they don’t wait forever. */
	g_debug ("failed to cancel: %s", error->message);
	if (error->domain == G_DBUS_ERROR || error->domain == G_IO_ERROR) {
		g_prefix_error (&error, "Failed to cancel: ");
		pk_client_state_finish (state, error);
	}
	return;
}
```

The distinction being that if the `Cancel()` call failed on the D-Bus client side, the service won’t have received the `Cancel()` method call, and hence won’t send any kind of reply which would call `pk_client_state_finish()` via a different path. If the `Cancel()` call failed on the service side, it may have done.

An alternative condition we could check for would be `!g_error_matches (error, PK_TRANSACTION_ERROR, PK_TRANSACTION_ERROR_NOT_RUNNING)`.

On the other hand, my fix from #464 was based on code inspection rather than actually being able to reproduce an issue with that commit, so might be completely invalid.

pwithnall:
@mcrha did you have any thoughts about the above?

mcrha:
> The fix for the use-after-free (which does look like a correct fix) completely undoes #464, but that’s not mentioned in the commit message so I think it might be an oversight.

Yes, it is an oversight, I didn't look for "where that thing has got from" at all.

I do not recall many details on this change, it's too long time ago for me. I only know the cancel callback could be called when the 'state' was already freed, thus causing use-after-free. Having the state ref-counted would avoid that.

With your proposal above you still do not have any assurance the pk_client_cancel_cb() is called with a valid 'state' structure. I guess.

mcrha:
I opened https://github.com/PackageKit/PackageKit/issues/489 as the follow up. That will need some input from the maintainers.

PR #475, opened 2021-05-20
PkPackage: Add update severity property

mcrha:
The DNF backend can read update severity, beside the update type.
This change passes the value to the PkPackage object, thus the caller
can read it. It's set when the caller asks for the updates.

----------------------------------------------------------------------

This change has incorporated notes from https://github.com/hughsie/PackageKit/pull/474

### DIFF
--- a/backends/dnf/dnf-backend.c
+++ b/backends/dnf/dnf-backend.c
@@ -32,15 +32,20 @@
 void
 dnf_emit_package (PkBackendJob *job, PkInfoEnum info, DnfPackage *pkg)
 {
+	PkInfoEnum update_severity;
+
+	update_severity = GPOINTER_TO_UINT (g_object_get_data (G_OBJECT (pkg), PK_DNF_UPDATE_SEVERITY_KEY));
+
 	/* detect */
 	if (info == PK_INFO_ENUM_UNKNOWN)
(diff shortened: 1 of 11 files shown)


Conan-Kudo:
Can you please split the general changes and the DNF backend changes into separate commits?

mcrha:
Will it help to anything? Adding a property, which is not used anywhere, feels like a useless change. Anyway, I'll do it. Hopefully not breaking the change.

mcrha:
Done.

Conan-Kudo (approved):

Conan-Kudo:
I'll need to run the tests locally since Travis CI isn't running, but it looks okay to me. I'll merge it if the tests pass locally. 👍🏾

mcrha:
Thanks. Is the reference for "since 1.2.4" correct? To be honest, I'd like to have it in the next release, which, I hope, will not be in a distant future.

Conan-Kudo:
Yes, that's correct!

Conan-Kudo:
@mcrha Can you rebase this on current `main` branch? We replaced Travis CI with GitHub Actions now...

Conan-Kudo:
Merged, thanks for the patch!

PR #476, opened 2021-05-20
Install size versus download size in PkDetails

mcrha:
This consists of two changes:
* the first changes the dnf backend to use the install size, instead of the package size, in the PkDetails; two reasons for it: a) the documentation mentions it'll be an install size for installed packages; b) the user is more interested in the install size than the package size;
* related to the previous point, add "download-size" into the PkDetails - that removes the ambiguity of the PkDetails::size whether it's installed or package size, because there's no property about it in the PkDetails (one can figure out whether the package is installed or not, but it means running additional commands, which are unnecessary). It also can provide an information whether the package is ready to be installed (its download size is 0).

--------------------------------------------------------------

This has incorporated a comment from https://github.com/hughsie/PackageKit/pull/474

### DIFF
--- a/backends/dnf/pk-backend-dnf.c
+++ b/backends/dnf/pk-backend-dnf.c
@@ -1916,14 +1916,15 @@ backend_get_details_thread (PkBackendJob *job, GVariant *params, gpointer user_d
 		pkg = g_hash_table_lookup (hash, package_ids[i]);
 		if (pkg == NULL)
 			continue;
-		pk_backend_job_details (job,
-					package_ids[i],
-					dnf_package_get_summary (pkg),
-					dnf_package_get_license (pkg),
-					PK_GROUP_ENUM_UNKNOWN,
-					dnf_package_get_description (pkg),
-					dnf_package_get_url (pkg),
-					(gulong) dnf_package_get_installsize (pkg));
+		pk_backend_job_details_full (job,
+					     package_ids[i],
+					     dnf_package_get_summary (pkg),
+					     dnf_package_get_license (pkg),
+					     PK_GROUP_ENUM_UNKNOWN,
+					     dnf_package_get_description (pkg),
+					     dnf_package_get_url (pkg),
+					     (gulong) dnf_package_get_installsize (pkg),
+					     dnf_package_is_downloaded (pkg) ? 0 : dnf_package_get_downloadsize (pkg));
 	}
 
 	/* done */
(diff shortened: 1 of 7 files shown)


Conan-Kudo (approved):

Conan-Kudo:
Like #475, I'll test this locally and merge it if it goes well. 👍🏾

hughsie:
@Conan-Kudo you don't want to port PK to github CI do you? :)

Conan-Kudo:
> @Conan-Kudo you don't want to port PK to github CI do you? :)

It's probably the next thing I'll do, actually. 😉

hughsie:
> It's probably the next thing I'll do, actually. wink

If you do, add yourself to AUTHORS too :)

mcrha:
Neal, on a related note, I just found https://gitlab.gnome.org/GNOME/gnome-software/-/issues/129#note_1120140 , do you think the proposed change would make sense or I should just close the Software bug? The change there only suggests how it can work, there are more things to be done to have it usable (+/- what is acceptable here).

Conan-Kudo:
Wouldn't implementing `dependson` and `whatdepends` also solve that problem?

mcrha:
No idea, I tried this and it seems to return correct things. I can try that too. There are some glitches, like libdnf claiming I've not installed the dependency package, even when it's installed, but I didn't dive to that (yet). I'm new to these things.

Conan-Kudo:
No worries, if you need help, let me know. 👍🏾

mcrha:
> Wouldn't implementing `dependson` and `whatdepends` also solve that problem?

A very quick search didn't reveal anything in the libdnf of that name. Is that a dnf command or anything else?

Conan-Kudo:
The CLI command for `dependson` is `dnf repoquery --depends <package>` and the equivalent for `whatdepends` is `dnf repoquery --whatdepends <package>`.

mcrha:
I see. Thanks. I'm afraid this will need even more time and resources to solve the dependencies, while the existing libdnf API provides the ways to get the dependencies. Even it's not exactly the same as `dnf install PACKAGE`, which resolves needed dependencies to be installed better than the patch I attached into the GNOME Software bug report. I suppose you are in favor of the change, thus I can look on this later and open a new issue/pull request, to not polute this merge request even more, with a semi-related thing.

One last question, would it make sense to provide the dependency sizes in a separate values on the PkDetails, that being `dependencies-download-size` and `dependencies-install-size`?

Conan-Kudo:
> I see. Thanks. I'm afraid this will need even more time and resources to solve the dependencies, while the existing libdnf API provides the ways to get the dependencies. Even it's not exactly the same as `dnf install PACKAGE`, which resolves needed dependencies to be installed better than the patch I attached into the GNOME Software bug report. I suppose you are in favor of the change, thus I can look on this later and open a new issue/pull request, to not polute this merge request even more, with a semi-related thing.
> 
> One last question, would it make sense to provide the dependency sizes in a separate values on the PkDetails, that being `dependencies-download-size` and `dependencies-install-size`?

I would think so.

Conan-Kudo:
Thanks for the patch, merged!

mcrha:
> I suppose you are in favor of the change, thus I can look on this later and open a new issue/pull request

I gave this some testing and I do not like the libdnf returns misleading numbers, due to not recognizing some packages as installed. I know it's a bug on their side, but it makes me hesitate to implement this.

I wanted to let you know I did not forget of it.

Issue #477, opened 2021-05-20
packagekit-direct with backends that use pk-backend-spawn?


imz:
I gave a try to the `packagekit-direct` executable and found out that it can't use backends that use functions from `pk-backend-spawn.h`, e.g.:

```
# /usr/lib/packagekit-direct refresh
Failed to load the backend: opening module aptcc failed : /usr/lib64/packagekit-backend/libpk_backend_aptcc.so: undefined symbol: pk_backend_spawn_set_name
```

The build rules exclude the corresponding sources actually:

```
packagekit_direct_exec = executable(
  'packagekit-direct',
  'pk-backend.c',
  'pk-backend.h',
  'pk-backend-job.c',
  'pk-backend-job.h',
  'pk-direct.c',
  'pk-shared.c',
  'pk-shared.h',
  dependencies: [
    packagekit_glib2_dep,
    libsystemd,
    elogind,
    gmodule_dep,
  ],
  install: true,
  install_dir: get_option('libexecdir'),
  c_args: [
    '-DG_LOG_DOMAIN="PackageKit"',
    '-DLIBDIR="@0@"'.format(join_paths(get_option('prefix'), get_option('libdir'))),
    '-DSYSCONFDIR="@0@"'.format(get_option('sysconfdir')),
    '-DVERSION="@0@"'.format(meson.project_version()),
    '-DGETTEXT_PACKAGE="@0@"'.format(meson.project_name()),
    '-DPACKAGE_LOCALE_DIR="@0@"'.format(package_locale_dir),
  ]
)
```

--as opposed to the ones for `packagekitd`:

```
shared_sources = files(
  'pk-dbus.c',
  'pk-dbus.h',
  'pk-transaction.c',
  'pk-transaction.h',
  'pk-transaction-private.h',
  'pk-backend.c',
  'pk-backend.h',
  'pk-backend-job.c',
  'pk-backend-job.h',
  'pk-shared.c',
  'pk-shared.h',
  'pk-spawn.c',
  'pk-spawn.h',
  'pk-engine.h',
  'pk-engine.c',
  'pk-backend-spawn.h',
  'pk-backend-spawn.c',
  'pk-scheduler.c',
  'pk-scheduler.h',
  'pk-transaction-db.c',
  'pk-transaction-db.h',
)

...

packagekitd_exec = executable(
  'packagekitd',
  'pk-main.c',
  shared_sources,
  pk_resources,
  dependencies: [
    packagekit_glib2_dep,
[...]

PR #480, opened 2021-05-21
packagekit-direct: make backends that use pk-backend-spawn work (#477)

imz:
The packagekit-direct executable couldn't use backends that use
functions from pk-backend-spawn.h, e.g., the APT backend:

# /usr/lib/packagekit-direct refresh
Failed to load the backend: opening module aptcc failed : /usr/lib64/packagekit-backend/libpk_backend_aptcc.so: undefined symbol: pk_backend_spawn_set_name


imz:
Fixes: https://github.com/PackageKit/PackageKit/issues/477

Issue #484, opened 2021-06-01
Packagekit failed to start, after enabling new pacman "ParallelDownloads" options


ariefra:
After updating to pacman 6.0 and enable newly introduce ParallelDownloads option in /etc/pacman.conf,
PackageKit failed to start with error message:
"Failed to initialize alpm: using /etc/PackageKit/alpm.d/pacman.conf: /etc/PackageKit/alpm.d/pacman.conf:8/etc/pacman.conf:37unrecognised directive 'ParallelDownloads'"

Temporary work around, disable ParallelDownloads option.

smac89:
While we are on the topic, this one has been open since 2019:
https://github.com/PackageKit/PackageKit/issues/342

sidt4:
Not an issue anymore. `ParallelDownloads` works fine.

PR #487, opened 2021-06-11
Pacman 6.0.0

eworm-de:
This is the current code we use to build the Arch Linux package `packagekit` `1.2.3-4`.

### DIFF
--- a/backends/alpm/meson.build
+++ b/backends/alpm/meson.build
@@ -1,4 +1,4 @@
-alpm_dep = dependency('libalpm', version: '>=12.0.0')
+alpm_dep = dependency('libalpm', version: '>=13.0.0')
 
 shared_module(
   'pk_backend_alpm',
--- a/backends/alpm/pk-alpm-config.c
+++ b/backends/alpm/pk-alpm-config.c
@@ -870,7 +870,7 @@ pk_alpm_spawn (const gchar *command)
 }
 
 static gint
-pk_alpm_fetchcb (const gchar *url, const gchar *path, gint force)
+pk_alpm_fetchcb (void *ctx, const gchar *url, const gchar *path, gint force)
 {
 	GRegex *xo, *xi;
 	gint result = 0;
@@ -969,9 +969,9 @@ pk_alpm_config_configure_alpm (PkBackend *backend, PkAlpmConfig *config, GError 
 	config->xfercmd = NULL;
 
 	if (xfercmd != NULL) {
-		alpm_option_set_fetchcb (handle, pk_alpm_fetchcb);
+		alpm_option_set_fetchcb (handle, pk_alpm_fetchcb, NULL);
 	} else {
-		alpm_option_set_fetchcb (handle, NULL);
+		alpm_option_set_fetchcb (handle, NULL, NULL);
 	}
 
 	/* backend takes ownership */
--- a/backends/alpm/pk-alpm-config.c
+++ b/backends/alpm/pk-alpm-config.c
@@ -807,7 +807,7 @@ static gboolean
 pk_alpm_config_configure_repos (PkBackend *backend, PkAlpmConfig *config,
 				   alpm_handle_t *handle, GError **error)
 {
-	alpm_siglevel_t base, level, mask, local, remote;
+	alpm_siglevel_t base, level = 0, mask = 0, local, remote;
 	const alpm_list_t *i;
 	PkAlpmConfigSection *options;
 
--- a/backends/alpm/pk-alpm-config.c
+++ b/backends/alpm/pk-alpm-config.c
@@ -635,8 +635,10 @@ pk_alpm_config_initialize_alpm (PkAlpmConfig *config, GError **error)
 	}
 
 	if (config->is_check) {
+		gchar* path;
+
 		g_free(config->dbpath);
-		gchar* path = g_strconcat (config->root,
+		path = g_strconcat (config->root,
 						 "/var/lib/PackageKit/alpm" + dir,
 						 NULL);
 		g_mkdir_with_parents(path, 0755);
--- a/backends/alpm/pk-alpm-config.c
+++ b/backends/alpm/pk-alpm-config.c
@@ -40,7 +40,7 @@ static gchar *xfercmd = NULL;
 typedef struct
 {
 	 gboolean		 checkspace, color, disabledownloadtimeout, ilovecandy,
-				totaldl, usesyslog, verbosepkglists, is_check;
+				noprogressbar, totaldl, usesyslog, verbosepkglists, is_check;
 
 	 gchar			*arch, *cleanmethod, *dbpath, *gpgdir, *logfile,
 				*root, *xfercmd;
@@ -147,6 +147,14 @@ pk_alpm_config_set_ilovecandy (PkAlpmConfig *config)
 }
 
 static void
+pk_alpm_config_set_noprogressbar (PkAlpmConfig *config)
+{
+	g_return_if_fail (config != NULL);
+
+	config->noprogressbar = TRUE;
+}
+
+static void
 pk_alpm_config_set_totaldl (PkAlpmConfig *config)
 {
 	g_return_if_fail (config != NULL);
@@ -182,6 +190,7 @@ static const PkAlpmConfigBoolean pk_alpm_config_boolean_options[] = {
 	{ "Color", pk_alpm_config_set_color },
 	{ "DisableDownloadTimeout", pk_alpm_config_set_disabledownloadtimeout },
 	{ "ILoveCandy", pk_alpm_config_set_ilovecandy },
+	{ "NoProgressBar", pk_alpm_config_set_noprogressbar },
 	{ "TotalDownload", pk_alpm_config_set_totaldl },
 	{ "UseSyslog", pk_alpm_config_set_usesyslog },
 	{ "VerbosePkgLists", pk_alpm_config_set_verbosepkglists },
@@ -596,6 +605,11 @@ pk_alpm_config_parse (PkAlpmConfig *config, const gchar *filename,
 
 		if (g_strcmp0 (key, "Usage") == 0 && str != NULL) {
 			/* Ignore "Usage" key instead of crashing */
+			continue;
+		}
+
+		if (g_strcmp0 (key, "ParallelDownloads") == 0 && str != NULL) {
+			/* Ignore "ParallelDownloads" key instead of crashing */
 			continue;
 		}
 
--- a/backends/alpm/pk-alpm-config.c
+++ b/backends/alpm/pk-alpm-config.c
@@ -47,7 +47,7 @@ typedef struct
 
 	 alpm_list_t		*cachedirs, *holdpkgs, *ignoregroups,
 				*ignorepkgs, *localfilesiglevels, *noextracts,
-				*noupgrades, *remotefilesiglevels;
+				*noupgrades, *remotefilesiglevels, *hookdirs;
 
 	 alpm_list_t		*sections;
 	 GRegex			*xrepo, *xarch;
@@ -106,6 +106,7 @@ pk_alpm_config_free (PkAlpmConfig *config)
 	FREELIST (config->noextracts);
 	FREELIST (config->noupgrades);
 	FREELIST (config->remotefilesiglevels);
+	FREELIST (config->hookdirs);
 
 	alpm_list_free_inner (config->sections, pk_alpm_config_section_free);
 	alpm_list_free (config->sections);
@@ -369,6 +370,7 @@ static const PkAlpmConfigList pk_alpm_config_list_options[] = {
 	{ "NoUpgrade", G_STRUCT_OFFSET (PkAlpmConfig, noupgrades) },
 	{ "RemoteFileSigLevel", G_STRUCT_OFFSET (PkAlpmConfig,
 						 remotefilesiglevels) },
+	{ "HookDir", G_STRUCT_OFFSET (PkAlpmConfig, hookdirs) },
 	{ NULL, 0 }
 };
 
@@ -1010,6 +1012,10 @@ pk_alpm_config_configure_alpm (PkBackend *backend, PkAlpmConfig *config, GError 
 	/* alpm takes ownership */
 	alpm_option_set_noupgrades (handle, config->noupgrades);
 	config->noupgrades = NULL;
+
+	/* alpm takes ownership */
+	alpm_option_set_noupgrades (handle, config->hookdirs);
+	config->hookdirs = NULL;
 
 	pk_alpm_config_configure_repos (backend, config, handle, error);
 
--- a/backends/alpm/pk-alpm-config.c
+++ b/backends/alpm/pk-alpm-config.c
@@ -942,6 +942,9 @@ pk_alpm_config_configure_alpm (PkBackend *backend, PkAlpmConfig *config, GError 
 {
 	PkBackendAlpmPrivate *priv = pk_backend_get_user_data (config->backend);
 	alpm_handle_t *handle;
+	gchar **arches;
+	gint i;
+	alpm_list_t *arches_list = NULL;
 
 	g_return_val_if_fail (config != NULL, FALSE);
 
@@ -951,7 +954,14 @@ pk_alpm_config_configure_alpm (PkBackend *backend, PkAlpmConfig *config, GError 
 
 	alpm_option_set_checkspace (handle, config->checkspace);
 	alpm_option_set_usesyslog (handle, config->usesyslog);
-	alpm_option_set_arch (handle, config->arch);
+
+	arches = g_strsplit (config->arch, ",", -1);
+	for (i = 0; arches[i]; i++) {
+		arches_list = alpm_list_add (arches_list, arches[i]);
+	}
+	alpm_option_set_architectures (handle, arches_list);
+	g_strfreev (arches);
+	alpm_list_free (arches_list);
 
 	/* backend takes ownership */
 	g_free (xfercmd);
--- a/backends/alpm/pk-alpm-error.c
+++ b/backends/alpm/pk-alpm-error.c
@@ -72,7 +72,6 @@ pk_alpm_error_emit (PkBackendJob *job, GError *error)
 		break;
 	case ALPM_ERR_DB_OPEN:
 	case ALPM_ERR_DB_NOT_FOUND:
-	case ALPM_ERR_PKG_REPO_NOT_FOUND:
 		code = PK_ERROR_ENUM_REPO_NOT_FOUND;
 		break;
 	case ALPM_ERR_DB_CREATE:
--- a/backends/alpm/pk-alpm-sync.c
+++ b/backends/alpm/pk-alpm-sync.c
@@ -176,7 +176,7 @@ pk_backend_sync_thread (PkBackendJob* job, GVariant* params, gpointer p)
 	PkBackendAlpmPrivate *priv = pk_backend_get_user_data (backend);
 	PkBitfield flags;
 	gboolean only_trusted;
-	const alpm_list_t *i;
+	alpm_list_t *i;
 	alpm_list_t *asdeps = NULL, *asexplicit = NULL;
 	alpm_transflag_t alpm_flags = 0;
 	const gchar** package_ids;
@@ -190,9 +190,7 @@ pk_backend_sync_thread (PkBackendJob* job, GVariant* params, gpointer p)
 
 	if ((gboolean)p) {
 		i = alpm_get_syncdbs(priv->alpm);
-		for (; i != NULL; i = i->next) {
-			pk_alpm_update_database(job, TRUE, i->data, &error);
-		}
+		pk_alpm_refresh_databases (job, TRUE, i, &error);
 	}
 
 	/* download only */
--- a/backends/alpm/pk-alpm-sync.c
+++ b/backends/alpm/pk-alpm-sync.c
@@ -191,7 +191,7 @@ pk_backend_sync_thread (PkBackendJob* job, GVariant* params, gpointer p)
 	if (!only_trusted && !pk_alpm_disable_signatures (backend, &error))
 		goto out;
 
-	if ((gboolean)p) {
+	if ((gblooean)p) {
 		i = alpm_get_syncdbs(priv->alpm);
 		pk_alpm_refresh_databases (job, TRUE, i, &error);
 	}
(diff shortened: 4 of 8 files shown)


eworm-de:
CC @rastermann @pontaoski @antonio-rojas 

Is anybody happy with this?

eworm-de:
Supersedes #407
Fixes #484

rastermann:
Oh thanks. I was out of time last weekend to do this. It'd be nice to see these merged as this series addresses packagekit working on arch - specifically with the new pacman APIs.

antonio-rojas:
There are still some (small) issues: pkcon get-update-detail crashes the daemon

smac89:
Please don't forget #342

meson.build
```diff
@@ -73,7 +73,15 @@ add_project_arguments ('-DPK_COMPILATION', language: 'c')
 conf = configuration_data()
 conf.set_quoted('DATADIR', join_paths(get_option('prefix'), get_option('datadir')))
 conf.set_quoted('LIBDIR', join_paths(get_option('prefix'), get_option('libdir')))
-conf.set_quoted('SYSCONFDIR', get_option('sysconfdir'))
+if get_option('sysconfdir') == 'etc'
+  pk_sysconfdir = join_paths(get_option('prefix'), get_option('sysconfdir'))
```
Conan-Kudo, line 77: What is this?!

Conan-Kudo (changes requested):

meson.build
rastermann: if you compile packagekit form source ... as a normal build without --prefix=/usr ... then it SHOULD install in /usr/local/etc (as prefix is /usr/local and you don't want it going and installing things out-of-prefix that it does not have to).

the problem is it wants to install in "etc" even without the / in front and this then breaks a whole bunch of #define paths (but the meson install actually prepends the / so the files install but then the defines are broken as they contain no leading /). you should be able to install packagekit in a prefix of /usr/local or /opt/whatever for testing/dev/debugging and not mess with the system install of packagkit. in fact the whole sysconfdir handling in packagekit's meson build + defines was rather hairy and it didn't differentiate between looking at existing system files that packagkit is not installing e.g. from pacman and other system packaging systems vs. its own files it installed in sysconfdir. i split these up (sysconfdir vs pk_sysconfdir) so they don't assume it's all the same. as PACKAGE builds will explicitly define sysconfdir as /etc this if clause doesn't trigger as 'etc' without the / in front is meson running as defaults, so this won't affect installs to /usr.

rastermann:
> There are still some (small) issues: pkcon get-update-detail crashes the daemon. Likely the cause of
> https://bugs.kde.org/show_bug.cgi?id=438353
> and
> https://bugs.archlinux.org/task/71196

heisenbug... running under valgrind and it works. :( it turns out valgrind ends up zeroing some stack vars that were uninitialized that seems to have caused this. it looks like this bug was lurking there always as this code did not change. it looks like no one uses asan or valgrind or even coverity on packagekit... :)

try this patch: quick and dirty:

```
--- PackageKit-1.2.3/backends/alpm/pk-alpm-config.c~	2021-06-13 10:51:22.442130102 +0100
+++ PackageKit-1.2.3/backends/alpm/pk-alpm-config.c	2021-06-13 11:06:40.003672764 +0100
@@ -807,7 +807,7 @@
 pk_alpm_config_configure_repos (PkBackend *backend, PkAlpmConfig *config,
 		                           alpm_handle_t *handle, GError **error)
 {
-	alpm_siglevel_t base, level, mask, local, remote;
+	alpm_siglevel_t base, level = 0, mask = 0, local, remote;
 	const alpm_list_t *i;
 	PkAlpmConfigSection *options;

```
this should clear up the get-update-detail issue. 

as for https://bugs.kde.org/show_bug.cgi?id=438353 - some pkcon command .. as i don't seem to see an issue poking around there or in using e's package updater gui with any updates (so far only 1 to mess with as packages haven't updated much since i updated to pacman 6). similarly for https://bugs.kde.org/show_bug.cgi?id=438353 - i'm not using kde (enlightenment also uses packagekit thus one reason i have dug in to updating it for pacman 6). e seems all happy as a pig in mud right now but keep in mind it goes directly to the dbus interface itself and doesn't use libpackagekit.

as for #342 i can see why this doesn't get found - at least my pacman.conf has no such option in it... but the alpm backend in packagekit doesn't handle a lot of directives. this is the list of them it handles:

Architecture CacheDir CleanMethod DBPath GPGDir LogFile RootDir XferCommand HoldPkg IgnoreGroup IgnorePkg LocalFileSigLevel NoExtract NoUpgrade RemoteFileSigLevel CheckSpace Color DisableDownloadTimeout ILoveCandy NoProgressBar TotalDownload UseSyslog VerbosePkgLists SigLevel Usage ParallelDownloads Server

That means the only directive it doesn't handle is indeed HookDir - the rest are handled or ignored. The following fixes that:

```
--- PackageKit-1.2.3/backends/alpm/pk-alpm-config.c.orig	2021-06-13 11:55:35.398817736 +0100
+++ PackageKit-1.2.3/backends/alpm/pk-alpm-config.c	2021-06-13 11:55:59.478660912 +0100
@@ -47,7 +47,7 @@

 	 alpm_list_t	        *cachedirs, *holdpkgs, *ignoregroups,
 		                        *ignorepkgs, *localfilesiglevels, *noextracts,
-		                        *noupgrades, *remotefilesiglevels;
+		                        *noupgrades, *remotefilesiglevels, *hookdirs;

 	 alpm_list_t	        *sections;
 	 GRegex	        *xrepo, *xarch;
@@ -106,6 +106,7 @@
 	FREELIST (config->noextracts);
 	FREELIST (config->noupgrades);
 	FREELIST (config->remotefilesiglevels);
+	FREELIST (config->hookdirs);

 	alpm_list_free_inner (config->sections, pk_alpm_config_section_free);
 	alpm_list_free (config->sections);
@@ -369,6 +370,7 @@
 	{ "NoUpgrade", G_STRUCT_OFFSET (PkAlpmConfig, noupgrades) },
 	{ "RemoteFileSigLevel", G_STRUCT_OFFSET (PkAlpmConfig,
 			                                         remotefilesiglevels) },
+	{ "HookDir", G_STRUCT_OFFSET (PkAlpmConfig, hookdirs) },
 	{ NULL, 0 }
 };

@@ -1011,6 +1013,10 @@
 	alpm_option_set_noupgrades (handle, config->noupgrades);
 	config->noupgrades = NULL;

+	/* alpm takes ownership */
+	alpm_option_set_noupgrades (handle, config->hookdirs);
+	config->hookdirs = NULL;
+
 	pk_alpm_config_configure_repos (backend, config, handle, error);

 	return handle;
```

:) I'll throw these above fixes into trunk packagekit in arch for testing.

antonio-rojas:
> heisenbug... running under valgrind and it works. :( it turns out valgrind ends up zeroing some stack vars that were uninitialized that seems to have caused this. it looks like this bug was lurking there always as this code did not change. it looks like no one uses asan or valgrind or even coverity on packagekit... :)
> 
> try this patch: quick and dirty:
> 
> this should clear up the get-update-detail issue.

Doesn't work for me. The "upgrades" parameter seems to be deleted some time during the pk_backend_job_update_detail call.

```
(gdb) bt
#0  0x00007fd5d15ea547 in __strlen_avx2 () at /usr/lib/libc.so.6
#1  0x00007fd5d1bbff04 in g_strdup (str=0x60c385b0 <error: Cannot access memory at address 0x60c385b0>) at ../glib/glib/gstrfuncs.c:363
#2  0x00007fd5d1bc23aa in g_strdupv (str_array=0x7fd5ca7b8cb8) at ../glib/glib/gstrfuncs.c:2587
#3  0x00007fd5d1b0d5a4 in _g_type_boxed_copy (type=0x556968bfc6b0 [None], value=0x7fd5ca7b8cb8) at ../glib/gobject/gtype.c:4355
#4  0x00007fd5d1adf957 in boxed_proxy_collect_value (value=0x7fd5ca7b8940, n_collect_values=1, collect_values=0x7fd5ca7b8960, collect_flags=0)
    at ../glib/gobject/gboxed.c:249
#5  0x00007fd5d1aedf94 in g_object_set_valist (object=0x556968bea6c0, first_property_name=0x55696755c46d "package-id", var_args=0x7fd5ca7b89e0)
    at ../glib/gobject/gobject.c:2454
#6  0x00007fd5d1aeea27 in g_object_set (_object=0x556968bea6c0, first_property_name=0x55696755c46d "package-id") at ../glib/gobject/gobject.c:2635
#7  0x0000556967546093 in pk_backend_job_update_detail
    (job=0x556968cc0770, package_id=0x556968d6ae60 "nss;3.67-1;x86_64;core", updates=0x7fd5ca7b8cb8, obsoletes=0x7fd5ca7b8cb0, vendor_urls=0x7fd5b88487d0, bugzilla_urls=0x0, cve_urls=0x0, restart=PK_RESTART_ENUM_NONE, update_text=0x7fd5d0190e48 "Update to a new upstream version", changelog=0x0, state=PK_UPDATE_STATE_ENUM_STABLE, issued_text=0x7fd5b8e0fee0 "2021-06-11T15:48:00Z", updated_text=0x7fd5b858a010 "2021-06-13T11:18:26Z")
    at ../PackageKit/src/pk-backend-job.c:1079
#8  0x00007fd5d018ebf4 in pk_backend_get_update_detail_thread (job=0x556968cc0770, params=0x556968c15660, p=0x556968cf43b0)
    at ../PackageKit/backends/alpm/pk-alpm-update.c:191
#9  0x0000556967544fc4 in pk_backend_job_thread_setup (thread_data=0x556968cf15e0) at ../PackageKit/src/pk-backend-job.c:724
#10 0x00007fd5d1bcdbbd in g_thread_proxy (data=0x7fd5c011a000) at ../glib/glib/gthread.c:826
#11 0x00007fd5d1c0495f in linux_pthread_proxy (data=0x7fd5c011a000) at ../glib/glib/gthread-posix.c:1269
#12 0x00007fd5d165f259 in start_thread () at /usr/lib/libpthread.so.0
#13 0x00007fd5d15885e3 in clone () at /usr/lib/libc.so.6
```

rastermann:
Let me roll that back... pkcon get-update-detail still crashes. my patches do fix an issue .. but not what they claim to... my compiling with -g -O0 happened to fix it in combo with my uninit stack var fix. when i build with optimizations now... it still crashes. specifically:

```
(gdb) bt
#0  0x00007ffff7961515 in __strlen_avx2 () at /usr/lib/libc.so.6
#1  0x00007ffff7e990c5 in g_strdup () at /usr/lib/libglib-2.0.so.0
#2  0x00007ffff7e9d3de in g_strdupv () at /usr/lib/libglib-2.0.so.0
#3  0x00007ffff7ddec59 in  () at /usr/lib/libgobject-2.0.so.0
#4  0x00007ffff7df129c in g_object_set_valist () at /usr/lib/libgobject-2.0.so.0
#5  0x00007ffff7df1585 in g_object_set () at /usr/lib/libgobject-2.0.so.0
#6  0x0000555555562862 in pk_backend_job_update_detail ()
#7  0x00007ffff548a926 in  () at /usr/lib/packagekit-backend/libpk_backend_alpm.so
#8  0x000055555556473a in  ()
#9  0x00007ffff7eaa0c1 in  () at /usr/lib/libglib-2.0.so.0
#10 0x00007ffff79d6259 in start_thread () at /usr/lib/libpthread.so.0
#11 0x00007ffff78ff5e3 in clone () at /usr/lib/libc.so.6
```

when i was inspecting was was passed in all seemed fine with gdb debug on. hmm. danged heisenbugs.

rastermann:
ok. fixed get-update-detail with:

```
diff --git a/backends/alpm/pk-alpm-update.c b/backends/alpm/pk-alpm-update.c
index aa641c02b..4071a10f6 100644
--- a/backends/alpm/pk-alpm-update.c
+++ b/backends/alpm/pk-alpm-update.c
@@ -39,33 +39,32 @@
 #include "pk-alpm-transaction.h"
 #include "pk-alpm-update.h"

-static gchar *
+static gchar **
 pk_alpm_pkg_build_replaces (PkBackendJob *job, alpm_pkg_t *pkg)
 {
 	PkBackend *backend = pk_backend_job_get_backend (job);
 	PkBackendAlpmPrivate *priv = pk_backend_get_user_data (backend);
 	const alpm_list_t *i;
-	GString *string = NULL;
+	gchar **replaces = NULL;
+	gint count = 0;

 	g_return_val_if_fail (pkg != NULL, NULL);

 	/* make a list of the packages that package replaces */
 	for (i = alpm_pkg_get_replaces (pkg); i != NULL; i = i->next) {
-	        alpm_pkg_t *replaces = alpm_db_get_pkg (priv->localdb, i->data);
-
-	        if (replaces != NULL) {
-		                g_autofree gchar *package = pk_alpm_pkg_build_id (replaces);
-		                if (string == NULL) {
-		                        string = g_string_new (package);
-		                } else {
-		                        g_string_append_printf (string, "&%s", package);
+	        alpm_pkg_t *package = alpm_db_get_pkg (priv->localdb, i->data);
+
+	        if (package != NULL) {
+		                gchar *id = pk_alpm_pkg_build_id (package);
+		                if (id) {
+		                        replaces =  g_realloc (replaces, ((++count) + 1) * sizeof(gchar *));
+		                        replaces[count - 1] = id;
+		                        replaces[count] = NULL;
 		                }
 	        }
 	}

-	if (string == NULL)
-	        return NULL;
-	return g_string_free (string, FALSE);
+	return replaces;
 }

 static gchar **
@@ -143,8 +142,8 @@ pk_backend_get_update_detail_thread (PkBackendJob *job, GVariant* params, gpoint
 	        PkRestartEnum restart = PK_RESTART_ENUM_NONE;
 	        PkUpdateStateEnum state = PK_UPDATE_STATE_ENUM_STABLE;
 	        alpm_time_t built, installed;
-	        g_autofree gchar *upgrades = NULL;
-	        g_autofree gchar *replaces = NULL;
+	        g_autofree gchar *upgrades[2] = { NULL, NULL };
+	        g_autofree gchar **replaces;
 	        g_auto(GStrv) urls = NULL;
 	        g_autofree gchar *issued = NULL;
 	        g_autofree gchar *updated = NULL;
@@ -158,7 +157,7 @@ pk_backend_get_update_detail_thread (PkBackendJob *job, GVariant* params, gpoint

 	        old = alpm_db_get_pkg (priv->localdb, alpm_pkg_get_name (pkg));
 	        if (old != NULL) {
-		                upgrades = pk_alpm_pkg_build_id (old);
+		                upgrades[0] = pk_alpm_pkg_build_id (old);
 		                if (pk_alpm_pkg_same_pkgver (pkg, old)) {
 		                        reason = "Update to a newer release";
 		                } else {
@@ -188,8 +187,8 @@ pk_backend_get_update_detail_thread (PkBackendJob *job, GVariant* params, gpoint
 		                        updated = pk_alpm_time_to_iso8601 (installed);
 	        }

-	        pk_backend_job_update_detail (job, *packages, &upgrades,
-			                                      &replaces, urls, NULL, NULL,
+	        pk_backend_job_update_detail (job, *packages, upgrades,
+			                                      replaces, urls, NULL, NULL,
 			                                      restart, reason, NULL, state,
 			                                      issued, updated);
 	}
```

putting these in arch trunk for pkgbuild for testing now. work for me. haven't addressed the "kde bugs" above as i can't see them here. right now.

antonio-rojas:
> ok. fixed get-update-detail with:

> putting these in arch trunk for pkgbuild for testing now. work for me. haven't addressed the "kde bugs" above as i can't see them here. right now.

Thanks! I confirm this fixes pkcon get-update-details. But now I'm seeing some double-free crashes when running Discover
```
#0  0x00007ff1e00e99a9 in free () at /usr/lib/libc.so.6
#1  0x00007ff1e0773c06 in g_free (mem=0x2de56042c) at ../glib/glib/gmem.c:199
#2  0x00007ff1de56005e in g_autoptr_cleanup_generic_gfree (p=0x7ff1dd399ca0) at /usr/include/glib-2.0/glib/glib-autocleanups.h:28
#3  0x00007ff1de560768 in pk_backend_get_update_detail_thread (job=0x562d9a92ad50, params=0x7ff1d0004c60, p=0x562d9a942d90)
    at ../PackageKit-1.2.3/backends/alpm/pk-alpm-update.c:146
#4  0x0000562d99bbf0fd in pk_backend_job_thread_setup (thread_data=0x562d9a9ad750) at ../PackageKit-1.2.3/src/pk-backend-job.c:724
#5  0x00007ff1e07a1bbd in g_thread_proxy (data=0x562d9a8fa5e0) at ../glib/glib/gthread.c:826
#6  0x00007ff1e07d895f in linux_pthread_proxy (data=0x562d9a8fa5e0) at ../glib/glib/gthread-posix.c:1269
#7  0x00007ff1e0233259 in start_thread () at /usr/lib/libpthread.so.0
#8  0x00007ff1e015c5e3 in clone () at /usr/lib/libc.so.6
```

rastermann:
try this patch on top:

```
--- PackageKit-1.2.3/backends/alpm/pk-alpm-update.c~	2021-06-13 13:04:07.774556838 +0100
+++ PackageKit-1.2.3/backends/alpm/pk-alpm-update.c	2021-06-13 19:33:22.943089089 +0100
@@ -142,8 +142,9 @@
 	        PkRestartEnum restart = PK_RESTART_ENUM_NONE;
 	        PkUpdateStateEnum state = PK_UPDATE_STATE_ENUM_STABLE;
 	        alpm_time_t built, installed;
-	        g_autofree gchar *upgrades[2] = { NULL, NULL };
-	        g_autofree gchar **replaces;
+	        gchar *upgrades[2] = { NULL, NULL };
+	        gchar **replaces;
+	        gchar **charptr;
 	        g_auto(GStrv) urls = NULL;
 	        g_autofree gchar *issued = NULL;
 	        g_autofree gchar *updated = NULL;
@@ -191,6 +192,12 @@
 			                                      replaces, urls, NULL, NULL,
 			                                      restart, reason, NULL, state,
 			                                      issued, updated);
+	        if (updgrades[0]) g_free (upgrades[0]);
+	        if (replaces) {
+		                for (charptr = replaces; charptr[0]; charptr++)
+		                        g_free (charptr[0]);
+		                g_free(replaces);
+	        }
 	}

 	pk_alpm_finish (job, error);
```

antonio-rojas:
Seems to work, thanks (unfortunately doesn't fix the Discover issues though, but that might not even be a PK issue)

antonio-rojas:
> https://bugs.kde.org/show_bug.cgi?id=438353
> and
> https://bugs.archlinux.org/task/71196

seem to be fallout from the database duplication in a76711fd450a326da97bd90e085734c20a852e00
PK queries the pacman db for details of the available updates, but these updates are still not present in the real pacman db, only in the PK specific copy.

antonio-rojas:
> > https://bugs.kde.org/show_bug.cgi?id=438353
> > and
> > https://bugs.archlinux.org/task/71196
> 
> seem to be fallout from the database duplication in [a76711f](https://github.com/PackageKit/PackageKit/commit/a76711fd450a326da97bd90e085734c20a852e00)
> PK queries the pacman db for details of the available updates, but these updates are still not present in the real pacman db, only in the PK specific copy.

I opened https://github.com/PackageKit/PackageKit/issues/491 to track these issues. Since they have nothing to do with pacman 6, they shouldn't hold this back.

eworm-de:
Updated with the latest changes from our svn repository.
This is now up to date and ready to merge?

eworm-de:
Nobody complained, so I guess I got this right.

@hughsie, want to have a look at 21464a3a776583bada68052dd01fe9d03d58d13d? That touches non-alpm files for `sysconfdir` handling.

rastermann:
@eworm-de my commit log explains it ... part of the fixing up was me finding the mixup of sysconfdir usage :)

meson.build
Conan-Kudo: Oh dear, well, I guess this fix makes sense in light of that. Is there a Meson bug filed about this?

Conan-Kudo:
@eworm-de @rastermann Now that I fully understand that change, I feel like this is a Meson bug. ☹️

hughsie:
>  I feel like this is a Meson bug

/me also. I'd much rather fix meson to do the sane thing rather than some huge workaround.

rastermann:
you can wait for meson to fix it - but the fix there in pkgkit will not be affected and can be removed eventually when the fix filters though meson. the rest of the changes are still also valid as there is different between eh "host system etc dir" and pkgkit own etc dir - they may be in different places. you have to differentiate between the 2. :) one is provided buy the host, one is provided by pkgkit.

aleixpol:
Let's merge this? archlinux is just packaging it in, so I guess we better have the same version upstream. Also main as is doesn't build. This is mostly what we've been testing when using packagekit on arch since the MR got submitted.

aleixpol (approved):

aleixpol:
FWIW, Dropping the meson changes would also make sense to me, but we shouldn't just leave alpm as is.

Conan-Kudo:
If you drop the Meson changes, I'll happily merge it.

aleixpol (changes requested):
Then if @eworm-de isn't faster, I'll try to put together a PR without the meson bit.

eworm-de:
Pushed without meson changes.

aleixpol (approved):
Thanks!

eworm-de:
The meson change is here: https://github.com/eworm-de/PackageKit/tree/pacman-6-with-meson

eworm-de:
... and rebased on main.

Conan-Kudo (approved):

Conan-Kudo:
Merged, thanks!

rastermann:
Just got back home to this. Is there something outstanding I need to get to? The meson changes still stand as I said already - meson has its brokenness and they were there to fix it to respect prefix AND to split up sysconf dir between what is expected on the host side form pacman in sysconfdir vs what is installed and used by pkgkit which might not be in the same sysconfdir. packages that explicitly set sysconfdir to /etc are working around this bug and anyone not explicitly setting it will find problems with the build.

Conan-Kudo:
Please work with the Meson project to get it fixed there.

rastermann:
that can't fix the fact that pkgkit's build does not differentiate between the sysconfdir for pkgkit itself (depending where it installed - for example if you don't provide an explicit sysconfdir then it may install its own files in /usr/local/etc but the HOST sysconfdir is different - eg /etc where pacman config might be that packagegkit has to work with). of my patches a tiny 2 lines are fixing etc to /etc ... the rest fixes the above. ONCE meson has been fixed the etc -> /etc can be removed - but the above fixes for sysconfdir differentiation of pkgkit's own stuff vs host stuff is still needed and it depends on a workaround meson. it's up to you. if you don't want this problem fixed... then no point worrying about those patches. :)

Issue #489, opened 2021-06-16
PkClient thread safety on (sync) call cancel


mcrha:
As had been said on the https://github.com/PackageKit/PackageKit/pull/471 , those changes partly reverted things from #464, which is not good.

What I found (from what I see when gnome-software calls the PkClient functions):
* the 'state' structure itself is in two stages, the first is an active stage, when it does one or more D-Bus calls in a row
* the second stage is the one when it is waiting for the "Finished" D-Bus signal
* the stage itself matters with respect of the name-owner-changed; freeing the 'state' structure is safe when its waiting on the Finished signal, but it cannot be done when not, because there will be delivered a callback of the D-Bus call

The sync calls:
* the synchronous call pushes its own thread default main context and iterates it until the call is done (gnome-software uses a dedicated thread for the synchronous call [1])
* all the calls are "synchronized" through this new main context, because the D-Bus call results, D-Bus signals and such is delivered there
* the passed-in cancellable can be cancelled from any thread - definitely a different thread than the one the main context is working on
* the GDBus call from the pk_client_cancellable_cancel_cb() is scheduled on the other main context, not the one above - here comes one part of the fun, because the pk_client_cancel_cb() can be delivered in that other thread, which can be after the pk_client_state_finish() call, thus on a freed memory.

The first problem is easy, just have a variable for the stage and act according to its value.

The second problem is *the problem*. It requires thread synchronization, but I do not see a good solution for it with a bare structure. Being the PkClientState a GObject, one can use GWeakRef, but with the bare structure it's not that easy. Imagine there's added a mutex into pk_client_cancellable_cancel_cb() and pk_client_state_finish(). Nobody can guarantee that the pk_client_cancellable_cancel_cb() will get its lock before the pk_client_state_finish(). I can imagine a situation where the pk_client_state_finish() acquires the lock and a tick after the other thread invokes pk_client_cancellable_cancel_cb(), which will wait for the lock, but that will be released and freed slightly later, when the pk_client_state_finish() receives back the execution.

What do you think? Would it be acceptable to turn the PkClientState into the GObject and use the GWeakRef where appropriate?

[...]

PR #490, opened 2021-06-17
PkClient: Improve thread safety on an operation cancellation

mcrha:
The PkClient operates in two stages. The first is an "active" stage,
which makes one or more D-Bus calls. The second is a "passive" stage,
when it is waiting for the "Finished" D-Bus signal from the daemon.
The cancellation can come in any time. This change keeps the cancellation
up to the D-Bus calls during the first stage and actively stops the waiting
during the second stage.

As the GCancellable is cancelled in a different thread, and any D-Bus call
is also done and its result delivered in that thread default main context,
rather than in the main context when the PkClientState had been created,
the PkClientState is switched to a GObject descendant, to be able to benefit
from the GWeakRef, which can avoid use-after-free easily.

The GDBusProxy callbacks can be schedules in the middle of the object free,
thus the added callbacks also benefit from the GWeakRef usage.

All of this helps to better cover the PackageKit daemon disappear during
an ongoing call to it.

Closes https://github.com/PackageKit/PackageKit/issues/489

### DIFF
--- a/lib/packagekit-glib2/pk-client.c
+++ b/lib/packagekit-glib2/pk-client.c
@@ -80,7 +80,59 @@ enum {
 
 G_DEFINE_TYPE (PkClient, pk_client, G_TYPE_OBJECT)
 
-typedef struct {
+#define PK_TYPE_CLIENT_STATE (pk_client_state_get_type ())
+
+G_DECLARE_FINAL_TYPE (PkClientState, pk_client_state, PK, CLIENT_STATE, GObject)
+
+static GWeakRef *
+pk_client_weak_ref_new (gpointer object)
+{
+	GWeakRef *weak_ref;
+
+	weak_ref = g_slice_new0 (GWeakRef);
+	g_weak_ref_init (weak_ref, object);
+
+	return weak_ref;
+}
+
+static void
+pk_client_weak_ref_free (gpointer ptr)
+{
+	GWeakRef *weak_ref = ptr;
+
+	g_return_if_fail (weak_ref != NULL);
+
+	g_weak_ref_clear (weak_ref);
+	g_slice_free (GWeakRef, weak_ref);
+}
+
+static void
+pk_client_weak_ref_free_gclosure (gpointer ptr,
+				  GClosure *closure)
+{
+	pk_client_weak_ref_free (ptr);
+}
+
+static void
+pk_client_properties_changed_cb (GDBusProxy *proxy,
+				 GVariant *changed_properties,
+				 const gchar* const  *invalidated_properties,
+				 gpointer user_data);
+static void
+pk_client_signal_cb (GDBusProxy *proxy,
+		     const gchar *sender_name,
+		     const gchar *signal_name,
+		     GVariant *parameters,
+		     gpointer user_data);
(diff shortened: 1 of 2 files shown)


hughsie:
Is it possible to add self tests for this? I think you understand the problem better than me now, tbh.

mcrha:
It depends. How do I achieve proper timing, to issue the g_cancellable_cancel() at the right time? I've been able to do it by cheating in the code, by adding there artificial delays, but such things cannot be in the production code.

Conan-Kudo:
I'm not sure myself how we'd even put this in the unit tests...

hughsie:
@mcrha do the tests work for you? I tried:

    sudo ./src/packagekitd --verbose --backend=dummy &
    ./lib/packagekit-glib2/pk-test-daemon

and got:

    PackageKit:ERROR:../lib/packagekit-glib2/pk-test-daemon.c:404:pk_test_client_search_name_cb: assertion failed (error == NULL): Operation was cancelled (g-io-error-quark, 19)
    Bail out! PackageKit:ERROR:../lib/packagekit-glib2/pk-test-daemon.c:404:pk_test_client_search_name_cb: assertion failed (error == NULL): Operation was cancelled (g-io-error-quark, 19)

mcrha:
It aborts, but on a lower line:
```
PackageKit:ERROR:../lib/packagekit-glib2/pk-test-daemon.c:795:pk_test_control_get_properties_cb: assertion failed (text == "application/x-rpm;application/x-deb"): ("application/x-rpm" == "application/x-rpm;application/x-deb")
Bail out! PackageKit:ERROR:../lib/packagekit-glib2/pk-test-daemon.c:795:pk_test_control_get_properties_cb: assertion failed (text == "application/x-rpm;application/x-deb"): ("application/x-rpm" == "application/x-rpm;application/x-deb")

```

mcrha:
I might run the tests incorrectly, I'm sorry. I see the `/packagekit-glib2/client` failing at the same line with the same error message too.

mcrha:
I see the `pk_test_client_cancel_cb()` had been called there, before the `pk_test_client_search_name_cb` had been called, thus it's correct to receive "Operation cancelled", no?

mcrha:
Ah, I see what the tests want to do. I updated the patch (reverted one change, which caused it).

hughsie:
@mcrha can we get those last two commits in Fedora rawhide please? If they don't cause any problems in Fedora then I think we do an upstream tarball and include the fix in Fedora 34 too.

mcrha:
I do not have commit rights for PackageKit and I believe it's for good.

It would be good to use also https://github.com/rpm-software-management/libdnf/pull/1252 in the PacakgeKit before the release, because it causes a significant performance improvement. I did not open a merge request for it yet.

mcrha:
> It would be good to use also [rpm-software-management/libdnf#1252](https://github.com/rpm-software-management/libdnf/pull/1252) in the PacakgeKit before the release, because it causes a significant performance improvement. I did not open a merge request for it yet.

I opened https://github.com/PackageKit/PackageKit/pull/493 for that.

PR #492, opened 2021-06-22
Allow admin users to remove packages without password prompt

mcatanzaro:
This replaces #404.

A local, active admin user can install packages without a password
prompt, but has to enter the admin password to remove packages. This
doesn't make much sense. It should be parallel.

Note that this change has no effect on what users are able to do,
because it only applies to admin users. The password only protects
against unlocked workstation attackers, where an attacker gains physical
access to an unlocked desktop. It's pretty weird to prevent such an
attacker from removing software, but allow installing new stuff.

This makes users more vulnerable to secret agents who access your laptop
when you leave it unlocked while ordering a coffee. This is just not a
very practical concern for most users. I'd be more upset about somebody
viewing my files if I leave my laptop unlocked, not so upset about
somebody uninstalling Calendar or whatnot.

https://pagure.io/fedora-workstation/issue/233

### DIFF
--- a/policy/org.freedesktop.packagekit.rules
+++ b/policy/org.freedesktop.packagekit.rules
@@ -1,5 +1,6 @@
 polkit.addRule(function(action, subject) {
-    if (action.id == "org.freedesktop.packagekit.package-install" &&
+    if ((action.id == "org.freedesktop.packagekit.package-install" ||
+         action.id == "org.freedesktop.packagekit.package-remove") &&
         subject.active == true && subject.local == true &&
         subject.isInGroup("wheel")) {
             return polkit.Result.YES;


hughsie:
I thought you were going to do this downstream?

Conan-Kudo:
I suggested doing it upstream, since I'm not convinced there's a good reason to maintain it as a downstream-only change.

hughsie:
> I'm not convinced there's a good reason to maintain it as a downstream-only change

Does _the upstream maintainer doesn't think it's a good idea_ count?

Conan-Kudo:
> > I'm not convinced there's a good reason to maintain it as a downstream-only change
> 
> Does _the upstream maintainer doesn't think it's a good idea_ count?

You did not say that: https://github.com/PackageKit/PackageKit/pull/404#issuecomment-639455335

Quote:
> I agree with you dude, but this isn't a battle I really want to fight right now.

If there's a "battle" to be fought for it, let's have the battle, then. @mcatanzaro and I can have this conversation with the relevant Red Hat security folks. 

As for SUSE, they _already_ don't ship our polkit policy and use their own one that makes PackageKit even more unusable by default.

hughsie:
> I agree with you dude, but this isn't a battle I really want to fight right now.

Keep reading all the other comments in that discussion.

mcatanzaro:
> If there's a "battle" to be fought for it, let's have the battle, then. @mcatanzaro and I can have this conversation with the relevant Red Hat security folks. Point them our way, let's figure this out.

FWIW, yes I will help with any downstream issues, I don't expect any such battle except maybe occasional questions fromr users. (1) the change is already approved by the appropriate Fedora governance body (Workstation WG, since PackageKit is a desktop technology). (2) Product Security is full of smart people who understand threat modeling. The threat here is [sufficiently farfetched](https://github.com/PackageKit/PackageKit/pull/404#issuecomment-866090148) that it would be weird for anybody to be too concerned.

hughsie:
> that it would be weird for anybody to be too concerned

Can we stick to being nice to each other please.

mcatanzaro:
Apologies, that was rude of me.

Conan-Kudo:
Closing this because @mcatanzaro is looking to do this downstream.

PR #493, opened 2021-06-29
dnf: Use 'hy_query_get_advisory_pkgs', if available

mcrha:
This provides significant performance improvement when checking
package advisories.

Do not bump version dependency on libdnf, rather check during
the configure phase whether the function exists and conditional-compile
related code accordingly.

This depends on https://github.com/rpm-software-management/libdnf/pull/1252

### DIFF
--- a/backends/dnf/meson.build
+++ b/backends/dnf/meson.build
@@ -1,6 +1,11 @@
 appstream_dep = dependency('appstream-glib')
 dnf_dep = dependency('libdnf', version: '>=0.43.1')
 rpm_dep = dependency('rpm')
+c_args = ['-DG_LOG_DOMAIN="PackageKit-DNF"']
+
+if meson.get_compiler('c').has_function('hy_query_get_advisory_pkgs', prefix: '#include <libdnf/hy-query.h>', dependencies: dnf_dep)
+   c_args += ['-DHAVE_HY_QUERY_GET_ADVISORY_PKGS']
+endif
 
 shared_module(
   'pk_backend_dnf',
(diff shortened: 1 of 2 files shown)


backends/dnf/meson.build
```diff
@@ -1,6 +1,11 @@
 appstream_dep = dependency('appstream-glib')
 dnf_dep = dependency('libdnf', version: '>=0.43.1')
 rpm_dep = dependency('rpm')
+c_args = ['-DG_LOG_DOMAIN="PackageKit-DNF"']
+
+if meson.get_compiler('c').has_function('hy_query_get_advisory_pkgs', prefix: '#include <libdnf/hy-query.h>', dependencies: dnf_dep)
+   c_args += ['-DHAVE_HY_QUERY_GET_ADVISORY_PKGS=1']
```
hughsie, line 7: I don;t think you need the `=1`

backends/dnf/pk-backend-dnf.c
```diff
@@ -3787,7 +3840,9 @@ pk_backend_get_update_detail_thread (PkBackendJob *job, GVariant *params, gpoint
 					      NULL /* updated */);
 
 		g_ptr_array_unref (references);
+		#ifndef HAVE_HY_QUERY_GET_ADVISORY_PKGS
```
hughsie, line 3843: it's more usual to have the #ifdef without an indent

backends/dnf/meson.build
mcrha: right, it's not needed, but it also doesn't hurt. I can change it.

backends/dnf/pk-backend-dnf.c
mcrha: done

hughsie (approved):

PR #495, opened 2021-07-04
dnf: Implement DependsOn and RequiredBy

Conan-Kudo:
This allows software centers to make decisions based on the
dependency tree.

### DIFF
--- a/backends/dnf/pk-backend-dnf.c
+++ b/backends/dnf/pk-backend-dnf.c
@@ -1911,6 +1911,100 @@ out:
 }
 
 static void
+backend_dependency_query_thread (PkBackendJob *job, GVariant *params, gpointer user_data)
+{
+	gboolean ret;
+	guint i;
+	DnfState *state_local;
+	DnfPackage *pkg;
+	PkBackendDnfJobData *job_data = pk_backend_job_get_user_data (job);
+	PkBitfield filters;
+	PkRoleEnum role;
+	g_autofree gchar **package_ids = NULL;
+	g_autoptr(DnfSack) sack = NULL;
+	g_autoptr(GError) error = NULL;
+	g_autoptr(GHashTable) hash = NULL;
+	gboolean recursive;
+	g_variant_get (params, "(t^a&sb)",
+		       &filters,
+		       &package_ids,
+		       &recursive);
+
+	/* set state */
+	ret = dnf_state_set_steps (job_data->state, NULL,
+				   50, /* add repos */
+				   49, /* find packages */
+				   1, /* emit */
+				   -1);
+	g_assert (ret);
+
+	/* get role */
+	role = pk_backend_job_get_role (job);
+
+	/* get sack */
+	state_local = dnf_state_get_child (job_data->state);
+	sack = dnf_utils_create_sack_for_filters (job,
+						  filters,
+						  DNF_CREATE_SACK_FLAG_USE_CACHE,
+						  state_local,
+						  &error);
+	if (sack == NULL) {
+		pk_backend_job_error_code (job, error->code, "%s", error->message);
+		return;
+	}
+
+	/* done */
+	if (!dnf_state_done (job_data->state, &error)) {
+		pk_backend_job_error_code (job, error->code, "%s", error->message);
+		return;
+	}
+
+	/* get dependency information */
+	if (role == PK_ROLE_ENUM_DEPENDS_ON) {
+		// FIXME: dependson (map to repoquery --depends)
+	} else if (role == PK_ROLE_ENUM_REQUIRED_BY) {
+		// FIXME: requiredby (map to repoquery --whatdepends)
+	}
+
+	/* done */
+	if (!dnf_state_done (job_data->state, &error)) {
+		pk_backend_job_error_code (job, error->code, "%s", error->message);
+		return;
+	}
+
+}
+
+void
+pk_backend_depends_on (PkBackend *backend, PkBackendJob *job, PkBitfield filters, gchar **package_ids, gboolean recursive)
+{
+	PkBackendDnfPrivate *priv = pk_backend_get_user_data (backend);
+	g_autoptr(GError) error = NULL;
+
+	if (!pk_backend_ensure_default_dnf_context (backend, &error)) {
+		pk_backend_job_error_code (job, error->code, "%s", error->message);
+		pk_backend_job_finished (job);
+		return;
+	}
+	pk_backend_job_set_context (job, priv->context);
+	pk_backend_job_thread_create (job, backend_dependency_query_thread, NULL, NULL);
+}
+
+void
+pk_backend_required_by (PkBackend *backend, PkBackendJob *job, PkBitfield filters, gchar **package_ids, gboolean recursive)
+{
+	PkBackendDnfPrivate *priv = pk_backend_get_user_data (backend);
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
This PR currently doesn't actually _have_ the logic implemented, mostly because I'm not sure _how_ to do it yet. However, I figured creating this PR would make it easier to solicit some guidance on what I should be doing, or even if my starting approach is even right.

@dmach, @j-mracek, @jrohel, @hughsie: I would appreciate any help given here...

hughsie:
> make decisions based on the dependency tree

What's the 40,000ft use case? You have to be careful, and this is usually why SIMULATE gives you better results for _what's going to happen_.

Conan-Kudo:
The main use-case for implementing this is that Plasma Discover 5.22 now uses this for informing the user of what will change on installation or removal of an application.

hughsie:
> for informing the user of what will change on installation or removal of an application

You need to use SIMULATE for this. There's no way of showing the correct results without knowing what you're going to do, i.e. what the rest of the transaction looks like.

Conan-Kudo:
But that wouldn't also show what it depends on that's already installed, would it?

hughsie:
No, it _should_ show you the INSTALLING/REMOVING/UPDATING packages just like a normal "install" would do.

Conan-Kudo:
Maybe I'm not explaining this correctly, so perhaps a picture shows this better:

screenshot_20210409_080123

This is from Plasma Discover on openSUSE with the Zypp backend. This currently doesn't work on Fedora or openSUSE with the DNF backend.

hughsie:
Ahh yes; for this kind of thing it's fine -- my point was based on the assumption you wanted "if I installed this what else would have to be installed"

Conan-Kudo:
I've been up for too many hours, I'm bad at words now. 😛 

💤

hughsie:
Okay, lets fix up the FIXMEs and then LGTM.

Conan-Kudo:
Yay, this is finally done and verified it works!

Screenshot_20251207_124915

Conan-Kudo:
Wait, hold on, it thinks bash isn't installed... What?

backends/dnf/pk-backend-dnf.c
```diff
@@ -1841,6 +1841,197 @@ dnf_utils_find_package_ids (DnfSack *sack, gchar **package_ids, GError **error)
...
+			reldeps = dnf_package_get_requires (curr);
+		else
+			reldeps = dnf_package_get_provides (curr);
+
+		n = dnf_reldep_list_count (reldeps);
+		for (guint i = 0; i < n; i++) {
+			DnfReldep *reldep = dnf_reldep_list_index (reldeps, i);
+			const gchar *req = dnf_reldep_to_string (reldep);
+			hy_autoquery HyQuery query = hy_query_create (sack);
+			GPtrArray *results;
```
ximion:
```suggestion
			g_autoptr(GPtrArray) results = NULL;
```

backends/dnf/pk-backend-dnf.c
```diff
@@ -1841,6 +1841,197 @@ dnf_utils_find_package_ids (DnfSack *sack, gchar **package_ids, GError **error)
...
+					continue;
+
+				if (!g_hash_table_contains (visited, resid)) {
+					g_hash_table_add (visited, g_strdup (resid));
+					g_ptr_array_add (results_array, g_object_ref (res));
+					if (recursive)
+						g_queue_push_tail (queue, g_object_ref (res));
+				}
+			}
+			g_ptr_array_unref (results);
```
ximion:
```suggestion

```

backends/dnf/pk-backend-dnf.c
```diff
@@ -1841,6 +1841,197 @@ dnf_utils_find_package_ids (DnfSack *sack, gchar **package_ids, GError **error)
...
+	}
+
+	/* done */
+	if (!dnf_state_done (job_data->state, &error)) {
+		pk_backend_job_error_code (job, error->code, "%s", error->message);
+		return;
+	}
+
+	/* get dependency information */
+	for (i = 0; package_ids[i] != NULL; i++) {
```
ximion: nitpick
```suggestion
	for (guint i = 0; package_ids[i] != NULL; i++) {
```

backends/dnf/pk-backend-dnf.c
```diff
@@ -1841,6 +1841,197 @@ dnf_utils_find_package_ids (DnfSack *sack, gchar **package_ids, GError **error)
...
+
+	while (!g_queue_is_empty (queue)) {
+		g_autoptr(DnfPackage) curr = g_queue_pop_head (queue);
+		g_autoptr(DnfReldepList) reldeps = NULL;
+		guint n;
+
+		if (role == PK_ROLE_ENUM_DEPENDS_ON)
+			reldeps = dnf_package_get_requires (curr);
+		else
+			reldeps = dnf_package_get_provides (curr);
```
ximion: Is this correct? "Provides" in many package-managers means "packages/dependencies provided by package X" and not "packages that require package X to function" (which is what `REQUIRED_BY` is).

This might be okay in DNF semantics, so I am just asking to double-check (same below).

Conan-Kudo: It's okay because further down we invert the test for the query to find things that either provide the requires or require the provides. :)

backends/dnf/pk-backend-dnf.c
Conan-Kudo: Oh I see, combined with the thing above.

Conan-Kudo:
> I think at least a lot of string-duplication could be avoided for the emitted hash table, however I have not looked into the ownership semantics of the DNF API, so, it might very well be that the allocations are the best and safest way to do this.

This is admittedly a very cautious implementation.

backends/dnf/pk-backend-dnf.c
```diff
@@ -1841,6 +1841,197 @@ dnf_utils_find_package_ids (DnfSack *sack, gchar **package_ids, GError **error)
...
+		       &package_ids,
+		       &recursive);
+
+	/* set state */
+	ret = dnf_state_set_steps (job_data->state, NULL,
+				   50, /* add repos */
+				   49, /* find packages */
+				   1, /* emit */
+				   -1);
+	g_assert (ret);
```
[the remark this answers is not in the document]
Conan-Kudo: Probably? But we do this in other parts of this code already... If there's a better way to do it, it should be fixed across the board all at once.

backends/dnf/pk-backend-dnf.c
```diff
@@ -1841,6 +1841,197 @@ dnf_utils_find_package_ids (DnfSack *sack, gchar **package_ids, GError **error)
...
+	guint i;
+	DnfState *state_local;
+	DnfPackage *pkg;
+	PkBackendDnfJobData *job_data = pk_backend_job_get_user_data (job);
+	PkBitfield filters;
+	PkRoleEnum role;
+	g_autofree gchar **package_ids = NULL;
+	g_autoptr(DnfSack) sack = NULL;
+	g_autoptr(GError) error = NULL;
+	g_autoptr(GHashTable) hash = NULL;
```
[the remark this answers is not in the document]
Conan-Kudo: I don't entirely disagree, but other functions in this file basically do the same thing, so it's an established pattern, so I think I'll leave it for now.

PR #496, opened 2021-07-05
aptcc: fix serveral bugs relating to installing and searching packages

liushuyu:
This Pull Request will fix a few bugs relating to installing and searching packages when using the `aptcc` backend, namely:

- When installing a package that is not the latest available version, the backend will either treat the package as does not exist or install a version that is different than requested
- When downgrading a package, the backend will incorrectly reinstall the package using the current version instead of performing a downgrade to the specified version
- When searching/resolving a package or package ID, the backend will fail to list all the available versions or just fail to list anything

### DIFF
--- a/backends/aptcc/apt-cache-file.cpp
+++ b/backends/aptcc/apt-cache-file.cpp
@@ -382,9 +382,12 @@ PkgInfo AptCacheFile::resolvePkgID(const gchar *packageId)
 
     const pkgCache::VerIterator &candidateVer = findCandidateVer(pkg);
     // check to see if the provided package isn't virtual too
-    if (candidateVer.end() == false &&
-            strcmp(candidateVer.VerStr(), parts[PK_PACKAGE_ID_VERSION]) == 0)
-        return PkgInfo(candidateVer, piAction);
+    // also iterate through all the available versions
+    for (auto candidate = candidateVer; !candidate.end(); candidate++) {
+        if (strcmp(candidate.VerStr(), parts[PK_PACKAGE_ID_VERSION]) == 0) {
+            return PkgInfo(ver, piAction);
+        }
+    }
 
     return PkgInfo(ver, piAction);
 }
--- a/backends/aptcc/apt-intf.cpp
+++ b/backends/aptcc/apt-intf.cpp
@@ -442,7 +442,7 @@ void AptIntf::emitPackageProgress(const pkgCache::VerIterator &ver, PkStatusEnum
     g_free(package_id);
 }
 
-void AptIntf::emitPackages(PkgList &output, PkBitfield filters, PkInfoEnum state)
+void AptIntf::emitPackages(PkgList &output, PkBitfield filters, PkInfoEnum state, bool multiversion)
 {
     // Sort so we can remove the duplicated entries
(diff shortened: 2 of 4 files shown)


dantti:
the code looks fine, any thoughts? @ximion @julian-klode

julian-klode:
I think I'd prefer if the backend refused to install a version that is not the candidate, and not allow downgrades. You can do those things in APTs to some degree of success, but it's all tricky, and downgrades might just fail; I don't see reason to expose this at a higher level.

liushuyu:
> I think I'd prefer if the backend refused to install a version that is not the candidate, and not allow downgrades. You can do those things in APTs to some degree of success, but it's all tricky, and downgrades might just fail; I don't see reason to expose this at a higher level.

I am afraid you misunderstood. The root cause here is that the backend only reports one version to PackageKit. This would lead to disastrous issues where, if an update transaction would _downgrade_ a certain package due to dependency, PackageKit would possibly destroy the system because an incorrect version is selected.

Also, previously, listing other candidate versions of a package is impossible. Unless this is intentional, I could only see it as a bug.

julian-klode:
I think only reporting one version is sort of correct; picking the right alternative cannot (should not) be done sensibly by a frontend. PackageKit should not be trying to downgrade packages. It should expose candidate and installed versions; the candidate is never a downgrade (unless forced by extreme pinning).

There are no other candidate versions. Candidate version is the term that describes the version that apt will install. There is always just one such version per package, and that version is determined by pinning and the installed version.

ximion:
@liushuyu I think a good compromise here is to refuse downgrades, but allow installations of older versions when the package in question wasn't previously installed.
Either way, the current behavior is a bug. Can you please rebase your patch against the current master version of PK?

liushuyu:
> @liushuyu I think a good compromise here is to refuse downgrades, but allow installations of older versions when the package in question wasn't previously installed. Either way, the current behavior is a bug. Can you please rebase your patch against the current master version of PK?

Rebased. I have tested the rebased branch locally, and I did not find any new issues.

ximion:
I couldn't test this yet, but the changes do make sense to me an definitely improve the status quo, even though I don't like that this would actually make downgrades work again in aptcc ;-). Can you squash the two "aptcc: correctly lists all the available package versions" and "aptcc: fix a regression introduced in the last commit" together? (makes NEWS file writing as well as any future bisecting easier).

liushuyu:
> Can you squash the two "aptcc: correctly lists all the available package versions" and "aptcc: fix a regression introduced in the last commit" together? (makes NEWS file writing as well as any future bisecting easier).

Rebased the last two commits and melt them into one.

> I couldn't test this yet, but the changes do make sense to me an definitely improve the status quo, even though I don't like that this would actually make downgrades work again in aptcc ;-). 

The byproduct of fixing package listing is the downgrading being fixed as well. I still think only listing the most recent package version is a bug in downstream developers' eyes (since this backend is the only few that the behaviour deviates from the "majority" of the backends). And you are right, downgrading is not feasible most of the time but on certain occasions, this is unavoidable (e.g. opt-out of a testing repository like Debian backports).

ximion:

Issue #504, opened 2021-09-10
1.2.4 regression with offline updates/upgrades


mcrha:
The https://github.com/PackageKit/PackageKit/commit/9b7e083cf849c4ed4d66fe32250f1615ab577d94 causes a regression in offline upgrades, as reported here: https://bugzilla.redhat.com/show_bug.cgi?id=2002609

The user should be asked for the root credentials when changing offline updates/upgrades, but he/she is not asked for them. The PkOffline API doesn't have any flag for the interactivity, but it also doesn't make much sense here, because these offline updates/upgrades always require the credentials.

pwithnall:
The fix here is to add an interactive flag to the PkOffline API and propagate it to polkit, not to assume that all callers of that code will always want user interaction. Some systems may configure their polkit rules differently so that offline updates don’t need admin authorisation. Other callers of these code paths in `pk-engine.c` (written in future?) may not want user interaction.

mcrha:
> Some systems may configure their polkit rules differently so that offline updates don’t need admin authorisation. Other callers of these code paths in `pk-engine.c` (written in future?) may not want user interaction.

How does for example gnome-software know that? It currently sets the interactive flag for its job structure.

pwithnall:
> How does for example gnome-software know that? It currently sets the interactive flag for its job structure.

Not sure I understand what you’re referring to. gnome-software sets the interactive flag for a `GsPluginJob` if and only if the job is run from a plugin which is currently interactive (see `gs_plugin_job_set_interactive()` and `gs_plugin_interactive_inc()`). The `GsPlugin` interactivity handling could be better in gnome-software.

Or are you referring to something else?

mcrha:
I meant, the gnome-soft sets the interactivity flag unconditionally. Your argument is that some distros may not need the polkit prompt, but in that case the gnome-software should not set the interactivity flag, otherwise the user would be asked for the root credentials pointlessly. Thus there should be some way to know when the prompt makes sense and when not, thus when the gnome-software should have set interactivity flag and when not.

pwithnall:
> Your argument is that some distros may not need the polkit prompt, but in that case the gnome-software should not set the interactivity flag, otherwise the user would be asked for the root credentials pointlessly

That is incorrect. If the polkit rules are set up by the distro to not require admin authorisation, a prompt will not be displayed even if interactive=True.

[...]

PR #505, opened 2021-09-10
pk-engine: Always allow user interaction for offline update/upgrade

mcrha:
This partly reverts commit 9b7e083cf849c4ed4d66fe32250f1615ab577d94.
The offline update/upgrade triggers do need the user interaction, to
get the credentials from the polkit, otherwise the trigger cannot
do its job.

Closes https://github.com/PackageKit/PackageKit/issues/504

### DIFF
--- a/lib/packagekit-glib2/pk-offline.c
+++ b/lib/packagekit-glib2/pk-offline.c
@@ -95,8 +95,37 @@ pk_offline_action_from_string (const gchar *action)
 	return PK_OFFLINE_ACTION_UNKNOWN;
 }
 
+static GDBusCallFlags
+pk_offline_flags_to_gdbus_call_flags (PkOfflineFlags flags)
+{
+	if ((flags & PK_OFFLINE_FLAGS_INTERACTIVE) != 0)
+		return G_DBUS_CALL_FLAGS_ALLOW_INTERACTIVE_AUTHORIZATION;
+	return G_DBUS_CALL_FLAGS_NONE;
+}
+
 /**
  * pk_offline_cancel:
+ * @cancellable: A #GCancellable or %NULL
+ * @error: A #GError or %NULL
+ *
+ * Cancels the offline operation that has been scheduled. If there is no
+ * scheduled offline operation then this method returns with success.
+ * The function always allows user interaction. To change the behavior,
+ * use pk_offline_cancel_with_flags().
+ *
+ * Return value: %TRUE for success, else %FALSE and @error set
+ *
+ * Since: 0.9.6
+ **/
+gboolean
+pk_offline_cancel (GCancellable *cancellable, GError **error)
+{
+	return pk_offline_cancel_with_flags (PK_OFFLINE_FLAGS_INTERACTIVE, cancellable, error);
+}
+
+/**
+ * pk_offline_cancel_with_flags:
+ * @flags: bit-or of #PkOfflineFlags
  * @cancellable: A #GCancellable or %NULL
  * @error: A #GError or %NULL
  *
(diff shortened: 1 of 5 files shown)


hughsie:
@pwithnall please

pwithnall (changes requested):
I don’t think this is the right fix. Whether to allow user interaction should only ever be set at the top level of a call stack, typically where some code is being called from the UI or from a background service. Only then is it possible to know whether user interaction is definitely allowed/not allowed.

On the basis of what’s written in #504 I think the right fix would be to add an `interactive` property to the relevant bit of `PkOffline`.

mcrha:
The only problem is that the PkOffline is not an object, it's a call wrapper with bare functions. It's doable, of course.

lruzicka:
I would like to add that this issue is an accepted Beta blocker issue on Fedora 35, so it should be resolved possibly without much hesitation. If the correct fix takes much time to do, maybe the suggested fix could be used over the Beta release and revisited afterwards?

mcrha:
I just updated the branch to follow what the #504 suggests as a better solution.

lib/packagekit-glib2/pk-offline.c
```diff
@@ -144,10 +173,31 @@ pk_offline_cancel (GCancellable *cancellable, GError **error)
...
+	return pk_offline_clear_results_with_flags (PK_OFFLINE_FLAGS_NONE, cancellable, error);
+}
+
+/**
+ * pk_offline_clear_results_with_flags:
+ * @flags: bit-or of #PkOfflineFlags
+ * @cancellable: A #GCancellable or %NULL
+ * @error: A #GError or %NULL
+ *
+ * Crears the last offline operation report, which may be success or failure.
```
pwithnall, line 192:
```suggestion
 * Clears the last offline operation report, which may be success or failure.
```

lib/packagekit-glib2/pk-offline.h
```diff
@@ -71,6 +71,21 @@ typedef enum
 	PK_OFFLINE_ERROR_LAST
 } PkOfflineError;
 
+/**
+ * PkOfflineFlags:
+ * @PK_OFFLINE_FLAGS_NONE:		No specific flag
+ * @PK_OFFLINE_FLAGS_INTERACTIVE:	Run the action in an interactive mode
```
pwithnall, line 77:
```suggestion
 * @PK_OFFLINE_FLAGS_INTERACTIVE:	Run the action in an interactive mode, allowing polkit authentication dialogs
```

pwithnall (commented):
Two really minor nitpicks, but otherwise this looks great to me (I’m not a maintainer though)

lib/packagekit-glib2/pk-offline.c
mcrha: Copy&paste thing, the original function has it too ;)

mcrha:
I squashed your suggestions to the main commit.

hughsie (approved):
Great stuff, thanks.

mcrha:
One thing I forgot to mention. The old pk_offline() functions call as non-interactive. Would it make sense to rather use the interactive mode for those always, as it's better to ask the credentials with the old API than to "always" fail?

hughsie:
> as it's better to ask the credentials with the old API than to "always" fail?

So "working more than before" or "behaviour change". I don't mind either way. I guess I'd prefer it worked in more cases, but I'm worried about things like openQA changing. I'll defer to your judgement here.

lruzicka:
> > as it's better to ask the credentials with the old API than to "always" fail?
> 
> So "working more than before" or "behaviour change". I don't mind either way. I guess I'd prefer it worked in more cases, but I'm worried about things like openQA changing. I'll defer to your judgement here.

As for OpenQA, I guess we will have to deal with changes, but we'd prefer if the behaviour was somehow consistent and we did know what happens under certain conditions :D

pwithnall:
I think it’s probably safest to have the default be to behave the same as before https://github.com/PackageKit/PackageKit/commit/9b7e083cf849c4ed4d66fe32250f1615ab577d94, since that was quite a recent change. That would maintain backwards compatibility for callers. So that means allowing user interaction by default.

mcrha:
I agree it makes sense to allow the user interaction for the "old" functions, that's what it used to be until recently. I updated the patch accordingly.

PR #515, opened 2021-10-30
Replace apper with Discover

HanYoung-uwu:
[Apper](https://invent.kde.org/system/apper/-/commits/master) is pretty much in maintenance mode, [Discover](https://invent.kde.org/plasma/discover/-/commits/master) is under active development.

### DIFF
--- a/docs/html/pk-users.html
+++ b/docs/html/pk-users.html
@@ -131,15 +131,15 @@
 </tr>
 <tr>
  <td>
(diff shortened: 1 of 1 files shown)


ximion:
tbh, I don't think replacing it is fair, as Apper is the original KDE PackageKit frontend and contains support for all the features PK has, while Discover is an app-focused frontend like GNOME Software.
So just adding it to the list, maybe before Apper, would have been better.

PR #525, opened 2021-12-22
alpm: Offer the database we create when updating

### LINKED ISSUE #491: [alpm] pk_alpm_find_pkg can't find packages in get-updates database.
Since commit a76711fd450a326da97bd90e085734c20a852e00, packagekit uses a separate pacman database for get-updates, to prevent leaving the system in a partially updated state. As a side effect, this prevents get-deails and get-update-details from getting info about the available updates, since pk_alpm_find_pkg only searches the "real" pacman database so it can't find the updated packages. 

This causes erros in frontends such as Discover:
https://bugs.kde.org/show_bug.cgi?id=438353
https://bugs.archlinux.org/task/71196

aleixpol:
We create a separate database when we refresh to make sure we don't
leave the system's at an odd state. In turn, this leaves the current
database at an odd state.

This makes our query operations use our database instead of the system's
while it's alive, so other transactions also can know about the
information we fetched from our remotes.

Also stop leaking the temporary database.

Fixes #491

### DIFF
--- a/backends/alpm/pk-alpm-packages.c
+++ b/backends/alpm/pk-alpm-packages.c
@@ -82,7 +82,7 @@ pk_alpm_find_pkg (PkBackendJob *job, const gchar *package_id, GError **error)
 	if (g_strcmp0 (repo_id, "installed") == 0) {
 		db = priv->localdb;
 	} else {
-		const alpm_list_t *i = alpm_get_syncdbs (priv->alpm);
+		const alpm_list_t *i = alpm_get_syncdbs (priv->alpm_check ? priv->alpm_check : priv->alpm);
 		for (; i != NULL; i = i->next) {
 			const gchar *repo = alpm_db_get_name (i->data);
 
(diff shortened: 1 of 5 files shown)


antonio-rojas:
Thanks for taking care of this. I had to manually remove /var/lib/PackageKit/alpm to make it see new updated the first time, after that it seems to work fine. I'll keep testing for some more time.

antonio-rojas:
> Thanks for taking care of this. I had to manually remove /var/lib/PackageKit/alpm to make it see new updates the first time, after that it seems to work fine. I'll keep testing for some more time.

All seems fine after some more testing. Discover show updates size correctly, and no more weird "can't find package" toast messages.

aleixpol:
As it's already merged downstream, I see little reason to wait for further review. It still would be appreciated so I'll keep it open for a few days otherwise I'll merge.

hughsie (approved):

Issue #450, opened 2021-01-20
Kernel updates marked manual


private-lock:
Hi

Reported this already [to launchpad](https://bugs.launchpad.net/ubuntu/+source/unattended-upgrades/+bug/1675079/comments/37) and [KDE](https://bugs.kde.org/show_bug.cgi?id=430296).

Applying kernel-updates through apt command-line, they are marked as "automatic" and fit for removal by a subsequent call of "apt autoremove" later, once more kernels supersede it. (WORKS AS EXPECTED)

But when I apply any new version of the kernel through the KDE/Plasma update notification, then it gets marked as manual (BUG) ... even if I clean-out the manual status before - this is what it looks right before and after applying updates:

root@Meerschweinchen:/home/holger# apt-mark showmanual 'linux-.*'
linux-base
linux-generic
root@Meerschweinchen:/home/holger# apt-mark showmanual 'linux-.*'
linux-base
linux-generic
linux-headers-5.4.0-64
linux-headers-5.4.0-64-generic
linux-image-5.4.0-64-generic
linux-modules-5.4.0-64-generic
linux-modules-extra-5.4.0-64-generic

Versions:
root@Meerschweinchen:/home/holger# lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.1 LTS
Release:        20.04
Codename:       focal
root@Meerschweinchen:/home/holger# uname -a
Linux Meerschweinchen 5.4.0-62-generic #70-Ubuntu SMP Tue Jan 12 12:45:47 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
root@Meerschweinchen:/home/holger# apt list packagekit discover
Auflistung... Fertig
discover/focal 2.1.2-8 amd64
packagekit/focal-updates,focal-security,now 1.1.13-2ubuntu1.1 amd64  [installiert]

mauromol:
I have the same problem in Kubuntu 20.04, my boot partition got full and I discovered this problem. Kernel updates installed through the Discover UI (through the "update all" button) are marked as manually installed, hence unattended-upgrades/apt autoremove cannot automatically delete old kernerls like they should, as documented at: https://help.ubuntu.com/community/RemoveOldKernels

hklene:
Meanwhile I switched to Kubuntu 20.10 - but the bug persists. I made a video for you of an update.

Latest versions:
root@Meerschweinchen:/home/holger# lsb_release -a
[...]

PR #529, opened 2022-02-01
aptcc: Keep automatic state from previously resolved transactions when applying them

ximion:
Hi!
As far as I could test, this patchset, which features quite a bit of annoying refactoring, resolves the issue outlined in https://github.com/PackageKit/PackageKit/issues/450 (kernel updates marked as manually installed while they were in fact dragged in by an update).

While looking into the aptcc code after such a long time, I also quickly discovered why doing that is not at all my favourite activity ;-)

In all my tests the new code behaved correctly when updating as well as performing any other package management activities, but of course, more testing is helpful.

@dantti Can you maybe have a look at this?

### DIFF
--- a/backends/aptcc/apt-cache-file.cpp
+++ b/backends/aptcc/apt-cache-file.cpp
@@ -355,7 +355,7 @@ bool AptCacheFile::isRemovingEssentialPackages()
     return false;
 }
 
-pkgCache::VerIterator AptCacheFile::resolvePkgID(const gchar *packageId)
+PkgInfo AptCacheFile::resolvePkgID(const gchar *packageId)
 {
     g_auto(GStrv) parts = nullptr;
     pkgCache::PkgIterator pkg;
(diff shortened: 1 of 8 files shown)


ximion:
Extra question for @dantti and possibly @julian-klode since he may actually know this: The code on https://github.com/PackageKit/PackageKit/blob/main/backends/aptcc/apt-intf.cpp#L2247-L2258 runs `tryToInstall` twice on every package, once with `AutoInst` for  `MarkInstall` set to `false` and once to `true` - why are we doing this? Wouldn't we simply always have `AutoInst` set to true, and report an error if that failed?

dantti:
This line wasn't written by me :) 
for (auto autoInst : { false, true }) {
I'll take a look at the code and see if I find some time to test it...

ximion:
> This line wasn't written by me :) for (auto autoInst : { false, true }) { I'll take a look at the code and see if I find some time to test it...

Ha! If only I hadn't assumed this code was extremely old and just used blame on it - @apachelogger explained why this code exists in the commit that introduced it: https://github.com/PackageKit/PackageKit/commit/ea201923418c47c64209c0b1e33f2a0f6bbea21d

So, questions answered :-) It's not directly relevant to this PR anyway, I just came across this piece of code while writing this change and wondered whet was going on.

julian-klode:
Oh I see the comment from the commit message now, this probably should be recorded in a comment in the code too.

ximion:
Thank you for the explanation, @julian-klode! I'll add this to the code as comment, so if anyone else wonders about this they'll have the answer right there.
On other matters, does newer APT maybe have some API that could help us get rid of this status parsing abomination: https://github.com/PackageKit/PackageKit/blob/main/backends/aptcc/apt-intf.cpp#L1894-L1966 ?
As is obvious, as soon as the locale isn't C/English, this all falls apart anyway (and it is almost always set to something else so we can translate debconf messages)

ximion:
@dantti (or anyone else): Do you see anything that should be changed before merging this? (I'll have another go over it, and I'm also running this on my computer by default now, nothing broke and updates are marked properly)

julian-klode:
@ximion I forwarded this to @DonKult  - I don't know anything

ximion:
> I don't know anything

Absolutely untrue, but thanks! :-)

DonKult:
Not sure what I am supposed to comment here as I have neither experience with PackageKit & co as a dev nor as a user – my knowledge level of anything related to apt front ends is very little above "they exist".

As far as the progress reporting goes, that hasn't changed in years. So, if PackageKit used to need that, it very likely still does. Not sure what it is doing through. Looks like it wants to know a per-package progress which apt doesn't provide as it is too much guess work. A tiny bit of ugliness might be avoidable via implementing `APT::Progress::PackageManager` instead of using the Fd implementation and parsing its output, but that still only gets the translated message. Not sure why that interface isn't getting the `action` (in `apt-pkg/deb/dpkgpm.cc`) but knowing how dpkg moves through these is probably not something anyone wants to support, so it might be for the best. I guess apts guess what the per-package progress is might be easier to expose, but then I don't see much point in it as a random spinner would be just as informative.

Anyway, probably better to talk about that in a dedicated bugreport with a clear "that is what we need" rather than in a subthread of a relatively unrelated pull request somewhere else.

ximion:
@DonKult Thank you for the comment!
> Looks like it wants to know a per-package progress which apt doesn't provide as it is too much guess work.

That's exactly what it is trying to do...

> I guess apts guess what the per-package progress is might be easier to expose, but then I don't see much point in it as a random spinner would be just as informative.

For fairly long package installations, a status update in the GUI is actually pretty useful. At the moment, progress there is jumping around a lot...

> Anyway, probably better to talk about that in a dedicated bugreport with a clear "that is what we need" rather than in a subthread of a relatively unrelated pull request somewhere else.

True :-)
I am inclined to merge this change (after looking through it again), as there's just very few people who can review aptcc PRs...

DonKult:
On Wed, Feb 02, 2022 at 05:19:49AM -0800, Matthias Klumpp wrote:
> > I guess apts guess what the per-package progress is might be easier to expose, but then I don't see much point in it as a random spinner would be just as informative.
> 
> For fairly long package installations, a status update in the GUI is actually pretty useful. At the moment, progress there is jumping around a lot...

How does that help though? Usually a package installs rather quickly,
so these 0, 25, 50, 75, 100 substeps flash by in an instant? Worse, if
you have a monster package unpacking forever I as user would assume the
process is stuck at 25% … hence me referring to the random spinner as
a visual indication of "I am doing my thing, be patient" as being just
as useful together with the slowly progressing global progress (which is
reasonably accurate now that apt displays it itself).

In an ideal world dpkg would tell us how far along it is with unpacking
the package and how that relates to all the other things it is supposed
to be doing, but I doubt even dpkg really knows that given all the
layers of compression, path exclusions, conditional execution paths and
such (and, to be fair, dpkg has no idea of "all the other things" its
supposed to do as apt tends to call it multiple times).

And then there are of course the maintainer scripts who helpfully
print a message that they will now eat up a considerable amount of CPU
time to build some caches or what not, so in an even more ideal world
the maintainer scripts would… oh well, its not soooo bad is it? ;P

ximion:
@DonKult I don't disagree with this at all :-D
Personally, I would be fine with a download progress bar and then with a per-package spinner for the actual installation. I may actually implement this for PackageKit, although currently I far this may require at least an API addition (but hopefully can be added without breaking things, maybe the API even allows for that). I thought this may also help with https://github.com/PackageKit/PackageKit/issues/159, but looks like for that using APTs own global progress after downloading would be *much* better (I think aptcc's code is ancient there and could use an update...)

PR #530, opened 2022-02-14
backends/nix: update use of forceValue for Nix 2.7.0 compatibility

aakropotkin:
Nix 2.7.0 changes the declaration of `forceValue`, removing a default parameter which PackageKit previously relied on. This causes PackageKit builds using Nix 2.7.0 to fail during compilation.

This patch simply adds the old default value explicitly to the use of `forceValue` to match the legacy behavior, and fix compilation failures with Nix 2.7.0.

This change is backwards compatible with Nix < 2.7.0 - and will produce identical calls to `forceValue` in a version agnostic manner.

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -56,5 +56,5 @@ endif
 
 install_data(
   'transactions.db',
-  install_dir: join_paths('/mnt/ludata/repos/PackageKit/outputs/out', 'var', 'lib', 'PackageKit'),
+  install_dir: join_paths(get_option('localstatedir'), 'lib', 'PackageKit'),
 )
--- a/data/meson.build
+++ b/data/meson.build
@@ -56,5 +56,5 @@ endif
 
 install_data(
   'transactions.db',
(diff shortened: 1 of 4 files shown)


data/meson.build
```diff
@@ -56,5 +56,5 @@ endif
 
 install_data(
   'transactions.db',
-  install_dir: join_paths(get_option('localstatedir'), 'lib', 'PackageKit'),
+  install_dir: join_paths('/mnt/ludata/repos/PackageKit/outputs/out', 'var', 'lib', 'PackageKit'),
```
hughsie, line 59: I don't think you want this committed, right? :)

aakropotkin:
Good catch. You're right. Just the code changes

data/meson.build
aakropotkin: I created a new PR : https://github.com/PackageKit/PackageKit/pull/532

aakropotkin:
A new PR has been created which corrects previously mentioned code-review issues.

https://github.com/PackageKit/PackageKit/pull/532

PR #537, opened 2022-03-02
docs: replace apper with discover

amifelipek:


docs/html/pk-using.html
```diff
@@ -70,9 +70,8 @@ <h2>Using graphical tools:</h2>
 <code>gnome-software</code> and <code>gnome-packagekit</code> provide a rich
 set of GTK tools for automatically updating your computer and installing software.
 </p>
-<a href="http://kde-apps.org/content/show.php/Apper?content=84745">
-Apper</a> is the name of the KDE graphical tools designed
-for PackageKit.
+<a href="https://apps.kde.org/discover/">
+Discover</a> is the name of the KDE interface for PackageKit.
```
aleixpol, line 74: KDE Plasma interface?

aleixpol (commented):
+1 otherwise

docs/html/pk-using.html
FHEK789: I used the same terminology as https://github.com/PackageKit/PackageKit/pull/515

aleixpol:
Fair enough.

I see @ximion has thoughts on the topics so I'd wait for his feedback.

PR #543, opened 2022-04-17
zypp: ignore locked packages

### LINKED ISSUE #325: PackageKit does not honor zypper package locks
opensuse leap 15.1
packagekit 1.1.10

for some purpose i lock a package with the software manager according to avoid any update

then

i ask to software update plasmoid to check if there is updates

then

in the list of the packages to update there is the locked package !
it must not be in this list.

rodrigopedra:
Closes #325 

Currently if a package is locked on ZYpp PackageKit ignores the lock and silently updates or installs it.

This PR adds a check if a package is locked a before adding it to the packages to be updated list.

Implementation and comment in code was inspired by the one in aptcc backend

https://github.com/PackageKit/PackageKit/blob/9339249866d44f06d713e92f2abf90b470b5deb3/backends/aptcc/apt-intf.cpp#L1303-L1308

### DIFF
--- a/backends/zypp/pk-backend-zypp.cpp
+++ b/backends/zypp/pk-backend-zypp.cpp
@@ -1244,7 +1244,7 @@ zypp_get_package_updates (string repo, set<PoolItem> &pks)
 			// We pretend locked packages are not upgradable at all since
 			// we can't represent the concept of holds in PackageKit.
 			// https://github.com/PackageKit/PackageKit/issues/325
-			continue;	
+			continue;
 		} else if (it->status().isToBeInstalled()) {
 			ui::Selectable::constPtr s =
 				ui::Selectable::get((*it)->kind(), (*it)->name());
--- a/backends/zypp/pk-backend-zypp.cpp
+++ b/backends/zypp/pk-backend-zypp.cpp
@@ -1239,13 +1239,19 @@ zypp_get_package_updates (string repo, set<PoolItem> &pks)
 		resolver->doUpdate ();
 	}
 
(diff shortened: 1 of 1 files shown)


JonathanKang:
Looks good to me.

backends/zypp/pk-backend-zypp.cpp
```diff
@@ -1239,13 +1239,19 @@ zypp_get_package_updates (string repo, set<PoolItem> &pks)
...
 	}
 
-	for (; it != e; ++it)
-		if (it->status().isToBeInstalled()) {
+	for (; it != e; ++it) {
+		if (it->status().isLocked()) {
+			// We pretend locked packages are not upgradable at all since
+			// we can't represent the concept of holds in PackageKit.
+			// https://github.com/PackageKit/PackageKit/issues/325
+			continue;
```
JonathanKang, line 1247: trailing white space.

rodrigopedra: I removed it, thanks!

rodrigopedra:
Not sure if this can count on deciding if this PR gets merged or not, but as you can see the reference above, this PR's code was added as a patch to openSUSE's build version of PackageKit, and has been deployed to users using openSUSE Tumbleweed.

As I am a Tumbleweed's user, I already have the patched version, and KDE Discover is working as expected by not ignoring locks and also not installing unwanted locked packages.

JonathanKang:
@hughsie The pull request looks like. Can you help merge this? Thanks.

PR #547, opened 2022-05-17
adapt for uos(uniontech os)

kklionz:
Dear Maintainers,
         To Adapt for UOS(UNIONTECH OS), lib/packagekit-glib2/pk-common.c is modified.
         UOS gets variable $releasever for field "MinorVersion" from file "/etc/os-version".
         Pls help to review it, thanks

         PS: Maybe UOS's handling is strange, but it is exactly...

UOS file "/etc/os-version" is as following
```
[Version] 
SystemName=UnionTech OS Server
SystemName[zh_CN]=统信服务器操作系统
ProductType=Server
ProductType[zh_CN]=服务器
EditionName=e
EditionName[zh_CN]=e
MajorVersion=20
MinorVersion=1050
OsBuild=12038.101
```
UOS file "/etc/os-release" is as following
```
PRETTY_NAME="UnionTech OS Server 20"
NAME="UnionTech OS Server 20"
VERSION_ID="20"
VERSION="20"
ID="uos"
PLATFORM_ID="platform:uel20"
HOME_URL="https://www.chinauos.com/"
BUG_REPORT_URL="https://bbs.chinauos.com/"
VERSION_CODENAME="fuyu"
```

And corresponding repo URL is as following 
``
https://euler-packages.chinauos.com/server-euler/fuyu/1050/everything/x86_64
``

### DIFF
--- a/lib/packagekit-glib2/pk-common.c
+++ b/lib/packagekit-glib2/pk-common.c
@@ -247,6 +247,7 @@ pk_parse_os_release (gchar **id, gchar **name, gchar **version_id, GError **erro
 	g_autofree gchar *contents = NULL;
 	g_autoptr(GKeyFile) key_file = NULL;
 	g_autoptr(GString) str = NULL;
+	gboolean isuos;
 
 	/* load data */
 	if (!g_file_test (filename, G_FILE_TEST_EXISTS))
@@ -264,14 +265,22 @@ pk_parse_os_release (gchar **id, gchar **name, gchar **version_id, GError **erro
 	}
 
 	/* get keys */
-	if (id != NULL) {
-		g_autofree gchar *tmp = g_key_file_get_string (key_file, "os-release", "ID", error);
+	g_autofree gchar *idtmp = g_key_file_get_string (key_file, "os-release", "ID", error);
+	if (NULL == idtmp)
+		return FALSE;
+	if (0 == g_strncasecmp(idtmp, "uos", -1))
+		isuos = TRUE;
+	else
+		isuos = FALSE;
+	if (NULL != id)
+	{
 		if (tmp == NULL)
 			return FALSE;
 		*id = g_shell_unquote (tmp, error);
 		if (*id == NULL)
 			return FALSE;
 	}
+
 	if (name != NULL) {
 		g_autofree gchar *tmp = g_key_file_get_string (key_file, "os-release", "NAME", error);
 		if (tmp == NULL)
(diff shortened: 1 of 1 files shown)


ximion:
What is this code trying to accomplish? If UOS ships a valid `/etc/os-release`, it should just work and there's no need to add any distro-specific handling.
Also, the repository URL gives me a 401 error code, are you sure it is correct / the repository is public?
To get PackageKit support, you will more likely want a backend for your package manager, not this patch. I'm not sure what UOS is based on, but you have a pretty good chance that one of the existing backends will work for you.

kklionz:
It is the mainpage of UOS
``
https://www.uniontech.com/
``
The PR is try to get correct "$releasever" in UOS.
UOS gets variable $releasever for field "MinorVersion" from file "/etc/os-version".
Getting 401 err code is because that UOS repo website has a authentication procedure through a 3rd-party website. To visit UOS repo website, a username and password must be provided.

Different from dnf and yum, DNF and yum get "$releasever" form rpm db, and PackageKits get this variable from file.
So, this PR is to adapt for UOS.

hughsie:
> /etc/os-version

That's not a specification, right? I don't think we should support `/etc/os-version` *and* `/etc/os-release` -- I think UOS just needs to modify the later to include the correct value. Lots of different software parses `/etc/os-release` and I don't think we want to modify everything...

kklionz:
```
/etc/os-version
That's not a specification, right? 
```
Yes :)

Sure, ```/etc/os-version``` is not a specification file and it is more reasonable to get ```"$releasever"``` from ```/etc/os-release```; but UOS uses ```/etc/os-version```  indeed.
Among these popular package management software, there are two ways to get  ```"$releasever"```.
1) Getting it from rpm database, such as yum and dnf
2) Getting it from some ini files, such as PackageKit.

If this PR can not be accepted, another way is OK?

hughsie:
Just to clarify, you want `MinorVersion=1050` to override `VERSION="20"` -- in which case, why?

kklionz:
UOS's repo URL is related to "MinorVersion" from ```/etc/os-version```  instead of ```VERSION``` from ```/etc/os-release```

When the ```MinorVersion``` is ```1050```, the UOS repo URL is ```https://euler-packages.chinauos.com/server-euler/fuyu/1050/everything/x86_64```

When the ```MinorVersion``` is ```1020```, the UOS repo URL is ```https://euler-packages.chinauos.com/server-euler/fuyu/1020/everything/x86_64```

kklionz:
Dear hughsie,
    U r right.
    With some inexplicable reason，UOS uses ```VERSION``` as a more big release project plan, and uses ```MinorVersion``` as a specific release version; and the repo url is distingushed with ```MinorVersion```.

hughsie:
It sounds like you want to define something like:

    VERSION_ID="1050"
    VERSION="20"

and then patch the tools that want the version for display to read the 2nd key, rather than teach *everything* that reads os-release to also read os-version... I mean, are you going to patch fwupd in the same way too? systemd?

kklionz:
This PR is only for getting correct ```$releasever``` in UOS, which variable is for parsing repo URL.
 fwupd and systemd have no variable ```$releasever``` and these package need not get URL through ```$releasever```， so these packages shall not be patched

hughsie:
I really think you should use:

    VERSION_ID="1050"
    VERSION="20"

-- is there a reason why you can't fix the os-release data?

kklionz:
Dear hughsie,
    I recheck the code, and it maybe have no relation to packagekit how to parse ```$releasever```
    Thanks for u discussing and this PR can be closed

PR #549, opened 2022-05-31
pk-client: Fix default value of cache-age property

pwithnall:
`pk_client_init()` sets the default value to `G_MAXUINT` but the
`GParamSpec` setup didn’t also use that value.

This probably introduces no functional changes, but it makes the code
not contradict itself.


ximion (approved):
LGTM

PR #550, opened 2022-05-31
D-Bus signal performance improvements

pwithnall:
Here are a few changes to improve PackageKit’s use of D-Bus signals to improve performance.

They are basically complete, but I haven’t polished them yet as I’d like some feedback on whether the approach is acceptable before spending time on that.

The headline is that they reduce the number of D-Bus signal emissions from packagekitd → gnome-software when starting gnome-software by about a factor of 10, from roughly 5000. This reduces gnome-software’s startup time from about 13.0s to about 12.5s (with a warm `packagekitd` process and warm libxmlb caches).

The changes are as follows, but please see the commit messages for more details:
 * Add a new `Transaction.Packages` signal to go alongside `Transaction.Package`. Emit it when there’s more than one package to be sent to the client (which is often). This is the commit which needs more polishing (such as adding a flag to enable it so that we maintain backwards compatibility for older clients), but it works (for the dnf backend; I do not plan to implement it for other backends, but it will be easy enough for others to do so).
 * Combine multiple separate `PropertiesChanged` signal emissions, since the signal supports sending an array of changed properties.
 * Emit progress-related `PropertiesChanged` signals on a timer so that they are aggregated and rate-limited. Without this, packagekitd often sends multiple `Percentage` updates every millisecond.

### DIFF
--- a/src/pk-transaction.c
+++ b/src/pk-transaction.c
@@ -103,6 +103,7 @@ struct PkTransactionPrivate
 	PolkitSubject		*subject;
 	GCancellable		*cancellable;
 	gboolean		 skip_auth_check;
+	gboolean		 client_supports_plural_signals;
 
 	/* needed for gui coldplugging */
 	gchar			*last_package_id;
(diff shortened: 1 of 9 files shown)


hughsie:
As a general concept, PK is only lightly maintained and we've not changed the DBus API in a long time. I think quite a bit of stuff would break unless there's some kind of backwards compat thing in place. I'm not sure I have a huge amount of appetite to add API tbh. @ximion might disagree tho, and that's fine.

ximion:
Hehe... I think we should be very careful changing the API, but I also think we should definitely still make improvements when possible and helpful for frontends. PK is still in wide use, and it really does not look like it will go away any time soon (I expected that native package managers would eventually replace it and implement the PK API, but it doesn't look like this will happen in any near future). I actually remember that back in 2013 I even identified the many-`Package` emissions as one of the things slowing PK down, and I'm kind of glad that this may get addressed.
IF we break things, I would really like for this to happen only once, so we can adjust all clients in one go rather than continuously (so we should make all the changes that we think are useful at once), or maybe support two APIs for a bit until the clients are adjusted.

I do think that 90% of all clients will use the PackageKitQt and PackageKitGLib2 bindings rather than the raw DBus API nowadays, so I am a bit more worried about these ;-)
(a quick search in the Debian archive revealed nothing using the raw DBus API currently, so stuff that does so may not be in Debian or may not be open source)

So, tl;dr, if @pwithnall sees a substantial benefit for existing software (and from the measurements there is one!), then I think it's worth making a change like this. But I will of course listen to @hughsie if he thinks this is a terrible idea and @pwithnall and I will get flooded with bug reports if we would pursue this ;-)

pwithnall:
> I think quite a bit of stuff would break unless there's some kind of backwards compat thing in place

I intend to make this completely backwards compatible if we go ahead with it. I just haven’t done that yet because I wanted feedback on the overall idea before spending more time on it.

It sounds like I should go ahead with adding backwards compatibility and polish now though?

pwithnall:
OK I’ve updated this to add a feature flag for it, so it is backwards compatible, and done a bit of polishing.

I have a few other follow-up ideas but I’ll submit them as separate MRs to avoid making this one too hard to review. I can split the existing `PropertiesChanged` commits out of this MR if that would help with parallelising review.

Conan-Kudo (approved):
This looks reasonable to me.

hughsie:
I'm conflicted. On one hand PackageKit is supposed to be in maintenance mode, but it's not **my** project and so I shouldn't be the one to veto things if someone else is going to marshal releases and fix bugs...

ximion:
I think I would be fine with "maintaining" it, meaning making the occasional release and doing basic bugfixing - I certainly wouldn't start any major feature development on PK though. I do know PK's code well enough though and have been working on it for a decade now that I think this super-slow development model would definitely work for it.
At the very least, doing a release from time to time is needed for backend maintainers to get their changes into a release :-)

As for this PR, I looked over it and it looks great to me at first glance, but I want to give it a slightly closer look in the next few days (just to make sure there isn't anything I overlooked) - sorry for taking a bit long with that, but I was really short on time the past two weeks.

src/pk-transaction.c
```diff
@@ -1232,6 +1233,143 @@ pk_transaction_package_cb (PkBackend *backend,
...
+
+	/* Emit the signal. Grouping multiple package details into a single
+	 * signal reduces the number of signals and hence the amount of context
+	 * switching between packagekitd, dbus-daemon and the client process.
+	 * This results in much improved performance compared to emitting one
+	 * signal per package.
+	 *
+	 * This should not hit the D-Bus limits (maximum array size of 64MB,
+	 * maximum message size of 128MB) until it’s listing on the order of
+	 * 100000 packages. */
```
ximion, line 1463: Should we add a check for that and split the array into multiple `Packages` signals in that case? I got very wary of "should not happen" cases after Fedora started its quest for "how many packages can we build from a texlive distribution?", which made us change default limits in PackageKit multiple times.
Not that I think we will *ever* overflow this during normal operation, but it might be some nice debugging aid if we just fail here with a nice error message.

ximion (approved):
I finally got to reviewing this, it's an excellent patchset with really good documentation! I initially didn't like that it duplicates quite a bit of existing code from the existing `Package` signal, but I also don't see a better way to do this, so I think it's just fine the way it is.

Adding an array-size sanity check might be nice to have - the maximum message size is so high that it exceeds the size of the entire Debian archive by ~1.7x, so if that limit is hit it will most likely be due to a backend bug.

If @hughsie doesn't veto the change, I would merge it and also add support for this feature in PK's APT backend. I think this change is very valuable, and reminds me of the days when I was optimizing PK by adding multi-threading support - implementing this feature in addition would have helped immensely back then on slower machines :-)

src/pk-transaction.c
Conan-Kudo: For what it's worth, I believe TeXLive is now at somewhere north of 7k subpackages now, so 10k might actually be too low.

hughsie:
> If @hughsie doesn't veto the change

I'm in no position to veto anything :)

ximion:
I'm waiting for some feedback from @pwithnall on possibly adding some limits to the `Packages` array size, or auto-splitting it in case it gets too big, and then this could be merged :-)

src/pk-transaction.c
pwithnall: > Should we add a check for that and split the array into multiple Packages signals in that case? I got very wary of "should not happen" cases after Fedora started its quest for "how many packages can we build from a texlive distribution?", which made us change default limits in PackageKit multiple times.

I don’t think we should go as far as adding code to split one over-long `Packages` signal into multiple smaller signals. That code would essentially never get tested, so would be full of bugs. Instead, I’ve pushed a fixup commit which falls back to the old loop of `Package` emissions if emitting the `Packages` signal fails (which it will do if the payload is too long).

That means that if we reach 100000 packages, packagekit will regress to emitting single `Package` signals for all of them. But it will still function correctly, and at that point the code can be reworked safely, because it’s being tested.

The fixup commit will need to be squashed in before the PR is merged. I kept it separate for ease of re-review. Aside from squashing it in, this PR is ready to be merged from my PoV.

ximion:
This has been merged now, thanks @pwithnall for the patch! :-)

aleixpol:
Now we need this one https://github.com/PackageKit/PackageKit/pull/585 or PackageKit Qt breaks.

aleixpol:
And now this one ports packagekit Qt to use plurals.
https://github.com/hughsie/PackageKit-Qt/pull/37

PR #557, opened 2022-06-21
zypp: ensure locked packages are neither installed, updated, removed, nor downloaded

rodrigopedra:
On my previous PR #543 I proposed a change to not list locked packages as being able to be updated. 

Which is fine, as one would not expect a locked package to be listed as being able to be to updated.

But locked packages could still be installed by PackageKit. For example, zypper by default install recommended packages, as can be seen on this configuration file, from their official repository:

https://github.com/openSUSE/zypper/blob/7a03e8594bb8dbf2af68decc4427d5244f68f46c/zypper.conf#L70-L82

So when a locked package is listed in the recommended listing for another package, it ends up being resolved by the backend, and then installed.

I tracked every occurrence where libzypp API was used to mark a package to be installed, updated, or removed, and added a guard to ignore locked packages.

For example, if you look into the function `backend_remove_packages_thread`, the guard was added before calling `item.status ().setToBeUninstalled()` or any other manual status change.

~~~cpp
// Added guard here ...
if (zypp_package_is_locked(solvable))
    continue;

PoolItem item(solvable);
if (solvable.isSystem ()) {
    // ... to prevent this
    item.status ().setToBeUninstalled (ResStatus::USER);
    items.push_back (item);
} else {
    item.status ().resetTransact (ResStatus::USER);
}
~~~

In addition I also added a guard to skip locked packages from being downloaded.

### DIFF
--- a/backends/zypp/pk-backend-zypp.cpp
+++ b/backends/zypp/pk-backend-zypp.cpp
@@ -1861,6 +1861,13 @@ zypp_is_no_solvable (const sat::Solvable &solv)
 	return solv == sat::Solvable::noSolvable;
 }
 
+static bool
+zypp_package_is_locked (const sat::Solvable &solv)
+{
+	ui::Selectable::Ptr sel( ui::Selectable::get( solv ) );
+	return sel->locked();
+}
+
 /**
   * backend_required_by_thread:
   */
@@ -1898,6 +1905,9 @@ backend_required_by_thread (PkBackendJob *job, GVariant *params, gpointer user_d
 						     "Package couldn't be found");
 			return;
 		}
+
+		if (zypp_package_is_locked(solvable))
+			continue;
 
 		PoolItem package = PoolItem(solvable);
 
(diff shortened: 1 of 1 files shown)


rodrigopedra:
@bmwiedemann this PR is a follow up of patch https://github.com/bmwiedemann/openSUSE/commit/7d61442da609091e49c2f869fdaa79fdcbf65051 proposed by OpenSUSE OBS user dimstar.

If you, or him, can review this I would really appreciate. 

I didn't find dimstar's github handle to tag him instead.

bmwiedemann:
@DimStar77 ; could also be interesting to @mlandres

DimStar77:
> @bmwiedemann this PR is a follow up of patch [bmwiedemann/openSUSE@7d61442](https://github.com/bmwiedemann/openSUSE/commit/7d61442da609091e49c2f869fdaa79fdcbf65051) proposed by OpenSUSE OBS user dimstar.
> 
> If you, or him, can review this I would really appreciate.
> 
> I didn't find dimstar's github handle to tag him instead.

The original patch actually comes from @JonathanKang - care to chime in, please?

PR #559, opened 2022-07-02
pk-main: Add newline on g_printerr() calls

pwithnall:
It doesn’t add one itself (by contrast, `g_message()` and friends do add
one).


ximion:
Looks good, but one of those lines even has a hint for translators but no Gettext prefix - maybe these strings were even intended to be localized originally?

ximion (approved):

pwithnall:
> Looks good, but one of those lines even has a hint for translators but no Gettext prefix - maybe these strings were even intended to be localized originally? (In other words, maybe we do want to add l10n to these as well?)

⇒ https://github.com/PackageKit/PackageKit/pull/561

PR #561, opened 2022-07-04
pk-main: Mark some early-start error messages as translatable

pwithnall:
As suggested by Matthias on
https://github.com/PackageKit/PackageKit/pull/559#issuecomment-1173005774

### DIFF
--- a/src/pk-main.c
+++ b/src/pk-main.c
@@ -165,13 +165,15 @@ main (int argc, char *argv[])
 	conf = g_key_file_new ();
 	conf_filename = pk_util_get_config_filename ();
 	if (conf_filename == NULL) {
-		g_printerr ("Config file was not found.\n");
+		g_printerr ("%s\n", _("Config file was not found."));
 		goto out;
 	}
 	ret = g_key_file_load_from_file (conf, conf_filename,
 					 G_KEY_FILE_NONE, &error);
 	if (!ret) {
-		g_printerr ("Failed to load config file: %s\n", error->message);
+		/* TRANSLATORS: The placeholder is an error message */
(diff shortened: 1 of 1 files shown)


src/pk-main.c
```diff
@@ -165,13 +165,13 @@ main (int argc, char *argv[])
 	conf = g_key_file_new ();
 	conf_filename = pk_util_get_config_filename ();
 	if (conf_filename == NULL) {
-		g_printerr ("Config file was not found.\n");
+		g_printerr (_("Config file was not found.\n"));
```
hughsie, line 168: I don't think the translators like newlines; e.g. I think you want:

g_printerr ("%s\n", _("Config file was not found."));

src/pk-main.c
```diff
@@ -196,7 +196,7 @@ main (int argc, char *argv[])
 	if (backend_name == NULL || g_strcmp0 (backend_name, "auto") == 0) {
 		ret  = pk_util_set_auto_backend (conf, &error);
 		if (!ret) {
-			g_printerr ("Failed to resolve auto: %s\n", error->message);
+			g_printerr (_("Failed to resolve auto: %s\n"), error->message);
```
hughsie, line 199: same:

    g_printerr ("%s: %s\n", _("Failed to resolve auto"), error->message);

and TRANSLATOR comment please.

pwithnall: Can’t use a format string `%s: %s` as the translators might want to quote the error message or add text after it.

I’ve used an intermediate variable

hughsie: I don't think we've done that in other places tho, right?

pwithnall: No reason to get it wrong here too

PR #562, opened 2022-07-04
pk-transaction: Add UpdateDetails signal

pwithnall:
Like the `Packages` signal added a few commits previously, the
`UpdateDetails` signal is a plural version of the existing
`UpdateDetail` signal.

This has reduced the number of signals emitted during starting up
gnome-software from 632 `UpdateDetail` signals (for a corresponding
number of pending dnf updates) to 1 `UpdateDetails` signal.

The new signal is only used if the client sets the
`supports-plural-signals=true` hint on a transaction. Otherwise the old
`UpdateDetail` signal will be used. This means the new signal is
backwards-compatible.

If the hint is set, the new signal will be used in parallel with the old
one, with some operations or backends using the old signal, and some
using the new one. The content of the signals will not be duplicated
(i.e. if `UpdateDetails` is emitted for N updates, there will not also
be N `UpdateDetail` signal emissions in parallel).

The new signal is currently only implemented for the dnf backend. Other
backends can implement it at their leisure.

### DIFF
--- a/backends/dnf/dnf-backend.c
+++ b/backends/dnf/dnf-backend.c
@@ -53,13 +53,43 @@ dnf_emit_package_list (PkBackendJob *job,
 		       PkInfoEnum info,
 		       GPtrArray *pkglist)
 {
-	guint i;
-	DnfPackage *pkg;
+	g_autoptr(GPtrArray) pk_packages = g_ptr_array_new_full (pkglist->len, (GDestroyNotify) g_object_unref);
 
-	for (i = 0; i < pkglist->len; i++) {
-		pkg = g_ptr_array_index (pkglist, i);
-		dnf_emit_package (job, info, pkg);
+	for (guint i = 0; i < pkglist->len; i++) {
+		DnfPackage *dnf_package;
+		g_autoptr(PkPackage) pk_package = NULL;
+		PkInfoEnum package_info;
+		PkInfoEnum update_severity;
+		const gchar *package_id;
+		g_autoptr(GError) local_error = NULL;
+
+		dnf_package = g_ptr_array_index (pkglist, i);
+		package_id = dnf_package_get_package_id (dnf_package);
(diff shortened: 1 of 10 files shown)


pwithnall:
This is based on (and depends on #550), so only the top two commits in this branch actually need separate review.

Conan-Kudo (approved):

pwithnall:
I’ve rebased this on #550 (though it will need to be rebased again before it can land), and have incorporated similar changes to https://github.com/PackageKit/PackageKit/pull/550/commits/276368d5cbf6e86e589de4a9db8b50bd9e4c24aa to fall back to the `UpdateDetail` emission in a loop if emitting `UpdateDetails` fails for whatever reason.

ximion:
> [I] have incorporated similar changes to [276368d](https://github.com/PackageKit/PackageKit/commit/276368d5cbf6e86e589de4a9db8b50bd9e4c24aa) to fall back to the `UpdateDetail` emission in a loop if emitting `UpdateDetails` fails for whatever reason.

Awesome, thank you! The patch looks fine to me!

ximion (approved):

pwithnall:
Do you want me to rebase, or will you do that when you merge?

ximion:
The patch has been merged (and rebased with no modifications on top of the other changes).
Thanks! :-)

ximion:
> Do you want me to rebase, or will you do that when you merge?

Ha, I was a few seconds faster ;-)
If you can have a quick look if anything is missing, that would be great (but the code runs fine here, at least).

pwithnall:
It’s all landed OK, although for some reason the two commits in this branch have been applied out of order, so the source tree won’t compile between them. But that’s not really an issue. Thanks for landing it!

ximion:
>  for some reason the two commits in this branch have been applied out of order

Dang, you spotted my interactive-rebase mistake that I noticed after pushing, but thought was too minor to fix ;-)
I do want to implement this for the APT backend, and then ideally get a new PK release, as many changes have accumulated that will help PK clients a lot, this one being one of the most important ones :-)

PR #568, opened 2022-08-25
Fix documentation of package-install policy

mcatanzaro:
The documentation says that unprivileged users can install signed packages. This has not been true since d462e2e62d700d0d29c41a8d9c4bbffbe99339ff from 2007. The comment is extremely stale. Fix it to match reality.


ximion:
This is indeed true, I remember the argument we had about changing this, even! Looks like the comment text was simply forgotten.
Thank you for the patch!

PR #570, opened 2022-08-26
trivial: aptcc: copy a string directly to std::string, which is expec…

imz:
…ted below

Below, each of the many invocations of starts_with(str, ...) takes
std::string as the argument, so str would anyway be converted to
std::string each time (and even copied again each time).

Improves: 323cc7483 trivial: aptcc: Some trivial cosmetic changes

(What really caused me to do this change was a non-GCC compiler having
troubles with that implicit conversion after the type of str was
annotated with g_autofree, which in its turn we had to change to
g_autofree_edg as another workaround. But nevertheless, this change can
be justified on its own, as at the top of this commit message.)

### DIFF
--- a/backends/aptcc/apt-intf.cpp
+++ b/backends/aptcc/apt-intf.cpp
@@ -1743,7 +1743,7 @@ void AptIntf::updateInterface(int fd, int writeFd, bool *errorEmitted)
             const gchar *status   = g_strstrip(split[0]);
             const gchar *pkg      = g_strstrip(split[1]);
             const gchar *percent  = g_strstrip(split[2]);
-            g_autofree gchar *str = g_strdup(g_strstrip(split[3]));
+            const std::string str = g_strstrip(split[3]);
 
             // major problem here, we got unexpected input. should _never_ happen
             if(pkg == nullptr && status == nullptr)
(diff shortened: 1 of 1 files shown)


backends/aptcc/apt-intf.cpp
```diff
@@ -1743,7 +1743,7 @@ void AptIntf::updateInterface(int fd, int writeFd, bool *errorEmitted)
             const gchar *status   = g_strstrip(split[0]);
             const gchar *pkg      = g_strstrip(split[1]);
             const gchar *percent  = g_strstrip(split[2]);
-            g_autofree gchar *str = g_strdup(g_strstrip(split[3]));
+            std::string const str = g_strstrip(split[3]);
```
dantti, line 1746: Mind changing to `const std::string` matching the rest of the style

dantti (commented):
I wonder why we duplicate the string... also wonder why not length check on array but seems it was never an issue :)

ximion:
> I wonder why we duplicate the string...

Yeah, that indeed looks a bit pointless...

backends/aptcc/apt-intf.cpp
imz: Good advice on the rest of the style! Pushed a new revision.

imz:
> > I wonder why we duplicate the string...
> 
> Yeah, that indeed looks a bit pointless...

Yes, one could think about removing more bogus code here as the next step. (Perhaps, when introducing strdup, the one who did that was not sure about the lifetime of the `split` array and its members... Or wanted to modify the string in-place, but my use of `const` proves that nothing like this is done anywhere in the code.)

dantti:
My guess is that I (I guess it was me) was maybe unsure about if the split would always have that number of parts, but code too old and I'm not very familiar with glib stuff also...

Issue #571, opened 2022-08-31
Add an option to use system settings for apt/dpkg force-conf


debug-richard:
I am currently working on a project that uses the PackageKit to update a Debian based embedded device.

By default, each update should overwrite the configuration files with the new version, which can be configured via /etc/apt/apt.conf.d/ and /etc/dpkg/dpkg.cfg (force-confnew). (I am aware of the problems this can cause)

The problem is that the aptcc backend enforces "force-confold" https://github.com/PackageKit/PackageKit/blob/7f0bc4b256d9c26449f2480d0e0229bbd24a2285/backends/aptcc/apt-intf.cpp#L167
and overrides the system setting in this way.

I could not find a way to disable this behavior without recompiling.

It would be great if PackageKit had a configuration option (via /etc/PackageKit/ ?) to just use the system settings for these parameters.

ximion:
It only overrides these settings if in noninteractive mode. Maybe we can do something clever here and not override the settings if a suitable one is already set.

PR #572, opened 2022-09-01
aptcc: Detect if apt system settings specify --force-conf* options

debug-richard:
This refers to #571 .
As suggested by @ximion, I created a patch that detects if the --force-conf* option is set and if it is, PackageKit does not change it.

This allows the admin to set a system-wide default behavior for config file updates, which is also used by PackageKit.

I also looked  at the [dpkg code](https://github.com/guillemj/dpkg/blob/1e61d9416719b628ec0e93ced10fe4c1f38ef105/src/main/configure.c#L123)  to find out what these options really do, because the documentation on them is confusing.
First of all, these settings are not affected by "DEBIAN_FRONTEND=noninteractive", which is only used for (post install) configuration prompts (accept eula, setup database...).

All three options disable the interactive prompt for configuration files.
--force-confdef **ignores confold/confnew** and keeps the old config files and installs new ones
--force-confold is only applied if confdef is **not** set and does the same as confdef
--force-confnew is only applied if confdef is **not** set and installs the files from the new package

If both --force-confdef and --force-confnew are specified, the latter is ignored, which is quite confusing.
--force-confmiss and --force-confask are not relevant if any of the other options are set.

### DIFF
--- a/backends/aptcc/apt-intf.cpp
+++ b/backends/aptcc/apt-intf.cpp
@@ -163,8 +163,13 @@ bool AptIntf::init(gchar **localDebs)
     m_interactive = pk_backend_job_get_interactive(m_job);
     if (!m_interactive) {
         // Do not ask about config updates if we are not interactive
-        _config->Set("Dpkg::Options::", "--force-confdef");
-        _config->Set("Dpkg::Options::", "--force-confold");
+        if (!isSystemDpkgConf()) {
+            _config->Set("Dpkg::Options::", "--force-confdef");
+            _config->Set("Dpkg::Options::", "--force-confold");
+        } else {
+            // If any option is set we should not change anything
+            cout << "Using system settings for --force-conf*" << endl;
+        }
         // Ensure nothing interferes with questions
         g_setenv("APT_LISTCHANGES_FRONTEND", "none", TRUE);
         g_setenv("APT_LISTBUGS_FRONTEND", "none", TRUE);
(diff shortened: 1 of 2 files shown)


backends/aptcc/apt-intf.cpp
```diff
@@ -163,8 +163,13 @@ bool AptIntf::init(gchar **localDebs)
...
     if (!m_interactive) {
         // Do not ask about config updates if we are not interactive
-        _config->Set("Dpkg::Options::", "--force-confdef");
-        _config->Set("Dpkg::Options::", "--force-confold");
+        if (!isSystemDpkgConf()) {
+            _config->Set("Dpkg::Options::", "--force-confdef");
+            _config->Set("Dpkg::Options::", "--force-confold");
+        } else {
+            // If any option is set we should not change anything
+            cout << "Using system settings for --force-conf*" << endl;
```
ximion, line 171: Please use `g_debug`

backends/aptcc/apt-intf.cpp
```diff
@@ -174,6 +179,22 @@ bool AptIntf::init(gchar **localDebs)
     return m_cache->CheckDeps(AllowBroken);
 }
 
+bool AptIntf::isSystemDpkgConf() {
```
ximion, line 182: That's a bit of a miseleading name - maybe something like `isDpkgForceConfSet` is a better name?

ximion (changes requested):

ximion:
I'll merge this and fix my comments afterwards, as I am currently working on some refactoring and it would be nice to have this change in early :-)
Also, the commends were not critical :-)

Thank you for the patch!

PR #575, opened 2022-09-10
Fix issue #342 core-dump using alpm backend with hookdirs config set

### LINKED ISSUE #342: PackageKit service fails with a core-dump due to unsupported directive HookDir in pacman.conf.

adrianinsaval:
~Note that I have not tested this,~ just looked at 

https://github.com/PackageKit/PackageKit/blob/7e700e2177ca3f84ea5a2978cdf687179b3445a9/backends/alpm/pk-alpm-config.c#L1017
and 
https://github.com/devkitPro/pacman/blob/49b08fa9d14a9cc06f01c624c644cb987ce97642/lib/libalpm/alpm.h#L1732

and thought it was a copy paste error.

### DIFF
--- a/backends/alpm/pk-alpm-config.c
+++ b/backends/alpm/pk-alpm-config.c
@@ -362,6 +362,7 @@ typedef struct
 /* keep this in alphabetical order */
 static const PkAlpmConfigList pk_alpm_config_list_options[] = {
 	{ "HoldPkg", G_STRUCT_OFFSET (PkAlpmConfig, holdpkgs) },
+	{ "HookDir", G_STRUCT_OFFSET (PkAlpmConfig, hookdirs) },
 	{ "IgnoreGroup", G_STRUCT_OFFSET (PkAlpmConfig, ignoregroups) },
 	{ "IgnorePkg", G_STRUCT_OFFSET (PkAlpmConfig, ignorepkgs) },
 	{ "LocalFileSigLevel", G_STRUCT_OFFSET (PkAlpmConfig,
@@ -370,7 +371,6 @@ static const PkAlpmConfigList pk_alpm_config_list_options[] = {
 	{ "NoUpgrade", G_STRUCT_OFFSET (PkAlpmConfig, noupgrades) },
 	{ "RemoteFileSigLevel", G_STRUCT_OFFSET (PkAlpmConfig,
 						 remotefilesiglevels) },
-	{ "HookDir", G_STRUCT_OFFSET (PkAlpmConfig, hookdirs) },
 	{ NULL, 0 }
 };
 
@@ -1014,7 +1014,9 @@ pk_alpm_config_configure_alpm (PkBackend *backend, PkAlpmConfig *config, GError 
 	config->noupgrades = NULL;
 
 	/* alpm takes ownership */
-	alpm_option_set_noupgrades (handle, config->hookdirs);
+	for(alpm_list_t *j = config->hookdirs; j; j = j->next) {
+		alpm_option_add_hookdir(handle, j->data);
+	}
 	config->hookdirs = NULL;
 
 	pk_alpm_config_configure_repos (backend, config, handle, error);


aleixpol:
Can you maybe test it then?

adrianinsaval:
Sorry, I'm short of time to figure out how to compile this right now, perhaps next week. But that there is an error in the code is pretty evident IMO

adrianinsaval:
now I actually tested and it does fix the issue after correcting the alphabetical order of `pk_alpm_config_list_options`, there was a comment there specifying that the order should be respected but this was not taken into account for that change.
Also confirmed that the option is respected and the hooks in the custom dir are being run

adrianinsaval:
This is ready for merge IMO, I fixed another bug where the system hook dir (default `/usr/share/libalpm/hooks/`) was being overridden and ignored instead of just adding the additional hookdirs as it was supposed, the fix is based on this: https://gitlab.archlinux.org/pacman/pacman/-/blob/master/lib/libalpm/handle.c#L452

backends/alpm/pk-alpm-config.c
```diff
@@ -1014,7 +1014,10 @@ pk_alpm_config_configure_alpm (PkBackend *backend, PkAlpmConfig *config, GError
 	config->noupgrades = NULL;
 
 	/* alpm takes ownership */
-	alpm_option_set_noupgrades (handle, config->hookdirs);
+	alpm_list_t *j;
+	for(j = config->hookdirs; j; j = j->next) {
```
hughsie, line 1018: can you put the alpm_list_t in the loop please, and squash all these down to a single commit pls

adrianinsaval: I can squash no problem but I don't know how to put alpm_list_t inside the loop, it's the variable used to iterate, don't I need it declared beforehand?

hughsie: `for(alpm_list_t *j = config->hookdirs; `

adrianinsaval: done

adrianinsaval:
thanks, any clue when this would be in a release? In the mean time I'll ask again the archlinux packagers to apply this patch.

PR #578, opened 2022-09-12
Revert "Revert "Shutdown the daemon on idle by default""

### LINKED ISSUE #460: Possible memory leak with dnf backend
packagekitd consumes what I think is an unreasonable amount of memory, especially for a daemon intended to be running at all times. On my system it quickly consumes several hundred MB and after some time typically grows beyond 800MB resident memory.

When profiled on fedora you you can see it allocates upwards of 60MB of heap space per what ill call "a bad query". When viewing the systemd unit you can see that the heap profiler reports ~150GB of allocations to satisfy the query before reporting
`get-update-detail transaction /2154_ceebebbb from uid 1000 finished with success after 265536ms`.

Attached is the pdf of the pprof output (i can provide the actual heap dumps as well from the full run if desired).
[dump.pdf](https://github.com/hughsie/PackageKit/files/6134308/dump.pdf)

Tested on fedora 33.
PackageKit 1.2.2
Probable caller of the "bad query" gnome-software.

yump:
This reverts commit dca1f5b2508a4632d0b9fefab771a5a9caf83a5c.

Which reverted commit 0c84d71509e851db20445c747529bd7d3724f081, which reverted commit c6eb3555ec5b41e988c111d276764d55fb83bda3.

Fixes #460.  High memory usage was also reported on Redhat's bug tracker back in 2016, and a couple more times since:

https://bugzilla.redhat.com/show_bug.cgi?id=1354074 
https://bugzilla.redhat.com/show_bug.cgi?id=1854875 
https://bugzilla.redhat.com/show_bug.cgi?id=1896964

The memory usage of packagekitd without the timeout has been observed growing well beyond half a GiB. On two of my own machines running Fedora 36 as I write this, 537 MiB and 696 MiB. 

As I understand it, this timeout causes some slightly surprising behavior when users mix command line dnf upgades with GUI PackageKit upgrades, and do not manually run an update check before rebooting for update. But that is an edge case, and the price of not having it is too high.

### DIFF
--- a/src/pk-main.c
+++ b/src/pk-main.c
@@ -182,12 +182,14 @@ main (int argc, char *argv[])
 	syslog (LOG_DAEMON | LOG_DEBUG, "daemon start");
 
 	/* after how long do we timeout? */
-	exit_idle_time = g_key_file_get_integer (conf, "Daemon", "ShutdownTimeout", NULL);
+	exit_idle_time = g_key_file_get_integer (conf, "Daemon", "ShutdownTimeout", &error);
 	/* THIS COMMENT IS A TSUNAMI STONE
 	 * Before removing the default timeout, please study the git history and
 	 * be sure that you are not regressing Redhat bugzilla #1354074 (again). */
-	if (exit_idle_time == 0)
+	if (error != NULL) {
 		exit_idle_time = 300;
+		g_clear_error(&error);
+	}
 	g_debug ("daemon shutdown set to %i seconds", exit_idle_time);
 
 	/* override the backend name */
--- a/src/pk-main.c
+++ b/src/pk-main.c
@@ -183,9 +183,9 @@ main (int argc, char *argv[])
 
 	/* after how long do we timeout? */
 	exit_idle_time = g_key_file_get_integer (conf, "Daemon", "ShutdownTimeout", NULL);
-        /* THIS COMMENT IS A TSUNAMI STONE 
-         * Before removing the default timeout, please study the git history and
-         * be sure that you are not regressing Redhat bugzilla #1354074 (again). */
+	/* THIS COMMENT IS A TSUNAMI STONE
+	 * Before removing the default timeout, please study the git history and
+	 * be sure that you are not regressing Redhat bugzilla #1354074 (again). */
 	if (exit_idle_time == 0)
 		exit_idle_time = 300;
 	g_debug ("daemon shutdown set to %i seconds", exit_idle_time);
--- a/src/pk-main.c
+++ b/src/pk-main.c
@@ -190,7 +190,7 @@ main (int argc, char *argv[])
          * regressing Redhat bugzilla #1354074 (again). */
 	if (error != NULL) {
 		exit_idle_time = 300;
-		g_clear_error(&error);
+		g_clear_error (&error);
 	}
 	g_debug ("daemon shutdown set to %i seconds", exit_idle_time);
 
--- a/src/pk-main.c
+++ b/src/pk-main.c
@@ -184,8 +184,10 @@ main (int argc, char *argv[])
 	/* after how long do we timeout? */
 	exit_idle_time = g_key_file_get_integer (conf, "Daemon", "ShutdownTimeout", &error);
 	/* THIS COMMENT IS A TSUNAMI STONE
-	 * Before removing the default timeout, please study the git history and
-	 * be sure that you are not regressing Redhat bugzilla #1354074 (again). */
+         * The automatic shutdown timeout is necessary to avoid very high
+         * memory usage with the dnf backend.  If you want to remove this
+         * timeout, please study the git history and be sure that you are not
+         * regressing Redhat bugzilla #1354074 (again). */
 	if (error != NULL) {
 		exit_idle_time = 300;
 		g_clear_error(&error);
--- a/src/pk-main.c
+++ b/src/pk-main.c
@@ -183,6 +183,11 @@ main (int argc, char *argv[])
 
 	/* after how long do we timeout? */
 	exit_idle_time = g_key_file_get_integer (conf, "Daemon", "ShutdownTimeout", NULL);
+        /* THIS COMMENT IS A TSUNAMI STONE 
+         * Before removing the default timeout, please study the git history and
+         * be sure that you are not regressing Redhat bugzilla #1354074 (again). */
+	if (exit_idle_time == 0)
+		exit_idle_time = 300;
 	g_debug ("daemon shutdown set to %i seconds", exit_idle_time);
 
 	/* override the backend name */
--- a/src/pk-main.c
+++ b/src/pk-main.c
@@ -192,7 +192,10 @@ main (int argc, char *argv[])
 		exit_idle_time = 300;
 		g_clear_error (&error);
 	}
-	g_debug ("daemon shutdown set to %i seconds", exit_idle_time);
+	if (exit_idle_time > 0)
+		g_debug ("daemon shutdown set to %i seconds", exit_idle_time);
+	else
+		g_debug ("daemon will not shut down on idle");
 
 	/* override the backend name */
 	if (backend_name != NULL) {
--- a/src/pk-main.c
+++ b/src/pk-main.c
@@ -184,10 +184,10 @@ main (int argc, char *argv[])
 	/* after how long do we timeout? */
 	exit_idle_time = g_key_file_get_integer (conf, "Daemon", "ShutdownTimeout", &error);
 	/* THIS COMMENT IS A TSUNAMI STONE
-         * The automatic shutdown timeout is necessary to avoid very high
-         * memory usage with the dnf backend.  If you want to remove this
-         * timeout, please study the git history and be sure that you are not
-         * regressing Redhat bugzilla #1354074 (again). */
+	 * The automatic shutdown timeout prevents memory leaks in some
+	 * backends from getting out of hand.  If you want to remove this
+	 * timeout, please study the git history and be sure that you are not
+	 * regressing Redhat bugzilla #1354074 (again). */
 	if (error != NULL) {
 		exit_idle_time = 300;
 		g_clear_error (&error);
--- a/src/pk-main.c
+++ b/src/pk-main.c
@@ -186,8 +186,8 @@ main (int argc, char *argv[])
 	/* THIS COMMENT IS A TSUNAMI STONE
 	 * The automatic shutdown timeout prevents memory leaks in some
 	 * backends from getting out of hand.  If you want to remove this
-	 * timeout, please study the git history and be sure that you are not
-	 * regressing Redhat bugzilla #1354074 (again). */
+	 * timeout, please study the Git history and be sure that you are not
+	 * regressing Red Hat bugzilla #1354074 (again). */
 	if (error != NULL) {
 		exit_idle_time = 300;
 		g_clear_error (&error);


hughsie:
This isn't a fix, it's a workaround. Somebody needs to fix libdnf.

yump:
It is my understanding that [libdnf is in the process of a major rewrite](https://fedoraproject.org/wiki/Changes/ReplaceDnfWithDnf5), including a new daemon with similar functionality to PackageKit.  Presumably the people with the best skills and knowledge to fix libdnf are engaged with that.

The branch name is honest, and it is an improvement over the status quo.  This is not a fix, but it is simple, doesn't pre-empt DNF5, and stops the bleeding. 512 MiB times [$2.56 / GiB](https://pcpartpicker.com/products/memory/#F=6000000,11000000&S=3200,6600&ff=ddr4&sort=ppgb&page=1) times between [70,000](https://www.reddit.com/r/Fedora/) and [200,000](https://data-analysis.fedoraproject.org/csv-reports/images/fedora-os-latest-stacked.png) people... is not inconsiderable.

Conan-Kudo:
Yeah, we're probably going to get a new PackageKit-DNF5 backend in the near future to deal with this longer-term.

Conan-Kudo (approved):

Conan-Kudo:
The problem is coming from hung gpg agents from librepo by libdnf, which is not easily fixed in dnf4...

yump:
> If you land this before you've fixed the regressions that it caused, then Fedora will need to revert it downstream. And since Fedora is the distro you're trying to fix, we'll be back to square one, except now everyone is dissatisfied.

Have I misunderstood the comment at [https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/1227#note_784327](https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/1227#note_784327)?  It sounds as if there only problems when command line updates are mixed with GUI updates, and I when I stopped packagekitd and ran Gnome Software, it seemed to auto-start packagekitd and function normally.

I've started a longer test -- mask packagekitd, wait a few days, update with dnf, un-mask pacakgekitd, and see if Gnome Software does anything unacceptably weird.  

While I'm waiting for the results of that, I suggest you boot your machine with `mem=4G` on the kernel command line and see how the other half lives.  Maybe you will not be so satisfied with spending 500-1000 MiB on supporting a use case that everyone warns against anyway.

If Fedora goes that way, I will of course be setting `ShutdownTimeout=60` in PackageKit.conf on every machine I administer, but I would rather [fix the problem at scale](https://blogs.gnome.org/tbernard/2020/01/17/doing-things-that-scale/).

For what it's worth, my mother, who uses Gnome Software to do her own updates, has been trained by half a lifetime of Windows usage that you have to check for updates, apply updates, reboot, check for updates again, maybe apply updates again, etc, until no available updates are reported.  She therefore feels perfectly at home with giving Gnome Software the necessary accommodations. Given how long the `dnf offline-upgrade` double reboot takes, it would seem reasonable for Gnome Software to synchronously do `refresh-cache; get-updates` *before* going down for updates, instead of relying on the signals.

yump:
@mcatanzaro Alright so the test is complete.  Here is exactly what I did, on a machine running Fedora 36:

1. `systemctl mask --now packagekit`, to prevent any automatic or accidental manual updates.
2. Wait a day for some updates to become available.
3. `systemctl unmask packagekit`
4. Open Gnome Software, go to update tab, click refresh button to check for updates. (packagekitd is started by dbus activation.)
5. Click "Download" for updates requiring reboot, and wait for it to complete.
6. Close Gnome Software.
7. `systemctl stop packagekit`
8. Update with `dnf up --refresh`.
9. `systemctl status packagekit`, to double-check it's still not running.
10. Reboot with the menu in Gnome shell's top bar, with the box checked to apply updates (even though there are no updates, because I just installed them with dnf).

And here is exactly what happened:

1. Machine reboots once and shows the "Installing updates" splash.
2. Progress bar goes from 0-100% nearly instantly.
3. Machine reboots again to the login screen and I log in as usual.

`dnf history --reverse` shows this:
```
ID     | Command line              | Date and time    | Action(s)      | Altered
--------------------------------------------------------------------------------
   551 | up --refresh              | 2022-09-14 16:14 | Upgrade        |   15   
   552 |                           | 2022-09-14 16:15 |                |    0   
```

The first transaction is the command line upgrade, and the second is the offline upgrade, which altered zero packages.  `dnf history info 552` reports:

```
Transaction ID : 552
Begin time     : Wed 14 Sep 2022 04:15:59 PM CDT
Begin rpmdb    : 6c42bc169442fa2b126755637f57753c197122238a0f9d3aad65ce638e9dd6b0
End time       : Wed 14 Sep 2022 04:15:59 PM CDT (0 seconds)
End rpmdb      : 6c42bc169442fa2b126755637f57753c197122238a0f9d3aad65ce638e9dd6b0
User           : root <root>
Return-Code    : Success
Releasever     : 
Command Line   : 
Comment        : 
Packages Altered:
```

That is, the extent of the horkage I observed is

1. The power menu says there are updates to apply on reboot, when that is not true.
2. This tricks the user into doing a no-op double-reboot (admittedly annoying with encrypted disks).

Do you have detailed instructions to reproduce the update failure [reported by Kalev Lember here](https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/1227#note_784327)?  Nothing I encountered in my test was anywhere near as bad as losing **500+ megabytes of RAM**.

And on that note... how has `mem=4G` been treating you?  [32% of Firefox users have 4 GB or less](https://data.firefox.com/dashboard/hardware).  When I tried that, I couldn't even open my Firefox session without it getting OOM killed, but to be fair I have a lot of windows open.

mcatanzaro:
> That is, the extent of the horkage I observed is
> 
> The power menu says there are updates to apply on reboot, when that is not true.
> This tricks the user into doing a no-op double-reboot (admittedly annoying with encrypted disks).

That's really not OK, though, so let's look into that. Looking at gnome-shell's endSessionDialog.js, I see that GNOME is actually receiving the info on prepared updates directly from PackageKit. (I had incorrectly assumed it was going via GNOME Software.) Opening the end session dialog should immediately refresh the update state from PackageKit, so this actually seems like it's probably a PackageKit bug that the state is wrong? That's probably good, because I was thinking something was wrong with GNOME Software that would need to be fixed there, but if the end session dialog is stale, then it's not involved. The code is [here](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/a4ae46a881fd5564aa5ea6d94ff44344f0513815/js/ui/endSessionDialog.js#L696). We can find out for certain by inspecting PackageKit's D-Bus properties when the end session dialog is in this bad state to see if PackageKit is providing bogus info. (PackageKit should get autostarted when the end session dialog is open.)

> Do you have detailed instructions to reproduce the update failure [reported by Kalev Lember here](https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/1227#note_784327)? Nothing I encountered in my test was anywhere near as bad as losing 500+ megabytes of RAM.

I don't, but I think your test was a pretty decent way to verify that whatever bug Kalev found _probably_ no longer occurs. Maybe a GNOME Software bug that has been fixed in the past two years? Based on the instructions you posted, I think you took reasonable care to ensure this bug no longer occurs.

> And on that note... how has mem=4G been treating you? [32% of Firefox users have 4 GB or less](https://data.firefox.com/dashboard/hardware). When I tried that, I couldn't even open my Firefox session without it getting OOM killed, but to be fair I have a lot of windows open.

Not brave enough, sorry. :P

gordonmessmer:
This week, I opened 4 memory-related PRs against PackageKit and libdnf, and on my local system packagekitd's RSS shrinks to around 75MB at idle, after instructing it to refresh data and invoking other interfaces like command-not-found.  I'm happy with progress here, but I've also run with ShutdownTimeout and this seems like a very usable configuration.

Is there any reason this change can't be merged (or that the default can't be changed in Fedora?)  Which is to say, specifically, @mcatanzaro do you have any objections to this change?

gordonmessmer:
> The problem is coming from hung gpg agents from librepo by libdnf, which is not easily fixed in dnf4...

Is there an open bug report that describes that in more detail?  I do see gpg-agent processes locally, but it doesn't seem like they use a *lot* of memory.

I was interested in trying to shut them down, but I actually couldn't identify where they'd been started.

Conan-Kudo:
> > The problem is coming from hung gpg agents from librepo by libdnf, which is not easily fixed in dnf4...
> 
> Is there an open bug report that describes that in more detail? I do see gpg-agent processes locally, but it doesn't seem like they use a _lot_ of memory.
> 
> I was interested in trying to shut them down, but I actually couldn't identify where they'd been started.

On my system, they don't result in memory leaks, but they do result in PK hanging when trying to quit during system shutdown until systemd kills its cgroup.

mcatanzaro:
So my only concern here is that last time we shut down PackageKit on idle, it [broke GNOME](https://github.com/PackageKit/PackageKit/commit/dca1f5b2508a4632d0b9fefab771a5a9caf83a5c). As long as that doesn't happen again, then I have no other concerns.

hughsie:
> As long as that doesn't happen again

I don't think there's anything added to stop this kind of breakage happening again.

gordonmessmer:
> So my only concern here is that last time we shut down PackageKit on idle, it broke GNOME

We have two descriptions of system behavior with auto shutdown.  @yump provided one here in September, indicating that after updating with dnf the GNOME UI would still reflect pending updates, and that the system would attempt to apply those updates, find that they were already installed, and reboot again.  As far as I can tell, Kalev Lember described the same situation, but said that that offline updates "just fail" without describing how they fail.

Between the two reports, yump's is more detailed, and it's hard to justify classifying GNOME's out-of-date indication that updates are pending as a major problem, especially since it would only affect users who know enough about command-line package management to prefer it for some reason.

ximion:
@hughsie 
> I don't think there's anything added to stop this kind of breakage happening again.

Can you maybe quickly summarize what the GNOME issue even is? I've scanned through the old bug reports, but I don't think I even understand the problem, and why it's supposed to be in PackageKit.
Keeping the PK process around feels quite wasteful in general, so it would actually be nice to get that timeout feature back (even though I really don't like that it seems to paper over bugs in other software).

mcatanzaro:
> Between the two reports, yump's is more detailed, and it's hard to justify classifying GNOME's out-of-date indication that updates are pending as a major problem, especially since it would only affect users who know enough about command-line package management to prefer it for some reason.

No, that's actually exactly the problem. The end session dialog needs to accurately reflect the state of available system updates even if dnf is used to apply updates behind PackageKit's back, or we'll get complaints about broken updates. Same applies for installed or removed packages: GNOME Software needs to know what's installed or what's not, or users will complain.

gordonmessmer:
If someone wrote a dnf plugin that signaled to packagekitd that it should wake up and refresh its view of whether or not there are updates pending, would that resolve the perceived risks of shutting down packagekit on idle?  Is that approach feasible?  (Which is to say, is there currently such a dbus API in packagekitd?)

gordonmessmer:
Yeah, this seems feasible.  With an appropriate plugin, dnf can notify PackageKit that updates have been applied (even if it is stopped at the time).  With that plugin installed, the GNOME UI will no longer present the user with the option to reboot and apply updates, which should clear Fedora to use ShutdownTimeout.

https://github.com/rpm-software-management/dnf-plugins-core/pull/479

ximion:
FWIW, this is the hook we use for APT/dpkg: https://github.com/PackageKit/PackageKit/blob/main/backends/apt/20packagekit - if DNF did the same, this issue would be gone, as far as I understand it.

gordonmessmer:
Thanks, @ximion .  I updated my PR to match the method used by the apt plugin, and verified that it also works.  

(Though for reasons I haven't looked into, RefreshCache causes gnome-software to refresh its view of updates available to download, and StateHasChanged doesn't seem to do that, it only updates gnome-software's view of whether updates are available to install after reboot.  The problem we've discussed here is still fixed, but it seems like StateHasChanged should do more than it does.)

Would it be better to include the dnf plugin in PackageKit, rather than dnf-plugins-core?  I can easily retract the other PR and add a new one on this project.

mcatanzaro:
Huh, nice. You're really taking care of everything. Yes, with this fixed, I _think_ we should be able to revert safely.

> Would it be better to include the dnf plugin in PackageKit, rather than dnf-plugins-core? I can easily retract the other PR and add a new one on this project.

No opinion.

mcatanzaro:
I actually kinda changed my mind. Keeping the plugin here in PackageKit rather than dnf-plugins-core seems _slightly_ nicer to me. And you've already proposed doing that in #600. I think this pull request is safe to land as soon as that one does.

src/pk-main.c
```diff
@@ -183,6 +183,11 @@ main (int argc, char *argv[])
 
 	/* after how long do we timeout? */
 	exit_idle_time = g_key_file_get_integer (conf, "Daemon", "ShutdownTimeout", NULL);
+        /* THIS COMMENT IS A TSUNAMI STONE 
+         * Before removing the default timeout, please study the git history and
+         * be sure that you are not regressing Redhat bugzilla #1354074 (again). */
+	if (exit_idle_time == 0)
+		exit_idle_time = 300;
```
mcatanzaro, line 190: Let's drop that comment:

```suggestion
	if (exit_idle_time == 0)
		exit_idle_time = 300;
```

yump: Let's not.  History speaks to its necessity, and I don't want this coming back again a few years down the road.

ximion: After dropping the change it's an inaccurate comment though, and even worse, it's a Fedora/DNF-specific workaround to an entirely different issue. So there's not much of a point to keeping a random comment in this place.
A comment with "there was a bug once for which there was a workaround here once" is quite odd...

src/pk-main.c
```diff
@@ -183,6 +183,11 @@ main (int argc, char *argv[])
 
 	/* after how long do we timeout? */
 	exit_idle_time = g_key_file_get_integer (conf, "Daemon", "ShutdownTimeout", NULL);
+        /* THIS COMMENT IS A TSUNAMI STONE 
+         * Before removing the default timeout, please study the git history and
+         * be sure that you are not regressing Redhat bugzilla #1354074 (again). */
+	if (exit_idle_time == 0)
```
ximion, line 189: I wonder if this should look like this:
```suggestion
	if (exit_idle_time < 0)
		exit_idle_time = 0;
	else if (exit_idle_time == 0)
		exit_idle_time = 300;
```
With https://github.com/PackageKit/PackageKit/blob/main/etc/PackageKit.conf#L17 being changed to say `Shut down the daemon after this many seconds idle. -1 means don't shutdown.`

That is because 0 is returned *both* if the `ShutdownTimeout` setting is set to 0, and if it isn't set at all, so if the user sets this to 0, we will still time out, which certainly isn't the desired behavior (not a bug in this patch, probably just a historical oversight).

ximion (changes requested):

src/pk-main.c
mcatanzaro: So here you are adding a comment that says "Before removing the default timeout" in the same commit that itself removes the default timeout. Hence, the new comment is immediately stale and incorrect. You have to at least edit it somehow to not be wrong.

src/pk-main.c
yump: Oh, it's for sure a bug in the patch, and trying to fix it I noticed that the comment was indented with tabs.

This whole thing seems terribly convoluted.  Following the propagation of `ShutdownTimeout`, it is read in as `int exit_idle_time`, then implicitly cast to `uint helper.exit_idle_time`, but only if already greater than zero so that's okay, then implicitly rounded up modulo 5 seconds by `g_timeout_add_seconds()`.

With -1 used thas way, it would be

-1 means no timeout  
0 means 300 seconds  
unspecified means 300 seconds  
anything else means itself rounded up modulo 5 seconds.

I haven't used GLib before, but it seems like by handling the error we can have `ShutdownTimeout=0` behave as documented in the config file comment.  Patch to follow.

src/pk-main.c
yump: Could you clarify "after dropping the change"? Which change, and how is the comment inaccurate without it?  @gordonmessmer's DNF plugin solves the problem of packagekit's model of the system getting out of sync with dnf, but if someone in the future comes up with some *other* reason they think packagekitd should run forever, they could easily reintroduce the resource squandering.

The magnitude of the issue is DNF-specific, but I don't think the issue itself is.  The cost of packagekitd is particularly high with the DNF backend, but with other backends it is not zero, as you said last week.

yump:
I tested [89e582e](https://github.com/PackageKit/PackageKit/pull/578/commits/89e582ec01c208646c67b06b18dcc3034c8453c1) on my machine, applied against the PackageKit in Fedora 37, and the config file seems to have the expected semantics, but someone with more C/GLib experience should verify that `g_clear_error()` is the right thing to do.

src/pk-main.c
yump: @mcatanzaro I am adding a comment that says "Before removing the default timeout" in the same commit that **adds back** the default timeout.  300s is the default timeout. Removing the default timeout is something that future maintainers should not do without great caution.

However if you misunderstood it, it is not clear, so I will word it.

src/pk-main.c
```diff
@@ -182,7 +182,16 @@ main (int argc, char *argv[])
...
-	exit_idle_time = g_key_file_get_integer (conf, "Daemon", "ShutdownTimeout", NULL);
+	exit_idle_time = g_key_file_get_integer (conf, "Daemon", "ShutdownTimeout", &error);
+	/* THIS COMMENT IS A TSUNAMI STONE
+         * The automatic shutdown timeout is necessary to avoid very high
+         * memory usage with the dnf backend.  If you want to remove this
+         * timeout, please study the git history and be sure that you are not
+         * regressing Redhat bugzilla #1354074 (again). */
+	if (error != NULL) {
+		exit_idle_time = 300;
+		g_clear_error(&error);
```
mcatanzaro, line 193: Yes, this is the right thing to do. You have to manually free the error and set it to NULL here, despite the g_autoptr, because the error is reused later on in this function and it's a bug to fail to unset the error before passing it to function that might set it. g_error_free() is not enough, so g_clear_error() is best.

However, code style fix:

```suggestion
		g_clear_error (&error);
```

src/pk-main.c
ximion: > Could you clarify "after dropping the change"? Which change, and how is the comment inaccurate without it?

With your change, the daemon will quit again. So the behavior the text speaks about is "times out now", which is not what it originally was about. With the status quo changed back again, that comment would be incredibly misleading.

> The magnitude of the issue is DNF-specific, but I don't think the issue itself is. The cost of packagekitd is particularly high with the DNF backend, but with other backends it is not zero, as you said last week.

It's pretty DNF-specific, the other backends consume memory, but nothing nearly this bad. But precisely because we want to consume as few background resources as possible, we *want* PK to time out again.

> Removing the default timeout is something that future maintainers should not do without great caution.

Honestly, I think you should either drop the previous comment, or reword it to say that the timeout should not be dropped because backends may leak resources.

src/pk-main.c
```diff
@@ -182,7 +182,16 @@ main (int argc, char *argv[])
 	syslog (LOG_DAEMON | LOG_DEBUG, "daemon start");
 
 	/* after how long do we timeout? */
-	exit_idle_time = g_key_file_get_integer (conf, "Daemon", "ShutdownTimeout", NULL);
+	exit_idle_time = g_key_file_get_integer (conf, "Daemon", "ShutdownTimeout", &error);
+	/* THIS COMMENT IS A TSUNAMI STONE
+         * The automatic shutdown timeout is necessary to avoid very high
```
ximion, line 187: Technically this is band-aid... Why not
```suggestion
	 * The automatic shutdown is used to prevent memory leaks in some backends to get out of hand. If you want to remove the timeout, please ensure you are not regressing Red Hat bugzilla #1354074.
```

src/pk-main.c
```diff
@@ -182,7 +182,16 @@ main (int argc, char *argv[])
...
+         * memory usage with the dnf backend.  If you want to remove this
```
ximion, line 188: style: Please use tabs and a single space, not just spaces.

ximion (changes requested):
Just a few tiny style things, then we can merge this!

ximion:
> someone with more C/GLib experience should verify that `g_clear_error()` is the right thing to do.

Yes, doing this is just fine here :-)

src/pk-main.c
```diff
@@ -182,7 +182,16 @@ main (int argc, char *argv[])
...
+         * timeout, please study the git history and be sure that you are not
+         * regressing Redhat bugzilla #1354074 (again). */
+	if (error != NULL) {
+		exit_idle_time = 300;
+		g_clear_error (&error);
+	}
 	g_debug ("daemon shutdown set to %i seconds", exit_idle_time);
```
ximion, line 195:
```suggestion
	if (exit_idle_time > 0)
		g_debug ("daemon shutdown set to %i seconds", exit_idle_time);
	else
		g_debug ("daemon will not shut down on idle");
```

src/pk-main.c
yump: Done, after a brief detour to set up a good `listchars` in my vimrc.

src/pk-main.c
yump: I removed the implication that the timeout was only relevant to dnf, but kept the slightly salty "(again)" and the suggestion to go digging past the first git blame.  I want anyone who comes along, thinking to remove the timeout, to understand that the exact thing that they want to do has caused problems more than once.

ximion (changes requested):
Some of the tiniest nitpicks imaginable, then this is good to go :-)

ximion:
Thank you for the patch! ;-)
I took a while, but I think with the DNF backend change we have a very good solution now.

PR #579, opened 2022-09-21
Update Hungarian translation

urbalazs:
This file was copied from Transifex. Please update the language files and pull the other translations from Transifex.


hughsie:
I synced Transifex manually just now, thanks.

PR #582, opened 2022-10-21
pk-engine: Fix checking D-Bus caller details when setting proxy

pwithnall:
This was broken in commit c8b531.

This should fix errors like the following when calling `SetProxy()`:
```
PackageKit-WARNING **: 14:50:25.673: failed to set proxy: GDBus.Error:org.gtk.GDBus.UnmappedGError.Quark._pk_2dengine_2derror_2dquark.Code3: setting the proxy failed: failed to get the uid
```

### DIFF
--- a/src/pk-engine.c
+++ b/src/pk-engine.c
@@ -450,6 +450,14 @@ pk_engine_set_proxy_internal (PkEngine *engine, const gchar *sender,
 	guint uid;
 	g_autofree gchar *session = NULL;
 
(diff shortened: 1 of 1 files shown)


pwithnall:
The errors are caused by `dbus->priv->proxy_uid` being `NULL` if `pk_dbus_get_uid()` is called before `pk_dbus_connect()` is called.

It’s causing the gnome-software PackageKit unit tests to fail. They work when running against a `packagekitd` with this PR applied.

hughsie (approved):

Issue #586, opened 2022-11-17
packagekitd spins on CPU forever after boot (stack trace included)


mttkay:
I'm on Fedora 36:

```shell
Linux carbon-x1 6.0.7-200.fc36.x86_64 #1 SMP PREEMPT_DYNAMIC Thu Nov 3 17:09:02 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

Recently, every time I boot into GNOME, `packagekitd` will start spinning on CPU and hog upwards of 3GB of RSS. I noticed because my laptop fans started blowing and wouldn't turn off.

It does not even respond to `TERM` when this happens; I eventually killed it via `QUIT` to dump core. This is what I saw happening at the time:

```
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
Core was generated by `/usr/libexec/packagekitd'.
Program terminated with signal SIGQUIT, Quit.
#0  0x00007efec3f0848a in __GI_fdatasync (fd=11) at ../sysdeps/unix/sysv/linux/fdatasync.c:28
28	  return SYSCALL_CANCEL (fdatasync, fd);
[Current thread is 1 (Thread 0x7efec3a33880 (LWP 3301))]
(gdb) info threads
  Id   Target Id                        Frame 
* 1    Thread 0x7efec3a33880 (LWP 3301) 0x00007efec3f0848a in __GI_fdatasync (fd=11) at ../sysdeps/unix/sysv/linux/fdatasync.c:28
  2    Thread 0x7efeb63ff640 (LWP 3310) 0x00007efec3f0590f in __GI___poll (fds=0x55ffc6e36320, nfds=2, timeout=100) at ../sysdeps/unix/sysv/linux/poll.c:29
  3    Thread 0x7efea748f640 (LWP 9992) syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
  4    Thread 0x7efeb5bfe640 (LWP 3312) 0x00007efec3f0590f in __GI___poll (fds=0x55ffc6e4b8c0, nfds=2, timeout=-1) at ../sysdeps/unix/sysv/linux/poll.c:29
```

And looking at what Thread 1 was doing at the time:

```
(gdb) info stack
#0  0x00007efec3f0848a in __GI_fdatasync (fd=11) at ../sysdeps/unix/sysv/linux/fdatasync.c:28
[...]

PR #587, opened 2022-11-25
Add client PID as transaction property and store its cmdline in db again

ximion:
Hi!
I was investigating https://github.com/PackageKit/PackageKit/issues/586 and thought that it would be incredibly useful for `pkmon` to show who has created a transaction. I also remembered that we had a feature once to add the client command-line to the transactions database, but I couldn't find that anywhere in the DB. So I looked into it and in commit https://github.com/PackageKit/PackageKit/commit/9912d52ba7c87bfb5a992e1c4914c188dc546c22 the cmdline feature was limited to only a few roles in the name of performance - however, at the time this code is run, the transaction role is UNKNOWN, so the cmdline will never be set and the feature will never be used.

So, I restored support for the lost functionality, and also added a new transaction property, `ClientPid` (calling it just `Pid` felt odd, as it's not the PID of the trnsaction or PK...) for transactions to provide the process ID of the process that created the transaction.

The `pkmon` tool reads this, determines the command-line and displays that to the user as debugging-aid.
By using the "new" `GetConnectionCredentials` call on D-Bus when available, we only need to make one call to the caller PID and UID, so in my tests I saw no impact on performance at all.

I wonder if providing the PID this way is safe, as in theory if you look up the PID too late of a short-running client process it might have been reused. But I couldn't actually come up with any scenario where this would actually be an issue, especially since this is not a security feature but more meant as debugging help. So no need to a pidfd etc. here.

What do you think? (mostly @hughsie but feedback from @pwithnall might also be nice!)

### DIFF
--- a/client/pk-monitor.c
+++ b/client/pk-monitor.c
@@ -128,6 +128,44 @@ pk_monitor_adopt_cb (PkClient *_client, GAsyncResult *res, gpointer user_data)
 	}
 }
 
+static gchar*
+pk_monitor_get_caller_info (GDBusProxy *bus_proxy, const gchar *bus_name)
+{
+	gboolean ret;
+	gchar *cmdline = NULL;
+	g_autofree gchar *filename = NULL;
+	g_autoptr(GError) error = NULL;
+	guint pid = G_MAXUINT;
+	g_autoptr(GVariant) value = NULL;
+
+	/* get pid from DBus */
+	value = g_dbus_proxy_call_sync (bus_proxy,
+					"GetConnectionUnixProcessID",
+					g_variant_new ("(s)",
+						       bus_name),
+					G_DBUS_CALL_FLAGS_NONE,
+					2000,
+					NULL,
+					&error);
+	if (value == NULL) {
+		g_warning ("Failed to get pid for %s: %s",
+			   bus_name, error->message);
+		return g_strdup_printf ("bus:%s", bus_name);
+	}
+	g_variant_get (value, "(u)", &pid);
+
+	/* get command line from proc */
+	filename = g_strdup_printf ("/proc/%i/cmdline", pid);
+	ret = g_file_get_contents (filename, &cmdline, NULL, &error);
+	if (!ret) {
+		/* we failed to get the command-line, maybe we don't have permission */
+		return g_strdup_printf ("pid:%i", pid);
+	}
+	/* the cmdline has its args nul-separated. We deliberately make use of this to only return
+	 * argv[0], i.e. the executable name */
+	return cmdline;
+}
+
 static void
 pk_monitor_progress_cb (PkProgress *progress, PkProgressType type, gpointer user_data)
 {
@@ -140,8 +178,10 @@ pk_monitor_progress_cb (PkProgress *progress, PkProgressType type, gpointer user
 	g_autofree gchar *package_id_tmp = NULL;
 	g_autofree gchar *summary = NULL;
 	g_autofree gchar *transaction_id = NULL;
+	g_autofree gchar *sender = NULL;
 	g_autoptr(PkItemProgress) item_progress = NULL;
 	g_autoptr(PkPackage) package = NULL;
+	GDBusProxy *bus_proxy = G_DBUS_PROXY (user_data);
 
 	/* get data */
 	g_object_get (progress,
@@ -153,6 +193,7 @@ pk_monitor_progress_cb (PkProgress *progress, PkProgressType type, gpointer user
 		      "item-progress", &item_progress,
 		      "package-id", &package_id,
 		      "transaction-id", &transaction_id,
+		      "sender", &sender,
 		      NULL);
 
 	/* don't print before we have properties */
@@ -186,6 +227,9 @@ pk_monitor_progress_cb (PkProgress *progress, PkProgressType type, gpointer user
 			 pk_item_progress_get_package_id (item_progress),
 			 pk_item_progress_get_percentage (item_progress),
 			 pk_status_enum_to_string (pk_item_progress_get_status (item_progress)));
+	} else if (type == PK_PROGRESS_TYPE_SENDER) {
+		g_autofree gchar *cmdline = pk_monitor_get_caller_info (bus_proxy, sender);
+		g_print ("%s\tsender       %s\n", transaction_id, cmdline);
 	}
 }
 
@@ -271,6 +315,9 @@ main (int argc, char *argv[])
 	guint i;
 	g_autoptr(PkControl) control = NULL;
 	g_autoptr(PkTransactionList) tlist = NULL;
+	g_autoptr(GDBusConnection) bus_conn = NULL;
+	g_autoptr(GDBusProxy) bus_proxy = NULL;
+	g_autoptr(GError) error = NULL;
 
 	const GOptionEntry options[] = {
 		{ "version", '\0', 0, G_OPTION_ARG_NONE, &program_version,
@@ -296,6 +343,28 @@ main (int argc, char *argv[])
 		goto out;
 	}
 
+	/* use the bus to resolve connection names to PIDs */
+	bus_conn = g_bus_get_sync (G_BUS_TYPE_SYSTEM, NULL, &error);
+	if (bus_conn == NULL) {
+		g_printerr ("Cannot connect to the system bus: %s\n", error->message);
+		retval = EXIT_FAILURE;
+		goto out;
+	}
+	bus_proxy = g_dbus_proxy_new_sync (bus_conn,
+				       G_DBUS_PROXY_FLAGS_DO_NOT_LOAD_PROPERTIES |
+				       G_DBUS_PROXY_FLAGS_DO_NOT_CONNECT_SIGNALS,
+				       NULL,
+				       "org.freedesktop.DBus",
+				       "/org/freedesktop/DBus/Bus",
+				       "org.freedesktop.DBus",
+				       NULL,
+				       &error);
+	if (bus_proxy == NULL) {
+		g_printerr ("Cannot connect to D-Bus: %s\n", error->message);
+		retval = EXIT_FAILURE;
+		goto out;
+	}
+
 	loop = g_main_loop_new (NULL, FALSE);
 
 	control = pk_control_new ();
@@ -316,7 +385,7 @@ main (int argc, char *argv[])
 
 	tlist = pk_transaction_list_new ();
 	g_signal_connect (tlist, "added",
-			  G_CALLBACK (pk_monitor_transaction_list_added_cb), NULL);
+			  G_CALLBACK (pk_monitor_transaction_list_added_cb), bus_proxy);
 	g_signal_connect (tlist, "removed",
 			  G_CALLBACK (pk_monitor_transaction_list_removed_cb), NULL);
 
--- a/lib/packagekit-glib2/pk-progress.c
+++ b/lib/packagekit-glib2/pk-progress.c
@@ -57,6 +57,7 @@ struct _PkProgressPrivate
 	guint64				 download_size_remaining;
 	guint64				 transaction_flags;
 	guint				 uid;
+	gchar				*sender;
 	PkItemProgress			*item_progress;
 	PkPackage			*package;
 };
@@ -76,6 +77,7 @@ enum {
 	PROP_DOWNLOAD_SIZE_REMAINING,
 	PROP_TRANSACTION_FLAGS,
 	PROP_UID,
+	PROP_SENDER,
 	PROP_PACKAGE,
 	PROP_ITEM_PROGRESS,
 	PROP_LAST
@@ -133,6 +135,9 @@ pk_progress_get_property (GObject *object, guint prop_id, GValue *value, GParamS
 		break;
 	case PROP_UID:
 		g_value_set_uint (value, progress->priv->uid);
+		break;
+	case PROP_SENDER:
+		g_value_set_string (value, progress->priv->sender);
 		break;
 	case PROP_PACKAGE:
 		g_value_set_object (value, progress->priv->package);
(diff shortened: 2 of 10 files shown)


aleixpol:
You should already have the caller pid, we don't need clients to be providing their own.

ximion:
> You should already have the caller pid, we don't need clients to be providing their own.

They don't do that with this patch, quite the opposite: The daemon provides the PID of the caller *back*, so that monitoring tools can fetch it and determine who triggered a transaction. Clients don't need to do anything, this patch purely fixes an old issue (command-line of the caller missing from the transactions database) and adds a new read-only property for monitoring tools to determine the PID of a client that belongs to a transaction.

pwithnall:
PIDs are racy. You should probably either use a pidfd or the unique D-Bus name for each client.

ximion:
> PIDs are racy. You should probably either use a pidfd or the unique D-Bus name for each client. The latter would be a bit simpler.

Exposing the D-Bus sender name would also mean that we would only need to resolve the name to a PID when needed, and wouldn't do that unconditionally. I just don't know how stable the bus sender names are (but when the D-Bus name is invalid, the transaction is invalid too, while the transaction might outlive the process that created it).
I've altered the patch to provide the bus name instead, that seems like a much better idea!

In my limited testing, I've seen no issues with this approach, everything still works as expected with the same performance. I've also split the one large patch into two, so it's a bit easier to review and reason about.

ximion:
Oh, and for the bugfix part of this patch, when the transaction sender is set by the scheduler, we still determine the caller PID immediately together with its UID, ideally using `GetConnectionCredentials` if we need both, and resolve to a command-line at a later stage in `pk_transaction_set_state` - I think this is fine, because when a transaction role is set, the sender is most likely still alive and the PID hasn't been reused.
Alternatively we could also resolve the D-Bus name to a PID on-demand in `pk_transaction_set_state`, but then we would need to make two D-Bus calls, one to get the UID immediately, and one later to get the PID.

Of course, you can also make an argument that we should maybe drop this feature, as it has been broken for 6 years and nobody noticed.... But I think recording the sender in the transactions database is still pretty useful and it's likely worth restoring (and if it turns out to be a major pain for some reason, we can still always drop it later).

src/pk-dbus.c
```diff
@@ -45,6 +45,66 @@ static gpointer pk_dbus_object = NULL;
...
+					g_variant_new ("(s)",
+						       sender),
+					G_DBUS_CALL_FLAGS_NONE,
+					2000,
+					NULL,
+					&error);
+	if (reply_var == NULL) {
+		g_warning ("Failed to get uid/pid for %s: %s",
+			   sender, error->message);
+		return G_MAXUINT;
```
pwithnall, line 87: The return type is boolean, and `G_MAXUINT` isn’t a boolean

src/pk-dbus.c
```diff
@@ -45,6 +45,66 @@ static gpointer pk_dbus_object = NULL;
...
+	}
+
+	have_pid = pid == NULL;
+	have_uid = uid == NULL;
```
pwithnall, line 91: It seems confusing to call a variable `have_pid` and then assign it to true iff `pid` is a `NULL` pointer.

src/pk-dbus.c
```diff
@@ -45,6 +45,66 @@ static gpointer pk_dbus_object = NULL;
...
+	g_variant_get (reply_var, "(a{sv})", &iter);
+	while (g_variant_iter_loop (iter, "{&sv}", &key, &value)) {
+		if (!have_pid && g_strcmp0 (key, "ProcessID") == 0)
+			*pid = g_variant_get_uint32 (value);
+		else if (!have_uid && g_strcmp0 (key, "UnixUserID") == 0)
+			*uid = g_variant_get_uint32 (value);
+
+		if (have_pid && have_uid) {
```
pwithnall, line 99: `have_pid` and `have_uid` aren’t updated inside the loop body, so this branch is not doing much

src/pk-shared.c
```diff
@@ -452,3 +452,28 @@ pk_string_replace (GString *string, const gchar *search, const gchar *replace)
...
 }
+
+/**
+ * pk_get_cmdline_for_pid:
+ * @dbus: the #PkDbus instance
+ * @pid: the process ID to get the command line for.
+ *
+ * Gets the command line for the given PID.
+ *
+ * Return value: the cmdline, or %NULL if it could not be obtained
```
pwithnall, line 463:
```suggestion
 * Return value: (nullable): the cmdline, or %NULL if it could not be obtained
```

src/pk-shared.c
```diff
@@ -452,3 +452,28 @@ pk_string_replace (GString *string, const gchar *search, const gchar *replace)
 out:
 	return count;
 }
+
+/**
+ * pk_get_cmdline_for_pid:
+ * @dbus: the #PkDbus instance
+ * @pid: the process ID to get the command line for.
+ *
+ * Gets the command line for the given PID.
```
pwithnall, line 461: Would be worthwhile mentioning that this is racy if the PID is reused while it’s called

src/pk-transaction.c
```diff
@@ -94,7 +97,8 @@ struct PkTransactionPrivate
 	gboolean		 emit_media_change_required;
 	gboolean		 caller_active;
 	gboolean		 exclusive;
-	guint			 uid;
+	guint32			 uid;
```
pwithnall, line 100: Would it make sense to rename this to `client_uid` while you’re here?

client/pk-monitor.c
```diff
@@ -128,6 +128,42 @@ pk_monitor_adopt_cb (PkClient *_client, GAsyncResult *res, gpointer user_data)
 	}
 }
 
+static gchar*
+pk_monitor_get_caller_info (GDBusProxy *bus_proxy, const gchar *name)
```
pwithnall, line 132: Might be better to call the argument `bus_name` rather than just `name` to make it a bit clearer it’s a unique or well-known D-Bus name.

client/pk-monitor.c
```diff
@@ -128,6 +128,42 @@ pk_monitor_adopt_cb (PkClient *_client, GAsyncResult *res, gpointer user_data)
...
+					g_variant_new ("(s)",
+						       name),
+					G_DBUS_CALL_FLAGS_NONE,
+					2000,
+					NULL,
+					&error);
+	if (value == NULL) {
+		g_warning ("Failed to get pid for %s: %s",
+			   name, error->message);
+		return g_strdup_printf ("bus:%s", name);;
```
pwithnall, line 153:
```suggestion
		return g_strdup_printf ("bus:%s", name);
```

(accidental double semicolon)

client/pk-monitor.c
```diff
@@ -128,6 +128,42 @@ pk_monitor_adopt_cb (PkClient *_client, GAsyncResult *res, gpointer user_data)
...
+	g_variant_get (value, "(u)", &pid);
+
+	/* get command line from proc */
+	filename = g_strdup_printf ("/proc/%i/cmdline", pid);
+	ret = g_file_get_contents (filename, &cmdline, NULL, &error);
+	if (!ret) {
+		/* we failed to get the command-line, maybe we don't have permission */
+		return g_strdup_printf ("pid:%i", pid);
+	}
+	return cmdline;
```
pwithnall, line 164: Returning `/proc/*/cmdline` directly will give a string with nul bytes between the arguments. i.e. It’ll appear to only be the `argv[0]`? Is that what you want? If so, it might clarify things to add a comment about that decision.

Otherwise, you’ll have to post-process the `cmdline` to replace `\0` with ` ` up to the file length.

client/pk-monitor.c
```diff
@@ -296,6 +341,28 @@ main (int argc, char *argv[])
 		goto out;
 	}
 
+	/* use the bus to resolve connetcion names to PIDs */
+	bus_conn = g_bus_get_sync (G_BUS_TYPE_SYSTEM, NULL, &error);
+	if (bus_conn == NULL) {
+		g_printerr ("Cannot connect to the system bus: %s", error->message);
```
pwithnall, line 347:
```suggestion
		g_printerr ("Cannot connect to the system bus: %s\n", error->message);
```

`g_printerr()` doesn’t automatically add a trailing newline.

client/pk-monitor.c
```diff
@@ -296,6 +341,28 @@ main (int argc, char *argv[])
 		goto out;
 	}
 
+	/* use the bus to resolve connetcion names to PIDs */
```
pwithnall, line 344:
```suggestion
	/* use the bus to resolve connection names to PIDs */
```

client/pk-monitor.c
```diff
@@ -296,6 +341,28 @@ main (int argc, char *argv[])
...
+				       G_DBUS_PROXY_FLAGS_DO_NOT_LOAD_PROPERTIES |
+				       G_DBUS_PROXY_FLAGS_DO_NOT_CONNECT_SIGNALS,
+				       NULL,
+				       "org.freedesktop.DBus",
+				       "/org/freedesktop/DBus/Bus",
+				       "org.freedesktop.DBus",
+				       NULL,
+				       &error);
+	if (bus_proxy == NULL) {
+		g_printerr ("Cannot connect to D-Bus: %s", error->message);
```
pwithnall, line 361:
```suggestion
		g_printerr ("Cannot connect to D-Bus: %s\n", error->message);
```

lib/packagekit-glib2/pk-progress.c
```diff
@@ -774,6 +779,51 @@ pk_progress_get_uid (PkProgress *progress)
...
+ * @name: a D-Bus name
+ *
+ * Set the D-Bus name of the client that started this transaction.
+ *
+ * Return value: %TRUE if value changed.
+ *
+ * Since: 1.2.6
+ **/
+gboolean
+pk_progress_set_sender (PkProgress *progress, const gchar *name)
```
pwithnall, line 794:
```suggestion
pk_progress_set_sender (PkProgress *progress, const gchar *bus_name)
```

lib/packagekit-glib2/pk-progress.c
```diff
@@ -774,6 +779,51 @@ pk_progress_get_uid (PkProgress *progress)
...
+	return TRUE;
+}
+
+/**
+ * pk_progress_get_sender:
+ * @progress: a valid #PkProgress instance
+ *
+ * Get the D-Bus name of the client that started this transaction.
+ *
+ * Return value: a unique D-Bus name
```
pwithnall, line 816: Probably safest to just call it a “D-Bus name”, as it could theoretically be a well-known _or_ a unique one. Although in practice it’ll always be a unique one

lib/packagekit-glib2/pk-progress.c
```diff
@@ -1044,6 +1094,18 @@ pk_progress_class_init (PkProgressClass *klass)
...
+	 * PkProgress:sender:
+	 *
+         * The D-Bus name of the client that started this transaction.
+         *
+	 * Since: 1.2.6
+	 */
+	pspec = g_param_spec_string ("sender", NULL, NULL,
+				     NULL,
+				     G_PARAM_READWRITE);
+	g_object_class_install_property (object_class, PROP_SENDER, pspec);
```
pwithnall, line 1107: Looks like these properties could be improved to be installed all at once and have more flags set, just as in https://github.com/ximion/appstream/pull/448. But that’s something for a separate PR. This change looks good in the context of the existing code.

lib/packagekit-glib2/pk-progress.h
```diff
@@ -106,6 +107,7 @@ typedef enum {
 	PK_PROGRESS_TYPE_PACKAGE,
 	PK_PROGRESS_TYPE_ITEM_PROGRESS,
 	PK_PROGRESS_TYPE_TRANSACTION_FLAGS,
+	PK_PROGRESS_TYPE_SENDER,
```
pwithnall, line 110: I think this will need to be inserted after `PK_PROGRESS_TYPE_INVALID` in the enum, or using a new PackageKit library with a client program which was compiled against an older version will result in any progress emissions of type `PK_PROGRESS_TYPE_SENDER` being interpreted by the client as `INVALID`.

src/pk-dbus.c
ximion: Ooof... Too bad the compiler wasn't smart enough to warn about this (and gboolean isn't bool...).

lib/packagekit-glib2/pk-progress.c
ximion: Yeah, I was actually thinking about that while writing it... But modernizing PK is a pretty huge task (and having this consistent for now felt nicer).
The AppStream bits were most likely actually inspired by PKs code, as I was working on both a lot when AppStream was created ;-)

lib/packagekit-glib2/pk-progress.h
ximion: Hmm... I wonder whether that would cause problems if SENDER was considered invalid for older clients... But it's indeed better to be conservative here, good catch! I think while modifying it I didn't see this as public API, which was a bad call.

src/pk-transaction.c
ximion: Probably... Just UID is actually a bit confusing for people new to the code, but I probably just got used to it.

ximion (commented):
Thank you for the review!
I'm not impressed with the new code in pkmon, but it's likely fine for a simple tool like this and doesn't need to be overcomplicated :-)

pwithnall (commented):
A few remaining comments while I skimmed through this, but it looks good to me (for what that’s worth).

ximion:
> but it looks good to me (for what that’s worth)

It's worth a lot! This started out as a quick patchset that I was uncertain about, and ended up as a much improved change (using bus names is much better!) that I feel comfortable merging!
So, thank you a lot for the review! PK 1.2.6 will be a very exciting release!

PR #591, opened 2022-12-30
Always check for invalidated dnf sacks and free them.

gordonmessmer:
While looking at packagekitd memory use on Fedora, I noticed that invalid dnf sacks are only freed if create_flags & DNF_CREATE_SACK_FLAG_USE_CACHE, but there are conditions in which that flag may be masked away.  If I'm not mistaken, this can prevent sacks from ever being freed and leads to indefinite memory growth.


hughsie:
Hmm; I think that was deliberate. Can you explain where we're leaking the non-cached sack? It's been many months since I looked at any of that code :)

gordonmessmer:
Hm... no, actually.  In part, I was mislead by the comment on lines 738-739, and I must've made a mistake in gdb while trying to confirm that "relying on the callback of the hash table" caused some kind of problem.

I can't make any sense of the else block on lines 737-741.  *shrug*

PR #592, opened 2022-12-30
Run pk_backend_refresh_repo in a new process

gordonmessmer:
Run pk_backend_refresh_repo in a new process to avoid keeping libdnf memory leaks.

This is not a real PR, I do not expect this to be merged.  I understand that running complex code in a process forked from a threaded process is not safe.

This PR is instead intended as a suggestion / request to move pk_backend_refresh_repo into a new helper application, and to fork&&exec that helper from pk_backend_refresh_cache_thread.

While reviewing packagekitd memory use on Fedora, I saw constant growth of memory use every time repo metadata was refreshed (easily demonstrated by running `pkcon refresh force` repeatedly.  Running packagekitd under valgrind appeared to locate numerous memory leaks, most of which were under libdnf's dnf_repo_update().  The number of unique leaks detected makes fixing them in the short term probably infeasible.  However, running the update in a new process means that at least leaks get collected when the process exits.  That does mean that error information from pk_backend_refresh_repo will be lost, but that seems like a sensible trade for now.

Whereas before this change, memory use grew each time the repo was refreshed, after this change memory use settled at ~ 200MB.  (Which is still kind of large, but significantly smaller than without it.)

Would you consider creating a new helper binary to perform repo refresh, and run that helper in a new process, rather than running pk_backend_refresh_repo in the packagekitd process?

### DIFF
--- a/backends/dnf/pk-backend-dnf.c
+++ b/backends/dnf/pk-backend-dnf.c
@@ -45,6 +45,7 @@
 
 #include "dnf-backend-vendor.h"
 #include "dnf-backend.h"
+#include "pk-backend-dnf-common.h"
 
 typedef struct {
 	DnfSack		*sack;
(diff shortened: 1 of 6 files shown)


hughsie:
I think the medium-term plan is to use dnf-daemon -- which hopefully is going to be designed so that the dnf context can be used again and again.

gordonmessmer:
What is "medium term"?  And whose plan are we discussing?  The last time I asked the Fedora maintainers about the dnf daemon, there were no plans to make it a mandatory component.

Right now, I think I see two opportunities to significantly improve PackageKit on Fedora.  The first is this one, which I have a high degree of confidence in.  If pk_backend_refresh_repo is moved to a new executable, and that executable is run in a new process, then memory use in packagekitd doesn't grow every time metadata is refreshed.  The second one, which I haven't started yet, is putting an expiration timer on the dnf sacks in sack_cache and removing them periodically.

I'm interested in continuing work on those if you're interested in merging them.

hughsie:
> What is "medium term"? And whose plan are we discussing?

Ha, I'm not sure. I did hope it was going to be implemented *last year* but that obviously didn't happen.

>  If pk_backend_refresh_repo is moved to a new executable

There are lifetime issues with this, and the cache invalidation stuff won't know about things happening in other processes.

>  The second one, which I haven't started yet, is putting an expiration timer on the dnf sacks in sack_cache

I like this one a lot more.

gordonmessmer:
> There are lifetime issues with this, and the cache invalidation stuff won't know about things happening in other processes.

Cache invalidation still happens in pk_backend_refresh_cache_thread, after synchronous execution of the child process that refreshes metadata.  That isn't changed in this proposal.

> I like this one a lot more.

I'm hopeful that it could save a lot of memory, and I'd like to explore it.  But valgrind seems to indicate that libdnf and librepo leak memory during the metadata refresh, and that's totally separate from the dnf sacks.  Fixing that probably means either extensive work in libdnf and librepo to fix leaks, or punting and doing the refresh in a separate process.  The latter is less elegant, but more effective.

gordonmessmer:
I've pushed new changes that implement package refresh in an external process.  This version should be very nearly ready to merge.  

Testing indicates that after running `pkcon refresh force` five times, packagekitd's RSS was just under 200MB, whereas without this change RSS was just under 400MB after five refreshes.

mcatanzaro:
> Testing indicates that after running pkcon refresh force five times, packagekitd's RSS was just under 200MB, whereas without this change RSS was just under 400MB after five refreshes.

I know that using a subprocess as a workaround for memory leaks is pretty hacky, but still... _nice_. That's too big to ignore.

gordonmessmer:
I'm following up on PRs today.  Please let me know if there are any issues that still need to be addressed with this PR.

Thank you.

backends/dnf/pk-backend-dnf.c
```diff
@@ -1815,12 +1671,22 @@ pk_backend_refresh_cache_thread (PkBackendJob *job,
 		}
 
 		/* check and download */
-		state_loop = dnf_state_get_child (state_local);
-		ret = pk_backend_refresh_repo (job, repo, state_loop, &error);
-		if (!ret) {
-			pk_backend_job_error_code (job, error->code, "%s", error->message);
-			return;
-		}
+		refresh[0] = LIBEXECDIR "/packagekit-refresh-repo";
```
Conan-Kudo, line 1674: If this is DNF specific, we probably should name the binary accordingly. `packagekit-dnf-refresh-repos` is probably a better name here.

Conan-Kudo (changes requested):

backends/dnf/pk-backend-dnf.c
gordonmessmer: Done.

Conan-Kudo:
Can you clean up the commits for mergeability? It's a bit messy right now...

gordonmessmer:
Squashed and rebased on the current main.

mcatanzaro:
So... this is ready to merge, then?

ximion:
Is this workaround still needed, with the autoshutdown patches available and about to land soonish, as well as the sack timeout stuff?

gordonmessmer:
> So... this is ready to merge, then?

As far as I know: yes.  I just applied this patch on Fedora's rawhide package and tested the resulting build.  Everything works as expected, here.

gordonmessmer:
> Is this workaround still needed, with the autoshutdown patches available and about to land soonish, as well as the sack timeout stuff?

Autoshutdown does make it less critical to handle memory leaks, but on the other hand, autoshutdown has been rolled back before, so my own preference is for making packagekitd more resilient against leaks in the configuration without shutdown on idle.  Without shutdown on idle, both this patch and the sack expiration patches are still relevant.

Conan-Kudo:
With no objections for about a month, I'm merging this.

PR #597, opened 2023-01-14
 Free DNF sacks when they are unused for 10 minutes. 

gordonmessmer:


gordonmessmer:
This works more or less as expected, but the RSS for packagekitd doesn't shrink as much as I think it should.  I'll run the result through valgrind to look for remaining leaks.  If I don't find anything interesting, it's possible that some of the remaining un-returned use is simply the result of memory fragmentation, which will be harder to solve.

backends/dnf/pk-backend-dnf.c
```diff
@@ -59,6 +61,7 @@ typedef struct {
 	GMutex		 sack_mutex;
 	GTimer		*repos_timer;
 	gchar		*release_ver;
+	guint		sack_expire_source;
```
hughsie, line 64: this should be `sack_expire_id`

backends/dnf/pk-backend-dnf.c
```diff
@@ -330,6 +351,7 @@ pk_backend_destroy (PkBackend *backend)
 		g_key_file_unref (priv->conf);
 	if (priv->context != NULL)
 		g_object_unref (priv->context);
+	g_source_remove (priv->sack_expire_source);
```
hughsie, line 354: i think you need to check if it's `!= 0` first.

backends/dnf/pk-backend-dnf.c
```diff
@@ -46,10 +46,12 @@
 #include "dnf-backend-vendor.h"
 #include "dnf-backend.h"
 
+#define SACK_TTL	600
```
hughsie, line 49: maybe `DNF_SACK_MAX_AGE 600 /* seconds */` or something?

backends/dnf/pk-backend-dnf.c
hughsie: also needs one space indent

backends/dnf/pk-backend-dnf.c
```diff
@@ -94,24 +97,37 @@ pk_backend_supports_parallelization (PkBackend *backend)
...
-	g_autoptr(GList) values = NULL;
 	g_autoptr(GMutexLocker) locker = g_mutex_locker_new (&priv->sack_mutex);
 
-	/* set all the cached sacks as invalid */
-	values = g_hash_table_get_values (priv->sack_cache);
-	for (l = values; l != NULL; l = l->next) {
-		cache_item = l->data;
-		if (cache_item->valid) {
-			g_debug ("invalidating %s as %s", cache_item->key, why);
-			cache_item->valid = FALSE;
```
hughsie, line 112: why did we have this boolean? i'm guessing if it's all single threaded it's safe to just remove them here.

backends/dnf/pk-backend-dnf.c
```diff
@@ -94,24 +97,37 @@ pk_backend_supports_parallelization (PkBackend *backend)
...
+}
+
 static void
 pk_backend_sack_cache_invalidate (PkBackend *backend, const gchar *why)
 {
-	GList *l;
-	DnfSackCacheItem *cache_item;
 	PkBackendDnfPrivate *priv = pk_backend_get_user_data (backend);
-	g_autoptr(GList) values = NULL;
 	g_autoptr(GMutexLocker) locker = g_mutex_locker_new (&priv->sack_mutex);
```
hughsie, line 126: does this suggest that it's not always single threaded? I can't remember the details.

backends/dnf/pk-backend-dnf.c
```diff
@@ -318,6 +335,10 @@ pk_backend_initialize (GKeyFile *conf, PkBackend *backend)
 						  g_free,
 						  (GDestroyNotify) dnf_sack_cache_item_free);
 
+	priv->sack_expire_source = g_timeout_add_seconds (SACK_TTL / 2,
```
hughsie, line 338: /2?

backends/dnf/pk-backend-dnf.c
```diff
@@ -804,7 +821,7 @@ dnf_utils_create_sack_for_filters (PkBackendJob *job,
 	cache_item = g_slice_new (DnfSackCacheItem);
 	cache_item->key = g_strdup (cache_key);
 	cache_item->sack = g_object_ref (sack);
-	cache_item->valid = TRUE;
+	cache_item->timer = g_timer_new();
```
hughsie, line 824: needs the space before the `(`

backends/dnf/pk-backend-dnf.c
gordonmessmer: I don't know.  There doesn't seem to be any commit history to suggest why it was handled this way.

I don't think this is single-threaded, but the sack caches are ref counted, so if one of them had been handed to a thread, then the cache item should be freed in pk_backend_sack_cache_invalidate, and the sack cache should be freed when the thread was done with it.   If no thread has a reference to the sack cache, then both cache item and sack cache should be freed in pk_backend_sack_cache_invalidate.

backends/dnf/pk-backend-dnf.c
gordonmessmer: The timeout and the sack cache expiration aren't necessarily going to align well, so if the check is done at the same interval as the TTL, then some sacks will probably live close to 2x the TTL.  Checking at TTL/2 means that the real max age is somewhere around 1.5x TTL.  Smaller intervals mean more frequent wakeups, so I tend to prefer this level of accuracy.

mcatanzaro:
Looks like this is ready for a second round of review.

backends/dnf/pk-backend-dnf.c
gordonmessmer: In case this concern is stalling review of this PR, I've built a modified version locally.  This version sets the sack TTL to just 6 seconds, and adds `g_usleep (10 * G_USEC_PER_SEC);` after every call to `dnf_utils_create_sack_for_filters`.  This should ensure that every DnfSackCacheItem is freed before the threat that requested it begins processing using the DnfSack from the cache.

I've run this modified version under valgrind, and it reports no use-after-free errors.

I've also run the modified version under gdb, and verified that when the cache item is freed, the DnfSack has a ref_count of 2, and that the DnfSack is freed in the thread where it is used.

Cache clear:

```
(gdb) p ((DnfSack *)_object)->parent_instance 
$4 = {g_type_instance = {g_class = 0x7fffd4004cb0}, ref_count = 2, qdata = 0x0}
(gdb) bt
#0  g_object_unref (_object=0x7fffd40050e0) at ../gobject/gobject.c:3785
#1  0x00007ffff73e7089 in dnf_sack_cache_item_free (cache_item=0x7fffd60199a0) at ../backends/dnf/pk-backend-dnf.c:145
#2  0x00007ffff7e37bd2 in g_hash_table_foreach_remove_or_steal (hash_table=0x5555555f0400, func=func@entry=0x7ffff73e6fd0 <pk_backend_check_sack_timer>, user_data=user_data@entry=0x0, notify=notify@entry=1)
    at ../glib/ghash.c:1981
#3  0x00007ffff7e37c20 in g_hash_table_foreach_remove (hash_table=<optimized out>, func=func@entry=0x7ffff73e6fd0 <pk_backend_check_sack_timer>, user_data=user_data@entry=0x0) at ../glib/ghash.c:2025
#4  0x00007ffff73e704b in pk_backend_sack_expire (user_data=user_data@entry=0x55555569c230) at ../backends/dnf/pk-backend-dnf.c:118
#5  0x00007ffff7e4c4c8 in g_timeout_dispatch (source=0x55555569c7e0, callback=0x7ffff73e7020 <pk_backend_sack_expire>, user_data=0x55555569c230) at ../glib/gmain.c:5007
#6  0x00007ffff7e4bcbf in g_main_dispatch (context=0x5555555bab40) at ../glib/gmain.c:3444
#7  g_main_context_dispatch (context=0x5555555bab40) at ../glib/gmain.c:4162
#8  0x00007ffff7ea1598 in g_main_context_iterate.constprop.0 (context=0x5555555bab40, block=1, dispatch=1, self=<optimized out>) at ../glib/gmain.c:4238
#9  0x00007ffff7e4b28f in g_main_loop_run (loop=0x5555555bac30) at ../glib/gmain.c:4438
#10 0x00005555555654d2 in main (argc=<optimized out>, argv=<optimized out>) at ../src/pk-main.c:245
```

Sack freed in the thread where it was used:

```
(gdb) p ((DnfSack *)_object)->parent_instance 
$5 = {g_type_instance = {g_class = 0x7fffd4004cb0}, ref_count = 1, qdata = 0x0}
(gdb) bt
#0  g_object_unref (_object=0x7fffd40050e0) at ../gobject/gobject.c:3785
#1  0x00007ffff73ea518 in glib_autoptr_clear_GObject (_ptr=0x7fffd40050e0) at /usr/include/glib-2.0/gobject/gobject-autocleanups.h:29
#2  glib_autoptr_clear_DnfSack (_ptr=0x7fffd40050e0) at /usr/include/libdnf/dnf-sack.h:37
#3  glib_autoptr_cleanup_DnfSack (_ptr=<synthetic pointer>) at /usr/include/libdnf/dnf-sack.h:37
#4  pk_backend_search_thread (job=0x555555890410, params=<optimized out>, user_data=<optimized out>) at ../backends/dnf/pk-backend-dnf.c:960
#5  0x000055555557bd1e in pk_backend_job_thread_setup (thread_data=0x555555898450) at ../src/pk-backend-job.c:724
#6  0x00007ffff7e759c2 in g_thread_proxy (data=0x5555558929e0) at ../glib/gthread.c:831
#7  0x00007ffff77f112d in start_thread (arg=<optimized out>) at pthread_create.c:442
#8  0x00007ffff7872bc0 in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:81
```

Conan-Kudo (approved):
This seems reasonable to me.

Conan-Kudo:
@gordonmessmer Can you review the remaining comments to resolve @hughsie's concerns?

gordonmessmer:
The testing that I did last week was intended to address both of the comments about thread safety.

backends/dnf/pk-backend-dnf.c
gordonmessmer: It's not single theaded, but removing a DnfSackCacheItem from sack_cache should be safe, since it only unrefs the dnf sack that the CacheItem references.  The thread that requested the dnf sack will unref and free the sack when it is done with it.

gordonmessmer:
I want to rephrase that:

If there are any concerns that have not been addressed, or any evidence that you would like to see that this change is safe, please let me know what would else would be helpful.

gordonmessmer:
Since #592 was merged: Is there anything I can do to move this PR forward, as well?

Conan-Kudo:
The lack of comments for the past few months inclines me to merge this, so I'm going to. Thanks for the patch @gordonmessmer!

ximion:
Tbh, this always felt like a weird thing to do to me and like a crude hack. But I thought it was addressed with the daemon shutting down anyway after 10min, so I was assuming we shouldn't merge this.

But I also don't have any stakes in the DNF backend, so I don't mind either way.

Conan-Kudo:
If it turns out badly, we can revert it, but I don't like leaving things in limbo if I don't have to.

gordonmessmer:
> Tbh, this always felt like a weird thing to do to me and like a crude hack. But I thought it was addressed with the daemon shutting down anyway after 10min

I do prefer the shutdown on idle behavior to always running, but this PR was opened before that one.

As far as I know, RHEL 10 will still include PackageKit, and I'd guess that it will include the shutdown on idle behavior.  But having this patch merged upstream increases the likelihood that RHEL's engineers might fix PackageKit memory use in RHEL 9 (and, I suppose as a remote possibility in RHEL 8).

Notably, one of the users affected by high memory use on RHEL 8 appears to have seen use as high as 7GB, (if I'm reading that correctly, which ... ) https://bugzilla.redhat.com/show_bug.cgi?id=1854875

I'm not sure how that happened, but in more common cases, dnf sacks are *quite* large, and IIRC, the dnf module will allocate a unique sack per function that uses one.  Keeping them briefly seems like a reasonable measure to reduce the setup cost, but keeping them forever probably isn't justifiable.

PR #599, opened 2023-01-15
valgrind warns that a conditional depends on an uninitialized value.

gordonmessmer:


ximion:
Thank you for the patch!

PR #603, opened 2023-01-30
alpm: add and support deepin group

BLumia:
This patch attempt to add support for deepin desktop environment (deepin, or DDE) when using pkcon under Arch Linux.

`PK_GROUP_ENUM_DESKTOP_DEEPIN` was added as an new enum for deepin/DDE. After compile, `./client/pkcon get-groups` should be able to list `desktop-deepin` as a group.

(but `./client/pkcon search group desktop-deepin -c -1` cannot list any package in that group even if I updated the `/etc/PackageKit/alpm.d/groups.list` file, not sure if I missed anything. How can I test it properly?)

by the way, the `groups.list` file might need an update since the group names in Arch Linux repo already changed a lot. I could submit another patch once this get merged :)


ximion:
Does Arch even have a tag for its packages in such a group? (Or even some method of desktop tagging at all, I've never noticed that on Arch...)

In any case, I think this feature of PackageKit has been superseded pretty much in its entirety by [AppStream](https://www.freedesktop.org/software/appstream/docs/), so supporting that may actually be the more useful way forward (and give you suitable metadata immediately).

BLumia:
> Does Arch even have a tag for its packages in such a group?

It has a [`groups` key in PKGBUILD](https://wiki.archlinux.org/title/PKGBUILD#groups) and you can see [a list of existing groups here](https://archlinux.org/groups/).

By looking into the current implementation, under Arch, PackageKit uses `groups.list` file to create a map to map the existing groups to PackageKit-defined groups if I understand it correctly.

hughsie:
I think categories are the logical replacements for groups -- I don't think anything uses the groups enums anymore.

BLumia:
> I think categories are the logical replacements for groups

Does it mean the group feature (`get-groups`) is deprecated and it's directly replaced by the catalog feature? I didn't find anything about its deprecation/superseding. If it's going to be dropped then it might be worth to document it in manpage and somewhere else.

Also, the `search` subcommand seemsn doesn't support to search by catalog, manpage says it only support name/details/group/file.

> I don't think anything uses the groups enums anymore.

[Apper by KDE](https://apps.kde.org/apper/) uses groups instead of catalog AFAICT.

ximion:
> Does it mean the group feature (`get-groups`) is deprecated and it's directly replaced by the catalog feature? I didn't find anything about its deprecation/superseding. If it's going to be dropped then it might be worth to document it in manpage and somewhere else.

It's not formally deprecated, it's just that there is a better way to do this now for software installer GUIs.

> [Apper by KDE](https://apps.kde.org/apper/) uses groups instead of catalog AFAICT.

Yes, because Apper is pretty much dead upstream for many years - nobody develops it anymore. 
Tools like GNOME Software or KDE Discover or AppCenter use AppStream's `as_pool_get_components_by_categories` with XDG categories. See https://www.freedesktop.org/software/appstream/docs/api/appstream-AsPool.html#as-pool-get-components-by-categories and https://specifications.freedesktop.org/menu-spec/latest/apa.html

BLumia:
So does it mean we no longer accept changes related to the `get-groups` feature? If it does then I think I can close this PR since it's not going to be merged...

BLumia:
By the way, I submitted `DDE` as a category name to upstream as you suggested: https://gitlab.freedesktop.org/xdg/xdg-specs/-/merge_requests/68 :)

docs/api/spec/pk-concepts.xml
```diff
@@ -695,6 +695,10 @@
             <entry><literal>desktop-kde</literal></entry>
             <entry>KDE</entry>
           </row>
+          <row>
+            <entry><literal>desktop-deepin</literal></entry>
+            <entry>Deepin (DDE)</entry>
```
BLumia, line 700:
```suggestion
            <entry><literal>desktop-dde</literal></entry>
            <entry>DDE</entry>
```

lib/packagekit-glib2/pk-enum.c
```diff
@@ -263,6 +263,7 @@ static const PkEnumMatch enum_group[] = {
 	{PK_GROUP_ENUM_DESKTOP_GNOME,		"desktop-gnome"},
 	{PK_GROUP_ENUM_DESKTOP_KDE,		"desktop-kde"},
 	{PK_GROUP_ENUM_DESKTOP_XFCE,		"desktop-xfce"},
+	{PK_GROUP_ENUM_DESKTOP_DEEPIN,		"desktop-deepin"},
```
BLumia, line 266:
```suggestion
	{PK_GROUP_ENUM_DESKTOP_DEEPIN,		"desktop-dde"},
```

backends/alpm/groups.list
```diff
@@ -11,6 +11,8 @@ compiz-fusion-kde	desktop-kde
 compiz-gnome		desktop-gnome
 compiz-gtk		desktop-gnome
 compiz-kde		desktop-kde
+deepin			desktop-deepin
+deepin-extra		desktop-deepin
```
BLumia, line 15:
```suggestion
deepin			desktop-dde
deepin-extra		desktop-dde
```

lib/packagekit-glib2/pk-enum.h
```diff
@@ -563,6 +564,7 @@ typedef enum {
 	PK_GROUP_ENUM_ACCESSORIES,
 	PK_GROUP_ENUM_ADMIN_TOOLS,
 	PK_GROUP_ENUM_COMMUNICATION,
+	PK_GROUP_ENUM_DESKTOP_DEEPIN,
```
BLumia, line 567:
```suggestion
	PK_GROUP_ENUM_DESKTOP_DDE,
```

lib/packagekit-glib2/pk-enum.h
```diff
@@ -523,6 +523,7 @@ typedef enum {
  * @PK_GROUP_ENUM_ACCESSORIES: Accessory packages
  * @PK_GROUP_ENUM_ADMIN_TOOLS: Administration tools packages
  * @PK_GROUP_ENUM_COMMUNICATION: Communication packages
+ * @PK_GROUP_ENUM_DESKTOP_DEEPIN: Deepin packages
```
BLumia, line 526:
```suggestion
 * @PK_GROUP_ENUM_DESKTOP_DDE: DDE packages
```

BLumia (commented):
I should use DDE instead of deepin since DDE is the name of the desktop environment, instead of deepin.

Edit: I think it might be good to document here about why we suggest to use *DDE* instead of *deepin*. *DDE* is the name of the DE, and *deepin* is the branding name of a Linux distribution. And also, when using deepin as a standalone branding name, the first letter needs to be in lower-case (use `deepin` instead of `Deepin`) if we follow the branding guideline closely. Also, the name that we suggest to use was discussed in [the deepin-devel mailing list](https://www.freelists.org/post/deepin-devel/xdgspec-DDE) and the result of the discussion was to use `DDE`. So as a result, I will use `DDE` here.

ximion:
> So does it mean we no longer accept changes related to the `get-groups` feature?

That's a good question, @hughsie what do you think? The biggest consequence I see from adding it is that every frontend that uses icons for the existing groups would have an empty "DDE" group now, with no icon and no long description until support is added to every frontend.

> By the way, I submitted `DDE` as a category name to upstream as you suggested: https://gitlab.freedesktop.org/xdg/xdg-specs/-/merge_requests/68 :)

To be honest, I *really* don't think that this will be accepted. The general sentiment is that the "Qt", "GTK" and GNOME/KDE groups were a mistake already, which is why no other DE besides the founders has any DE-specific category listed. In general, we don't want apps to be folded into DE-specific categories.

However, AppStream does know the concept of a [project group](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-project_group) for which Deepin certainly qualifies and that you can add to your metainfo files.
If you have apps that are core to the desktop (removing them would break your desktop environment), you can also mark them as compulsory for the desktop, see [compulsory_for_desktop](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-compulsory_for_desktop). That requires your DE to be registered as a [valid environment](https://specifications.freedesktop.org/menu-spec/latest/apb.html) in the menu spec, but you can also file a PR against AppStream if you want to have support for this a bit faster :-)

BLumia:
> If you have apps that are core to the desktop (removing them would break your desktop environment), you can also mark them as compulsory for the desktop, see [compulsory_for_desktop](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-compulsory_for_desktop). That requires your DE to be registered as a [valid environment](https://specifications.freedesktop.org/menu-spec/latest/apb.html) in the menu spec

See https://gitlab.freedesktop.org/xdg/xdg-specs/-/merge_requests/69 for the upstream MR :)

> but you can also file a PR against AppStream if you want to have support for this a bit faster :-)

Thanks! I'll wait for [xdg-spec!69](https://gitlab.freedesktop.org/xdg/xdg-specs/-/merge_requests/69) and see if that can be merged for now :)

Conan-Kudo:
I don't really have a problem with extending group support if someone cares. :)

BLumia:
> every frontend that uses icons for the existing groups would have an empty "DDE" group now, with no icon and no long description until support is added to every frontend.

For known frontends, I'm willing to help add such missing icon and/or description if needed :)

> That requires your DE to be registered as a [valid environment](https://specifications.freedesktop.org/menu-spec/latest/apb.html) in the menu spec,

https://gitlab.freedesktop.org/xdg/xdg-specs/-/merge_requests/69 was submitted few weeks ago and I currently don't know what's the next thing I need to do in order to get it merged 😂 

> but you can also file a PR against AppStream if you want to have support for this a bit faster :-)

See https://github.com/ximion/appstream/pull/472

BLumia:
Update: https://gitlab.freedesktop.org/xdg/xdg-specs/-/merge_requests/69 has been merged.

About the possibility of adding to the category list, I also saw there are similar request from LXQt: [desktop-file-utils#63](https://gitlab.freedesktop.org/xdg/desktop-file-utils/-/issues/63) and [desktop-file-utils!15](https://gitlab.freedesktop.org/xdg/desktop-file-utils/-/merge_requests/15). If we don't want this sort of entries be added to the category list, then maybe we should document them to explicitly tell developers we don't want apps to be folded into DE-specific categories.

BLumia:
Update: https://gitlab.freedesktop.org/xdg/xdg-specs/-/merge_requests/68 has been merged, and updated page are here: https://specifications.freedesktop.org/menu-spec/latest/apas02.html

So I guess maybe we can add support to those since it can be useful in some cases?

docs/api/spec/pk-concepts.xml
```diff
@@ -695,6 +695,10 @@
             <entry><literal>desktop-kde</literal></entry>
             <entry>KDE</entry>
           </row>
+          <row>
+            <entry><literal>desktop-dde</literal></entry>
+            <entry>DDE</entry>
```
Conan-Kudo, line 700:
```suggestion
            <entry>Deepin</entry>
```

lib/packagekit-glib2/pk-enum.h
```diff
@@ -523,6 +523,7 @@ typedef enum {
  * @PK_GROUP_ENUM_ACCESSORIES: Accessory packages
  * @PK_GROUP_ENUM_ADMIN_TOOLS: Administration tools packages
  * @PK_GROUP_ENUM_COMMUNICATION: Communication packages
+ * @PK_GROUP_ENUM_DESKTOP_DDE: DDE packages
```
Conan-Kudo, line 526:
```suggestion
 * @PK_GROUP_ENUM_DESKTOP_DDE: Deepin packages
```

Conan-Kudo (changes requested):
Two minor fixes, and then I think we can land this.

docs/api/spec/pk-concepts.xml
BLumia: Why? DDE is the name that we are supposed to use. DDE is the desktop environment's name, *deepin* is the distribution and branding name.

lib/packagekit-glib2/pk-enum.h
BLumia: Same as above, `DDE` is the one we suggested to use.

Conan-Kudo:
DDE is a terrible name, but whatever floats their boats, I guess.

Conan-Kudo (approved):

ximion:
> I don't really have a problem with extending group support if someone cares. :)

This will visually break all PK UIs, like Apper and GNOME PackageKit. Meh.

Conan-Kudo:
> > I don't really have a problem with extending group support if someone cares. :)
> 
> This will visually break all PK UIs, like Apper and GNOME PackageKit. Meh.

@BLumia you'll have to fix these issues or I'll need to revert this. I didn't think of that as a problem, tbh.

ximion:
> @BLumia you'll have to fix these issues or I'll need to revert this. I didn't think of that as a problem, tbh.

To be fair, I wouldn't consider this a big issue if these tools weren't completely unmaintained (so any adjustment would take forever to reach users). With this change, these tools will have an untranslated, icon-less entry for Deepin which contains zero entries on any distribution that isn't Arch.
Not crazy breakage, but something where I do wonder whether it's worth it, given that modern tools like GNOME Software don't even use the feature.

lib/packagekit-glib2/pk-enum.h
```diff
@@ -563,6 +564,7 @@ typedef enum {
 	PK_GROUP_ENUM_ACCESSORIES,
 	PK_GROUP_ENUM_ADMIN_TOOLS,
 	PK_GROUP_ENUM_COMMUNICATION,
+	PK_GROUP_ENUM_DESKTOP_DDE,
```
ximion, line 567: Please move that below `PK_GROUP_ENUM_NEWEST`, otherwise this will break ABI!

ximion:
I just checked the code: Apper will just emit a warning and not display it at all (since QPK doesn't support it), and GNOME PackageKit will flat out not display it at all. So, for these existing API users that I know, they don't iterate over `PkGroupEnum` anymore so this change will be mostly inert...
*However* this change still breaks ABI, and I think that *must* be fixed.

lib/packagekit-glib2/pk-enum.h
BLumia: Oh sorry I completely forget about this is also a library so ABI is a thing. The PR is here: https://github.com/PackageKit/PackageKit/pull/629. And I think maybe we should put it above NEWEST? (let's discuss this in that PR)

PR #611, opened 2023-03-13
pk-transaction: Add support for getting the RemainingTime property

pwithnall:
This seems to have been forgotten when the property was added.


ximion:
Ooof... Good catch!
Thank you for the patch!

PR #616, opened 2023-03-27
Pass -DPK_COMPILATION to the C++ compiler too.

arrowd:
This removes ad-hoc flag passing from backends that are written in C++.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -68,7 +68,7 @@ add_project_arguments ('-D_DEFAULT_SOURCE', language: 'c')
 add_project_arguments ('-DGLIB_VERSION_MIN_REQUIRED=GLIB_VERSION_2_44', language: 'c')
 
 # allow the daemon to include library files directly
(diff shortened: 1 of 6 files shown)


meson.build
```diff
@@ -69,6 +69,7 @@ add_project_arguments ('-DGLIB_VERSION_MIN_REQUIRED=GLIB_VERSION_2_44', language
 
 # allow the daemon to include library files directly
 add_project_arguments ('-DPK_COMPILATION', language: 'c')
```
ximion, line 71:
```suggestion
add_project_arguments ('-DPK_COMPILATION', language: ['c', 'cpp'])
```

This should work too and saves a line :-)

ximion (changes requested):

ximion:
Small comment, otherwise this looks good as `add_project_arguments` will not introduce a C++ compiler dependency if set globally (that happens later and only if a C++ backend is used).

ximion:

PR #619, opened 2023-04-07
Replace www.packagekit.org domain with Freedesktop URL

### LINKED ISSUE #618: www.packagekit.org domain was domain-grabbed; policy files, docs and other components still point to it
# Bug

There is multiple URLs  based in `http://www.packagekit.org` in the source code.

It seems that f-secure (as organisation) grabbed them – maybe that's even a benevolent move to avoid worse. I actually don't know why!

In any case, the domain doesn't seem to be under project control, and no links to it should be part of the source code anymore :)

# Affected files:

- policy/org.freedesktop.packagekit.policy.in
- etc/Vendor.conf
- lib/packagekit-glib2/pk-client.c
- docs/html/pk-reference.html
- docs/html/pk-faq.html
- client/pkmon.xml
- client/pkcon.xml
- PackageKit.doap

… of which, at worst the `etc/Vendor.conf` has some potential for abuse (being the URL shown to unsuspecting new users when a package can't be found). The rest is essentially project documentation.

marcusmueller:
Fixes #618

Split into three commits to make separation of concerns easier.

- metadata: Remove reference to defunct screenshots URL
- metadata: replace defunct www.packagekit.org URL with freedesktop URL
- docs: replace defunct www.packagekit.org URL with freedesktop URL

### DIFF
--- a/PackageKit.doap
+++ b/PackageKit.doap
@@ -22,6 +22,5 @@
 
   <license rdf:resource="http://www.gnu.org/licenses/gpl-2.0.txt" />
   <bug-database rdf:resource="http://bugzilla.freedesktop.org/" />
-  <screenshots rdf:resource="http://www.packagekit.org/pk-screenshots.html#gnome" />
 
 </Project>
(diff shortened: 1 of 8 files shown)


hughsie (approved):

ximion:

PR #622, opened 2023-04-19
Add FreeBSD CI

Zoid-dr:


aleixpol:
@arrowd can you review it maybe?

.cirrus.yml
```diff
@@ -0,0 +1,17 @@
+task:
+  freebsd_instance:
+    matrix:
+      - image_family: freebsd-13-2-snap
+  env:
+    LOCALBASE: /usr/local
+    LDFLAGS: -L$LOCALBASE/lib
```
arrowd, line 7: Is this really needed?

Zoid-dr: No. Fix.

arrowd:
Now the whole `env` block isn't needed too.

arrowd:
Looks good to me.

Tests are omitted for now, as one of them is failing and requires some debugging.

arrowd:
Can we get this in? I'm working on a FreeBSD backend and it is important to us to be sure the new code keeps working on FreeBSD.

Conan-Kudo:
I just noticed this after merging it, but please don't add `Sponsored by` information in PackageKit commits. I do not appreciate abusing revision control for permanent advertising of companies.

arrowd:
Ok, noted. This isn't a hard requirement for us.

Conan-Kudo:
@arrowd I've activated Cirrus CI on the project and it seems to be running. Is there a way to make it so that the name of the job isn't just "main" but something more useful (like have "freebsd" in the name or something)?

arrowd:
Sure, here is how this is done: https://github.com/CollaboraOnline/online/blob/master/.cirrus.yml#L2

Conan-Kudo:
Thanks, this is all set up now! Thanks for the contribution!

PR #630, opened 2023-05-17
Make use of SYSCONFDIR definition when setting a config file watch.

arrowd:
On FreeBSD the config file is located under /usr/local prefix.

### DIFF
--- a/src/pk-engine.c
+++ b/src/pk-engine.c
@@ -919,7 +919,7 @@ pk_engine_network_state_changed_cb (GNetworkMonitor *network_monitor,
 static void
 pk_engine_setup_file_monitors (PkEngine *engine)
 {
-	const gchar *filename = "/etc/PackageKit/PackageKit.conf";
+	const gchar *filename = SYSCONFDIR "/PackageKit/PackageKit.conf";
 	g_autoptr(GError) error = NULL;
 	g_autoptr(GFile) file_binary = NULL;
 	g_autoptr(GFile) file_conf = NULL;


arrowd:
Bump. Can we get this in?

ximion (changes requested):
Can you please remove the pretty verbose and reaching copyright disclaimer for a one-line change in the license text of this file? TBH, I don't think this one-liner even warrants adding copyright information to the file at all, but if that's a company requirement, you can add it (without the "Serenity Cyber Security" disclaimer text).

Apart from that, this patch looks fine :-)

arrowd:
> if that's a company requirement, you can add it (without the "Serenity Cyber Security" disclaimer text).

Yes, it is a requirement. Regarding the disclaimer, our lawyer said it is important. I'll ask if we're fine with removing it.

ximion:
I really don't think it is appropriate for a change and file like this, especially not since this disclaimer makes claims over code that was not written by you or your company (that can't be the intent, otherwise someone might think your company is liable for the origin of code not written by you). If this was a new feature in a separate file written entirely by you / your company, it would be a very different case for sure!

But this code does not even add/change any logic, it just changes the default search part of a file, while all code around it was written by Richard Hughes and many others who aren't even in the authors list.

src/pk-engine.c
```diff
@@ -17,6 +19,11 @@
  * You should have received a copy of the GNU General Public License
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Disclaimer: Serenity Cyber Security, LLC guarantees that this source code is
+ * open and governed by the terms of the GNU General Public License as published
+ * by the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
```
Conan-Kudo, line 26: My specific problem with this statement is that it implies Serenity Cyber Security has any special rights to make that statement. It does not. Requiring this statement for contributions is not only wrong IMO, it makes the company look like it has a role in the project governance and stewardship. This is obviously not true and misrepresents its stake in the project. Please drop it.

Conan-Kudo (changes requested):

src/pk-engine.c
arrowd: Yes, you're right. This was an misunderstanding on my side when reading our lawyer instructions.

This disclaimer should be placed into newly added files, which are entirely written by us.

Conan-Kudo:
> Is this better now? Or do you want me to remove copyright changes altogether?

If you can, that'd be preferred. Nobody has been adding those statements unless they've created brand new files.

arrowd:
> Nobody has been adding those statements unless they've created brand new files.

Which seems wrong to me. Anyways, I removed all our copyrights from the PR.

Conan-Kudo:
We typically have people added to the `AUTHORS` file: https://github.com/PackageKit/PackageKit/blob/main/AUTHORS

Conan-Kudo (approved):

Conan-Kudo:
@arrowd It looks like the tests are failing now for FreeBSD, could you take a look?

arrowd:
Yes, I'll try to take a look at it on the weekend.

PR #632, opened 2023-05-17
Enable testing on FreeBSD CI

arrowd:


Conan-Kudo (approved):

arrowd:
Strange, CI passed for this PR but failed for the merge commit. Is this just a fluke? Can the CI be restarted for the HEAD commit in the repository?

pwithnall:
> You're the author of that commit, do you have an idea what's the proper fix?

No I don’t, have you tried debugging it? Does FreeBSD use a different PackageKit backend or something like that? Are there valgrind warnings during the tests?

arrowd:
Yes, I tried debugging it a bit. The assertion seems to be reasonable - there is no way for "finished" signal to be emitted in that code path.

> Does FreeBSD use a different PackageKit backend or something like that?

FreeBSD doesn't have a native backend yet. I'm developing it in my own repo.

The testcase in question uses "dummy" backend on all platforms, it is hardcoded in the test.

> Are there valgrind warnings during the tests?

I didn't tried that. Valgrind had some problems on FreeBSD in the past, I'm not sure how usable it is at the moment.

pwithnall:
Are there any ways in which the FreeBSD test runs/setup differ from Linux? I don’t have any other ideas about what could be causing this, and don’t have easy access to a FreeBSD system to take a look.

arrowd:
> Are there any ways in which the FreeBSD test runs/setup differ from Linux? 

Don't think so.

All right, I'll try debug this further. Thanks for the input.

arrowd:
As evidenced by https://github.com/PackageKit/PackageKit/pull/641 assertions are disabled on Linux CI, which is why they are still green. FreeBSD CI compiles with `--buildtype debug`, which is probably the reason assertions are enabled.

ximion:
Is there any progress on this? The FreeBSD CI has been broken ever since it has been merged, and I am very inclined to disable it or silence the error so the CI status becomes actually useful again, at least until this issue is resolved.
TBH, I don't know why the tests would even fail like this on FreeBSD...

arrowd:
Didn't I explain it in my previous comment? Non FreeBSD CIs have assertions disabled which is why they don't catch a problem in the commit 409ba85b00f02bfd66589b457a148cce82ad2a34

PR #641, opened 2023-06-10
Do not merge: Testing if assertions are enabled on Linux CI

arrowd:


ximion:
You need to rebase this, FreeBSD CI is still broken :-/

arrowd:
This PR isn't meant to be merged. It shows that Linux CI passes while it should fail due to assertion failure.

ximion:
Okay, then we can close it - the Linux CI does call daemon tests as well now.

PR #649, opened 2023-07-30
apt: Fix 'search-file' regexp as per BRE pattern (which is the default)

sidt4:
We should either set the `'REG_EXTENDED'` flag (Extended Regular Expressions) to `regcomp()` and use the `"|"` regex pattern, or use the `"\\|"` regex pattern, as per the BRE (Basic Regular Expressions), which is the default when 'REG_EXTENDED' flag is not set.

From `'man grep'`:

Basic vs Extended Regular Expressions:

In basic regular expressions the meta‐characters ?, +, {, |, (, and ) lose their special meaning; instead use the backslashed versions \?, \+, \{, \|, \(, and \).

### DIFF
--- a/backends/apt/apt-job.cpp
+++ b/backends/apt/apt-job.cpp
@@ -1278,7 +1278,7 @@ PkgList AptJob::searchPackageFiles(gchar **values)
         }
 
         if (!search.empty()) {
-            search.append("|");
+            search.append("\\|");
         }
 
         if (value[0] == '/') {


ximion:
Good catch! - This whole code is really ugly in general though and would need a lot of refactoring due to https://lintian.debian.org/tags/uses-dpkg-database-directly
But for now, at least this patch makes the code a bit less buggy.
Thank you for the patch!

sidt4:
> https://lintian.debian.org/tags/uses-dpkg-database-directly

I guess we can access all functionality through `libapt-pkg` interface, which is the gateway to `apt` database.

ximion:
> I guess we can access all functionality through `libapt-pkg` interface, which is the gateway to `apt` database.

Not in the past, but the API surface has been expanded greatly in the past decade.

sidt4:
Should this be patched in `v1.2.6` in debian downstream ?

I see that you are the maintainer for https://tracker.debian.org/pkg/packagekit

ximion:
I'll make a new PackageKit release soon, so no need for that (in unstable/testing, at least).

sidt4:
I was more concerned about the `bookworm` not `testing` / `unstable` which doesn't impact normal users. Once the fix for https://gitlab.gnome.org/GNOME/gnome-software/-/issues/2254 lands, I'd get it backported to `gnome-software-43.5` in `bookworm`, which would require this patch.

PR #666, opened 2023-09-01
Use portable shebang in shell scripts.

arrowd:
This unbreaks most of tests on FreeBSD.

Unfortunately, one test is plagued by the same problem as described in https://github.com/ximion/appstream/pull/520 so I didn't enable testing yet.

### DIFF
--- a/backends/test/helpers/search-name.sh
+++ b/backends/test/helpers/search-name.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Copyright (C) 2007 Richard Hughes <richard@hughsie.com>
 #
--- a/tests/data/pk-spawn-test.sh
+++ b/tests/data/pk-spawn-test.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # Copyright (C) 2007 Richard Hughes <richard@hughsie.com>
 # Licensed under the GNU General Public License Version 2
 # This program is free software; you can redistribute it and/or modify


Conan-Kudo:
Please drop the `Sponsored by` statements. We don't permit them in this project.

arrowd:
Well, the fact is that I'm being paid for my work on PackageKit. Don't you find it unfair that companies that sponsor work on the project have no way to leave an acknowledgment/attribution to them in the code? First you ask us to remove copyrights from files we have not created (https://github.com/PackageKit/PackageKit/pull/630#issuecomment-1570721298) and now you want me to remove the "Sponsored by" line, which doesn't have any license power anyways. At the same time, @ximion and @aleixpol have already merged some of my commits with that line and had no problems with that.

And what about my FreeBSD backend PR? Is it realistic to wait for it to be reviewed and merged or we're wasting our time?

hughsie:
FWIW, I don't have a huge problem with them -- not ideal, but not a huge issue.

Conan-Kudo:
> And what about my FreeBSD backend PR? Is it realistic to wait for it to be reviewed and merged or we're wasting our time?

I am actually looking over that PR. It is, however, a lot of commits to go through.

arrowd:
> I am actually looking over that PR. It is, however, a lot of commits to go through.

This is great to hear. So, if you insist on removing the "Sponsored by:" line I'll do it, but I still feel it isn't a thing worth fighting for.

Conan-Kudo:
My main problem is what happens when the company's name gets reused years later. The idea has good intent (rewarding companies for supporting development), but company names often outlive the companies themselves.

A great example of this is Nextel. The company "Nextel" that was created a few years ago uses the same brand as the one from twenty years ago, but has nothing in common with them otherwise. It can create serious confusion and potentially other problems.

And even if the scenario above doesn't happen, sometimes it can become a problem when management changes within the same company. One day, they're fine with being associated with a project, and then another day they'd like nothing to do with it. And we're stuck.

Conan-Kudo (approved):

Conan-Kudo:
Thanks for the patch!

PR #678, opened 2023-09-28
dnf: Avoid downloading caches multiple times in one call

aleasto:


aleasto:
The issue here is that `dnf_utils_create_sack_for_filters` will eventually try to update the caches again.

To be honest, I'm not convinced that honouring the `cache-age` hint in `RefreshCache` makes any sense: we are explicitly asking to refresh the cache... But of the two this change is the less intrusive one.

@Conan-Kudo @gordonmessmer @hughsie

Conan-Kudo (approved):
I think this is probably reasonable.

backends/dnf/pk-backend-dnf.c
```diff
@@ -1706,6 +1706,9 @@ pk_backend_refresh_cache_thread (PkBackendJob *job,
 	/* invalidate the sack cache after downloading new metadata */
 	pk_backend_sack_cache_invalidate (backend, "downloaded new metadata");
 
+	// We just downloaded our cache, avoid doing so again
```
ximion, line 1709:
```suggestion
	/* We just downloaded our cache, avoid doing so again */
```

ximion (commented):
Just a style nitpick, otherwise I think this is sensible...

PR #681, opened 2023-09-29
PkClient: Replace the GSimpleAsyncResult with the GTask API

tintou:
The GSimpleAsyncResult API is deprecated and can be replaced by GTask, also avoid copying the GError when set.

### DIFF
--- a/lib/packagekit-glib2/pk-client.c
+++ b/lib/packagekit-glib2/pk-client.c
@@ -163,7 +163,7 @@ struct _PkClientState
 	GDBusProxy			*proxy_props;
 	GCancellable			*cancellable;
 	GCancellable			*cancellable_client;
-	GSimpleAsyncResult		*res;
+	GTask				*res;
 	PkBitfield			 filters;
 	PkClient			*client;
 	PkProgress			*progress;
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
Isn't this an ABI break?

cc: @hughsie @ximion

tintou:
No ABI break here

ximion:
> Isn't this an ABI break?
> 
> cc: @hughsie @ximion

No, API-wise this is fine and the behavior should also match. It's a pretty nice maintenance task, actually - very nice!
I just need a bit of spare time to look over the code in more detail (at first glance, it looked fine).

Conan-Kudo (approved):

sidt4:
Indeed, that's a much needed change :+1:

ximion:
I finally had time to look into this, and I have no notes - this patch looks good as-is! Pretty awesome!
The one thing I am a bit worried about is subtle behaviour changes that I may not be aware of that this switch might bring, but looking through the docs I didn't find any.
So, there's no reason not to merge this change - thank you for the patch!

ximion (approved):

PR #682, opened 2023-09-29
the issue with the file name has been fixed

erkanisik1:


ximion:
Thanks for the quick fix! :-)
Can you rebase in future instead of creating multiple merge commits in a PR? I've squashed all of them into one small commit, so it's not an issue here anymore, but would be nice in future.

Conan-Kudo:
@erkanisik1 FYI, [we have guidelines for patches](https://github.com/PackageKit/PackageKit/blob/main/HACKING) that we would appreciate it if you followed them so we can review and merge things more quickly in the future.

Issue #645, opened 2023-07-21
Two entries listed for each updated package in /var/lib/PackageKit/offline-update-competed


sidt4:
From: https://gitlab.gnome.org/GNOME/gnome-software/-/issues/2249#note_1796310

This happens on debian / apt. Not sure about other backends.

I simulated an offline update for a single package.

In `pk_offline_auth_set_results` ( `pk-offline-private.c` ), in the below code

```c
	/* save packages if any set */
	packages = pk_results_get_package_array (results);
	if (packages->len > 0) {
		g_autoptr(GString) string = NULL;
		string = g_string_new ("");
		for (i = 0; i < packages->len; i++) {
			package = g_ptr_array_index (packages, i);
			switch (pk_package_get_info (package)) {
			case PK_INFO_ENUM_UPDATING:
			case PK_INFO_ENUM_INSTALLING:
				g_string_append_printf (string, "%s,",
							pk_package_get_id (package));
				break;
			default:
				break;
			}
		}
		if (string->len > 0)
			g_string_set_size (string, string->len - 1);
		g_key_file_set_string (key_file,
				       PK_OFFLINE_RESULTS_GROUP,
				       "Packages",
				       string->str);
	}
```

There are 5 entries in `packages` all for the single package ( say, X ). The following were the `info` ( probably `action` ? ) for the 5 entries:

1. `PK_INFO_ENUM_PREPARING` - package X
1. `PK_INFO_ENUM_DECOMPRESSING` - package X
1. `PK_INFO_ENUM_INSTALLING` - package X
1. `PK_INFO_ENUM_PREPARING` - package X
1. `PK_INFO_ENUM_INSTALLING` - package X

So, there are 2 entries written to `/var/lib/PackageKit/offline-update-competed`

sidt4:
This was on latest debian unstable with `packagekit-1.2.6-5`.

sidt4:
This issue doesn't appear in `dnf` backend ( Tested on `PackageKit-1.2.6-6.fc38.x86_64` )

mcrha:
This is a duplicate of https://github.com/PackageKit/PackageKit/issues/284

Bushido1:
Hello,

I am experiencing this issue in Linux Ubuntu 22.04 LTS with gnome-software 41.5.
![image](https://github.com/PackageKit/PackageKit/assets/11409601/bf4d5e48-6c05-4a61-9d71-40c7174be314)

mcrha:
[...]

PR #686, opened 2023-10-05
offline-updates: Add only one entry per package to 'offline-update-competed' file

sidt4:
This is only needed to handle backends where we receive multiple `PK_INFO_ENUM_INSTALLING` (apt) / `PK_INFO_ENUM_UPDATING` status for a single package update.

Fixes https://github.com/PackageKit/PackageKit/issues/645


ximion:
I do not like this implementation, but I do think a deduplication safeguard makes sense, so I'll merge this and patch it up in a second commit (to not "steal" the credit and implement it afterwards).

Normally this would go through review properly, but I'm about to make a PK release and it would be nice to have this patch included.

PR #702, opened 2023-11-15
Replace all the GSimpleAsyncResult from the library with the GTask API 

tintou:
No API/ABI change here, use the modern GLib API for async operations.


Conan-Kudo:
Does the current minimum glib2 version in Meson cover this API usage?

tintou:
Yes, the GTask API is there since 2.36

Le 17 novembre 2023 12:59:30 GMT+01:00, "Neal Gompa (ニール・ゴンパ)" ***@***.***> a écrit :

Conan-Kudo (approved):

PR #706, opened 2023-12-18
Check string length correctly to prevent memory over-consumption (RHEL-19527)

rh-tguittet:
pk_strlen() cannot return larger values than the sentinel value (1024) in this case. This invalid check could allow an unprivileged user to consume all of the system's memory (tested for versions <= 1.2.6).

### DIFF
--- a/src/pk-transaction.c
+++ b/src/pk-transaction.c
@@ -2484,7 +2484,7 @@ pk_transaction_strvalidate (const gchar *text, GError **error)
 				     "Invalid input passed to daemon: zero length string");
 		return FALSE;
 	}
-	if (length > 1024) {
+	if (length >= 1024) {
 		g_set_error (error, PK_TRANSACTION_ERROR, PK_TRANSACTION_ERROR_INPUT_INVALID,
 			     "Invalid input passed to daemon: input too long: %u", length);
 		return FALSE;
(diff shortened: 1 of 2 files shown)


src/pk-transaction.c
```diff
@@ -2484,7 +2484,7 @@ pk_transaction_strvalidate (const gchar *text, GError **error)
 				     "Invalid input passed to daemon: zero length string");
 		return FALSE;
 	}
-	if (length > 1024) {
+	if (length == 1024) {
```
hughsie, line 2487: Maybe >= might be safer here? Thanks for the PR!

rh-tguittet: All other function calls to pk_strlen() use == but feel free to change the commit ;)

rh-tguittet: @hughsie are you satisfied with the uniformity of `==` or do you _want_ me to change it to `>=` ?

Conan-Kudo: He'd prefer you to change it, as would I.

rh-tguittet: Did that (sorry for the lag). Let me know if you wish me to go on patching all of the other checks for pk_strlen's output.

Conan-Kudo: Please do so, thanks! And squash the commit once you're done.

Conan-Kudo:
@rh-tguittet Can you please change the title of the commit message from "Fix RHEL-19527" to "Check string length correctly to prevent memory leak" or something similarly useful? Nobody can access that RHEL Jira issue.

Conan-Kudo (approved):

ximion:
Looks good to me :-)

ximion (approved):

PR #712, opened 2024-01-18
Add dnf5 posttrans notify plugin

aleasto:
Just like #600, we should do the same for dnf5

### DIFF
--- a/backends/dnf/meson.build
+++ b/backends/dnf/meson.build
@@ -1,6 +1,8 @@
 appstream_dep = dependency('appstream', version: '>=0.14.0')
 dnf_dep = dependency('libdnf', version: '>=0.43.1')
+dnf5_dep = dependency('libdnf5').partial_dependency(includes: true, compile_args: true)
 rpm_dep = dependency('rpm')
+sdbus_cpp_dep = dependency('sdbus-c++')
 c_args = ['-DG_LOG_DOMAIN="PackageKit-DNF"']
 
 if meson.get_compiler('c').has_function('hy_query_get_advisory_pkgs', prefix: '#include <libdnf/hy-query.h>', dependencies: dnf_dep)
@@ -18,6 +20,26 @@ python_package_dir = join_paths(python_package_dir, 'dnf-plugins')
 install_data(
   'notify_packagekit.py',
   install_dir: join_paths(python_package_dir),
+)
+
+add_languages('cpp')
+shared_module(
+  'notify_packagekit',
+  'notify_packagekit.cpp',
+  cpp_args: '-std=c++20',
+  include_directories: packagekit_glib2_includes,
+  dependencies: [
+    dnf5_dep,
+    sdbus_cpp_dep,
+  ],
+  name_prefix: '',
+  install: true,
+  install_dir: get_option('libdir') / 'libdnf5/plugins/',
+)
+
+install_data(
+  'notify_packagekit.conf',
+  install_dir: get_option('sysconfdir') / 'dnf/libdnf5-plugins/'
 )
 
 shared_module(
--- a/backends/dnf/notify_packagekit.cpp
+++ b/backends/dnf/notify_packagekit.cpp
@@ -1,0 +1,90 @@
+// Copyright (C) 2024 Alessandro Astone <ales.astone@gmail.com>
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
+#include <libdnf5/base/base.hpp>
+#include <libdnf5/common/exception.hpp>
+
+#include <sdbus-c++/sdbus-c++.h>
+
+#include <algorithm>
+#include <cstdlib>
+
+#include <packagekit-glib2/pk-version.h>
+
+using namespace libdnf5;
+
+namespace {
+
+constexpr const char * PLUGIN_NAME{"notify_packagekit"};
+
+constexpr plugin::Version PLUGIN_VERSION{.major = PK_MAJOR_VERSION, .minor = PK_MINOR_VERSION, .micro = PK_MICRO_VERSION};
+
+constexpr const char * attrs[]{"author.name", "author.email", "description", nullptr};
+constexpr const char * attrs_value[]{"Alessandro Astone", "ales.astone@gmail.com",
+        "Notify packagekitd when packages are installed, updated, or removed."};
+
+class NotifyPackagekitPlugin : public plugin::IPlugin {
+public:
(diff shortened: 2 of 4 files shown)


aleasto:
cc: @gordonmessmer as you implemented the dnf4 plugin, and @ximion 

I don't know if it's acceptable to just invoke `gdbus` rather than linking some dbus library, which would probably be GIO to match the rest of packagekit.

Also I don't know if it's acceptable to add a dependency to c++ like this.

meson.build
```diff
@@ -1,8 +1,8 @@
-project('PackageKit', 'c',
+project('PackageKit', 'c', 'cpp',
```
Conan-Kudo, line 1: This needs to be in the backend's `meson.build`, not here.

meson.build
```diff
@@ -1,8 +1,8 @@
...
   version : '1.3.0',
   license : 'LGPL-2.1+',
   meson_version : '>=0.50',
-  default_options : ['warning_level=2', 'c_std=c99'],
+  default_options : ['warning_level=2', 'c_std=c99', 'cpp_std=c++20'],
```
Conan-Kudo, line 5: This needs to be in the backend's `meson.build`, not here.

Conan-Kudo (changes requested):

meson.build
aleasto: Thanks, done.

backends/dnf/notify_packagekit.cpp
```diff
@@ -0,0 +1,80 @@
...
+        }
+        return nullptr;
+    }
+
+    void post_transaction(const libdnf5::base::Transaction & transaction) override;
+};
+
+void NotifyPackagekitPlugin::post_transaction(const libdnf5::base::Transaction & transaction) {
+    std::system("gdbus call --system --dest org.freedesktop.PackageKit --object-path /org/freedesktop/PackageKit --method org.freedesktop.PackageKit.StateHasChanged posttrans");
+}
```
Conan-Kudo, line 47: Can you use `sdbus-cpp` here instead of shelling out to `gdbus`?

backends/dnf/notify_packagekit.cpp
```diff
@@ -0,0 +1,80 @@
+#include <libdnf5/base/base.hpp>
```
Conan-Kudo, line 1: This needs a license header, similar to what is present in other source files.

backends/dnf/notify_packagekit.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include <libdnf5/common/exception.hpp>
+
+#include <algorithm>
+#include <cstdlib>
+
+using namespace libdnf5;
+
+namespace {
+
+constexpr const char * PLUGIN_NAME{"template"};
```
Conan-Kudo, line 11: plugin name is wrong

backends/dnf/notify_packagekit.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+constexpr plugin::Version PLUGIN_VERSION{.major = 1, .minor = 0, .micro = 0};
```
Conan-Kudo, line 13: This should probably match PackageKit's version.

Conan-Kudo (changes requested):

backends/dnf/notify_packagekit.cpp
aleasto: I'll see if it isn't necessarily complex.
Otherwise `gdbus` or alternatively `busctl` are pretty much guaranteed to be installed already so we wouldn't be adding new dependencies.

backends/dnf/meson.build
```diff
@@ -20,6 +22,26 @@ install_data(
...
 )
 
+add_languages('cpp')
+shared_module(
+  'notify_packagekit',
+  'notify_packagekit.cpp',
+  cpp_args: '-std=c++20',
+  include_directories: packagekit_glib2_includes,
+  dependencies: [
+    dnf5_dep,
```
aleasto, line 32: anyone familiar with meson can tell me why i don't actually need to add the dnf5 dependency to the module? it seems like headers are propagated globally? that's weird...

aleasto: Ah well, because they're installed in a default includes directory

ximion:
@Conan-Kudo Does this look good to you now? I intend to do a new PK release tomorrow...

Conan-Kudo:
The Fedora dockerfile needs to be bumped to 40.

aleasto:
Done

sidt4:
iirc, @mcrha has been working on some dnf5 stuff. Just wanted to check if there is no overlap here.

aleasto:
As far as I can see, @mcrha is implementing a dnf5 backend in gnome-software to ditch PackageKit altogether.

ximion:
> As far as I can see, @mcrha is implementing a dnf5 backend in gnome-software to ditch PackageKit altogether.

That is my understanding as well, which is slightly annoying as that moves us to a future were we yet again have to implement support for all the different PMs in many places, instead of having just one place to do the work once and for everyone.

I do understand where this is coming from though - PK could really use a lot more time investment to move it into the future.

Conan-Kudo (approved):

mcrha:
> I do understand where this is coming from though - PK could really use a lot more time investment to move it into the future.

Correct, that's the main reason for that work.

By the way, I'm not sure this change has a meaning at the moment, when, in fact, PckageKit uses dnf4 only. Yes, one can install/update/remove with both dnf versions on a single machine, still a better option would be to truly use dnf4 or dnf5, not a mix of both.

Conan-Kudo:
Well, I'd like to have a dnf5 backend, since other software centers using PackageKit should benefit from it.

mcrha:
Sure, that makes perfect sense and Philip, the gnome-software maintainer, prefers that idea too.

aleasto:
> > I do understand where this is coming from though - PK could really use a lot more time investment to move it into the future.
> 
> Correct, that's the main reason for that work.
> 
> By the way, I'm not sure this change has a meaning at the moment, when, in fact, PckageKit uses dnf4 only. Yes, one can install/update/remove with both dnf versions on a single machine, still a better option would be to truly use dnf4 or dnf5, not a mix of both.

PackageKit is eventually going to be run alongside dnf5, presumably in f41 already. This change is so that dnf5 CLI operations can notify PackageKit, addressing the concerns at https://github.com/PackageKit/PackageKit/pull/578#issuecomment-1384781816

PR #717, opened 2024-03-01
PkClient/Progress/Category: Use a more modern GObject style

tintou:
Use G_DEFINE_TYPE_WITH_PRIVATE which does everything for us. Also use GParamSpec directly to notify a property and use *_get_instance_private() to access the PkClientPrivate structure directly.

### DIFF
--- a/lib/packagekit-glib2/pk-category.c
+++ b/lib/packagekit-glib2/pk-category.c
@@ -34,8 +34,6 @@
 
 static void     pk_category_finalize	(GObject     *object);
 
-#define PK_CATEGORY_GET_PRIVATE(o) (G_TYPE_INSTANCE_GET_PRIVATE ((o), PK_TYPE_CATEGORY, PkCategoryPrivate))
-
 /**
  * PkCategoryPrivate:
  *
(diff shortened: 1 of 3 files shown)


hughsie:
Heavy approval by me, but the reason I didn't do this years ago was worries about if it was ABI compatible. IIRC debian got very upset when I did this with one of my other projects.

tintou:
This should still be ABI compatible (?) because I haven't removed the dangling FooPrivate *priv in the object, even if it is completely unused :slightly_smiling_face:

hughsie:
> because I haven't removed the dangling FooPrivate *priv in the object

Aha, great idea!

ximion:
> This should still be ABI compatible (?) because I haven't removed the dangling FooPrivate *priv in the object, even if it is completely unused 🙂

Yes, I noticed that and was looking for it ;-)

> IIRC debian got very upset when I did this with one of my other projects.

That would have had to be me, but as long as no publicly used API changes, I'd actually be happy with it. Very odd.

But I also very much like this change! It's a bit much to review, I'll have another look at it later and merge it if I don't spot anything suspicious (but from an initial look, this looks fine - thank you @tintou for doing that fairly boring maintenance work!).

ximion:
This was an incredibly mind-numbing review and I hope I didn't miss anything, but the changes look good as they are and I didn't see anything that could cause problems.

Thank you so much @tintou for taking on this task! This will make maintenance of PK a lot easier!

PR #721, opened 2024-03-08
alpm: add compatibility with libalpm 14

jelly:
In libalpm alpm_conflict_t package struct members are now of type alpm_pkg_t instead of char*.

libalpm sadly does not expose the version in `libalpm.h` in theory this can be made backwards compatible, but not sure how to hack that in with `pkgconf --modversion libalpm` and most distro's shipping pacman are rolling release so should have pacman 6.1

### DIFF
--- a/backends/alpm/meson.build
+++ b/backends/alpm/meson.build
@@ -1,4 +1,4 @@
-alpm_dep = dependency('libalpm', version: '>=13.0.0')
+alpm_dep = dependency('libalpm', version: '>=14.0.0')
 
 shared_module(
   'pk_backend_alpm',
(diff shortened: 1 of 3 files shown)


eworm-de:
This is missing support for `CacheServer` in configuration. Currently `packagekit` crashes if that is present.

https://gitlab.archlinux.org/pacman/pacman/-/commit/dc91476555e5ac636d27bbe635e647a70f911e71

ximion:
There's no way for me to judge alpm packages properly, but the code looks sane and the comment from @eworm-de was addressed, so - let's go!

Thanks for the patch!

jelly:
Thanks! We have it running with this patch since I created the PR and haven't had any reported issues so far.

PR #725, opened 2024-03-18
pk-offline-update: Add file to POTFILES

bitigchi:
Apparently the grammatical correct dash here is em, not a hyphen or en dash.


bitigchi:
Thanks for merging the other i18n patch. After this, is it possible to update the strings in Transifex?

PR #728, opened 2024-03-22
pk-offline-update: Fix crash due to invalid g_auto declaration in switch / case

sidt4:
`case PK_PROGRESS_TYPE_PERCENTAGE` is just a label and doesn't offer any scope as such. So, `tmp_perc` auto variable will be allocated in the stack, but will not be initialized to `NULL` for `case:` statements other than `PK_PROGRESS_TYPE_PERCENTAGE`, causing GLib to consider the `tmp_perc` garbage value as a valid memory address allocated via `g_malloc()`, triggering an invalid `free()` causing the crash.

Rather than adding a scope `{}` within `case`, and causing a huge noise in git diff, we just move the declaration to an upper scope.

### DIFF
--- a/client/pk-offline-update.c
+++ b/client/pk-offline-update.c
@@ -113,6 +113,7 @@ pk_offline_update_progress_cb (PkProgress *progress,
 	PkStatusEnum status;
 	gint percentage;
 	g_autofree gchar *msg = NULL;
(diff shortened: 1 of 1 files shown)


hughsie (approved):

ximion:
Thank you! :-)

Issue #747, opened 2024-04-28
pk_task_generic_finish: assertion 'G_IS_SIMPLE_ASYNC_RESULT (res)' failed


mikwee:
Ran `pkcon get-transactions` on Fedora 40, and got this error:
```
Getting transactions                    [=========================]         
Finished                                [=========================]         

(pkcon:11343): PackageKit-CRITICAL **: 11:56:10.603: pk_task_generic_finish: assertion 'G_IS_SIMPLE_ASYNC_RESULT (res)' failed
Segmentation fault (core dumped)
```
My vague knowledge of segfault makes me think it's a RAM issue, but I was told to open an issue here, so I did.

sidt4:
> My vague knowledge of segfault makes me think it's a RAM issue

Not a RAM issue.

### On Debian with `packagekit-1.2.8`.

```
$ pkcon get-transactions
Getting transactions                    [=========================]         
Finished                                [=========================]         

(pkcon:142994): PackageKit-CRITICAL **: 10:55:57.869: pk_task_generic_finish: assertion 'G_IS_SIMPLE_ASYNC_RESULT (res)' failed
Segmentation fault (core dumped)
```

### On latest git `packagekit` (with port to `GTask`):

```
$ pkcon get-transactions
Getting transactions                    [=========================]         
Finished                                [=========================]         
Segmentation fault (core dumped)
```

### Backtrace:

```
(gdb) bt
#0  0x00007f03fc79cd8c in pk_task_generic_finish (task=0x5555862162c0 [PkTaskText], res=0x555586213350, error=0x7fff186777a8) at ../lib/packagekit-glib2/pk-task.c:2192
#1  0x0000555585a2d6b6 in pk_console_finished_cb (object=0x5555862162c0 [PkTaskText], res=0x555586213350, data=0x555586203cb0) at ../client/pk-console.c:663
#2  0x00007f03fc4666b3 in g_task_return_now (task=task@entry=0x555586213350 [GTask]) at ../../../gio/gtask.c:1361
#3  0x00007f03fc467353 in g_task_return (type=<optimized out>, task=0x555586213350 [GTask]) at ../../../gio/gtask.c:1430
#4  g_task_return (task=0x555586213350 [GTask], type=<optimized out>) at ../../../gio/gtask.c:1387
#5  0x00007f03fc767435 in pk_client_state_finish (state=0x555586213210 [PkClientState], error=0x0) at ../lib/packagekit-glib2/pk-client.c:245
[...]

PR #751, opened 2024-05-10
pkcon: Use appropriate finish call for each async call

sidt4:
Fixes: #747

### DIFF
--- a/client/pk-console.c
+++ b/client/pk-console.c
@@ -637,7 +637,7 @@ pk_console_progress_cb (PkProgress *progress, PkProgressType type, gpointer data
 }
 
 static void
-pk_console_finished_cb (GObject *object, GAsyncResult *res, gpointer data)
+pk_console_process_results (PkResults *results, PkConsoleCtx *ctx, const GError *finish_error)
 {
 	const gchar *filename;
 	gboolean ret;
@@ -645,11 +645,8 @@ pk_console_finished_cb (GObject *object, GAsyncResult *res, gpointer data)
 	PkPackageSack *sack;
 	PkRestartEnum restart;
 	PkRoleEnum role;
-	PkConsoleCtx *ctx = (PkConsoleCtx *) data;
-	g_autoptr(GError) error = NULL;
 	g_autoptr(GFile) file = NULL;
 	g_autoptr(PkError) error_code = NULL;
-	g_autoptr(PkResults) results = NULL;
 
 	/* no more progress */
 	if (ctx->is_console) {
@@ -659,13 +656,11 @@ pk_console_finished_cb (GObject *object, GAsyncResult *res, gpointer data)
 		g_print ("%s\n", _("Results:"));
 	}
 
-	/* get the results */
-	results = pk_task_generic_finish (PK_TASK (ctx->task), res, &error);
 	if (results == NULL) {
 		/* TRANSLATORS: we failed to get any results, which is pretty
 		 * fatal in my book */
-		g_print ("%s: %s\n", _("Fatal error"), error->message);
-		switch (error->code - 0xff) {
+		g_print ("%s: %s\n", _("Fatal error"), finish_error->message);
+		switch (finish_error->code - 0xff) {
 		case PK_ERROR_ENUM_ALL_PACKAGES_ALREADY_INSTALLED:
 		case PK_ERROR_ENUM_REPO_NOT_AVAILABLE:
 			ctx->retval = PK_EXIT_CODE_NOTHING_USEFUL;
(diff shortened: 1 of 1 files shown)


ximion:
Instead of adding that extra variable, could you define two dispatch functions `pk_console_client_finished_cb` and `pk_console_task_finished_cb` that extract the PkResults and forward it to a generic `pk_console_process_results` function?

That would do away with the extra variable and might be a bit cleaner...

sidt4:
Fixed.

client/pk-console.c
```diff
@@ -637,30 +637,21 @@ pk_console_progress_cb (PkProgress *progress, PkProgressType type, gpointer data
...
 
-	/* no more progress */
-	if (ctx->is_console) {
-		pk_progress_bar_end (ctx->progressbar);
-	} else {
-		/* TRANSLATORS: the results from the transaction */
-		g_print ("%s\n", _("Results:"));
-	}
+	if (err != NULL)
+		error = g_error_copy (err);
```
ximion, line 653: Why are you copying the error here? Why not get rid of the local error and use the passed GError directly? (Ideally checking it for NULL before using it, just to be safe)

client/pk-console.c
```diff
@@ -833,6 +824,46 @@ pk_console_finished_cb (GObject *object, GAsyncResult *res, gpointer data)
...
 }
 
+static void
+pk_console_client_finished_cb (GObject *object, GAsyncResult *res, gpointer data)
+{
+	g_autoptr(GError) error = NULL;
+	g_autoptr(PkResults) results = NULL;
+	PkConsoleCtx *ctx = (PkConsoleCtx *) data;
+
+	/* no more progress */
```
ximion, line 834: Couldn't this code move into the generic function as well?

client/pk-console.c
```diff
@@ -637,30 +637,21 @@ pk_console_progress_cb (PkProgress *progress, PkProgressType type, gpointer data
 }
 
 static void
-pk_console_finished_cb (GObject *object, GAsyncResult *res, gpointer data)
+pk_console_process_results (PkResults *results, PkConsoleCtx *ctx, GError *err)
```
ximion, line 640: nit: Just to make abundantly clear that this is not a return error and a specific error is expected, naming the parameter something like `finish_error` would maybe make sense, so it's easier to grasp for someone reading this in future.

ximion (changes requested):

client/pk-console.c
sidt4: Because of this below code in the function. I just wanted to make minimal changes.

```c
	/* write the sack to disk */
	if (role == PK_ROLE_ENUM_GET_PACKAGES && filename != NULL) {
		file = g_file_new_for_path (filename);
		ret = pk_package_sack_to_file (sack, file, &error);
		if (!ret) {
			g_print ("%s: %s\n", _("Fatal error"), error->message);
			ctx->retval = PK_EXIT_CODE_TRANSACTION_FAILED;
			goto out;
		}
	}
```

client/pk-console.c
sidt4: Sure. Making it `const` will be an option too.

client/pk-console.c
sidt4: There were some issues with the progress bar, though I can't reproduce it now.

Will check and move it to generic function.

client/pk-console.c
ximion: Right, but please don't reuse a GError like that, that's a recipe for disaster usually.

Just make a local `g_autoptr(GError) error = NULL` for the function, and keep the `finish_error` that has been passed to the function intact and unchanged. For the error to work for `pk_package_sack_to_file`, it has to be NULL anyway.

client/pk-console.c
ximion: Good - if it can't be generalized, having it like this would also be okay, but at least in theory, oder of operations between finishing the async callback and ending the progress bar display shouldn't matter...

client/pk-console.c
sidt4: > (Ideally checking it for NULL before using it, just to be safe)

If we're planning to do that, we should do that in all places. Personally, we should just inspect the code and fix any programmer errors when `error` is NULL when it should not, rather than adding checks. So, I'm skipping the NULL check for now.

client/pk-console.c
sidt4: I couldn't reproduce it. So, I'm moving it to generic function.

ximion (approved):

ximion:
Thank you for the patch, this looks good now! :-)

PR #753, opened 2024-05-16
pk-offline-update: Fix crash during error cases

sidt4:
`g_prefix_error()` does nothing if the `error` or `*error` is NULL. So, we crash when accessing `error->message`.

Also, the error message is not propagated to the caller due to this issue.


ximion:
I checked the callers, this should indeed be a fresh error, and not a prefixed one (which makes sense, as per API conventions).

So yeah, nice find and thanks for the fix!

PR #766, opened 2024-06-04
pk-progress: Fix a leak of package ID

pwithnall:
And remove an invalid free of the item progress object at the same time. This was never a problem in practice, because it had already been cleared (correctly, using `g_object_unref()`) in `dispose`.


ximion:
Nice, thank you!

PR #768, opened 2024-06-07
dnf/zypp: Add support for language(ll_CC) to what-provides queries

pwithnall:
This was already documented in the [‘package component provides naming’ spec](./docs/provides-component-naming.txt), so should be uncontroversial.

This is part of a plan to drop the fedora-langpacks plugin from gnome-software, and replace it with the equivalent of a call to `pkcon what-provides 'language(zh_CN)'` (for example).

See https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/2055#note_2134286


ximion:
> This is part of a plan to drop the fedora-langpacks plugin from gnome-software, and replace it with the equivalent of a call to pkcon what-provides 'language(zh_CN)' (for example).

Very nice! It could potentially even be integrated with AppStream's language-pack support... The AppStream side was added ages ago when there was a huge push to get this done (even from the Ubuntu side), but then people had to work on other things and work on this eventually stopped, sadly...
https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Localization.html

It's nice to see this revived a bit :-)

PR #773, opened 2024-07-20
pk_dbus_get_session: Move ConsoleKit check under the appropriate ifdef case

arrowd:


pwithnall (commented):
This looks legit to me

Conan-Kudo (approved):

PR #774, opened 2024-07-22
pk-client: Fix runtime warning when backend does not support role

mcrha:
The daemon returns "op not supported by the backend" as part of the method call response, but this response can be received after the "Destroy" signal, which causes infinite wait for the transaction end and runtime warnings about NULL error being passed to the g_task_return_error() on the client side.

This is a regression from commit https://github.com/PackageKit/PackageKit/commit/44215db95d9cfda23045d9a9b97175519f473eb2

### DIFF
--- a/lib/packagekit-glib2/pk-client.c
+++ b/lib/packagekit-glib2/pk-client.c
@@ -1616,11 +1616,11 @@ pk_client_signal_cb (GDBusProxy *proxy,
 	if (g_strcmp0 (signal_name, "Destroy") == 0) {
 		g_autoptr(GError) local_error = NULL;
 
-		if (state->waiting_for_finished)
+		if (state->waiting_for_finished) {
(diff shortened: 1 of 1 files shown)


Conan-Kudo (approved):

pwithnall:
> The daemon returns "op not supported by the backend" as part of the method call response, but this response can be received after the "Destroy" signal

From this, it sounds to me that the bug is that the daemon is emitting the `Destroy` signal before the method call reply.

mcrha:
> From this, it sounds to me that the bug is that the daemon is emitting the `Destroy` signal before the method call reply.

Even if it does, the code as it was before this change still generated the runtime warnings due to NULL `error` parameter passed to the `g_task_return_error()` and caused infinite wait for completion. The code logic wasn't correct here.

PR #775, opened 2024-07-22
Enable testing in FreeBSD CI

arrowd:


ximion:
Looks like this fixes the CI, thanks!

arrowd:
The CI somehow passed in the PR, but failed after the merge. I'm looking into that.

arrowd:
Ok, it seems it has something to do with file monitoring stuff and is not reproducible in 100% cases.

Right now I'm working on improving GFileMonitor implementation for FreeBSD and it is known to fix AppStream tests that are currently also failing to do file monitoring bugs. I suspect that new implementation will fix this CI fluctuations too.

For now I propose to not revert this change, but just keep in mind that this exact test may fails sometimes.

Issue #771, opened 2024-06-30
'download-size' is not hooked up to UI or spawned python backends


joebonrichie:

sidt4:
Can you explain in more detail ?

Maybe you're referring to PR https://github.com/PackageKit/PackageKit/pull/780 & https://github.com/PackageKit/PackageKit/issues/781,

joebonrichie:
Closer to #781, however it seems it isn't accounting for python spawned backends; in addition, the download-size isn't reported via `pkcon get-details` for example.

We are using this patch for our out of tree backend: https://github.com/getsolus/PackageKit/commit/fc7579d760ad7632c234ef9312fb8ac35bcc63fb. This then allows the `eopkg` backend to report the download-size such that G-S can pick it up.

Obviously, the problem with this is it breaks all existing python spawned backends. With the way spawned backends is architectured it seems harder for spawned backends to opt-in to `pk_backend_job_details_full()` compared to compiled backends. It may be easier to break the API and fold `pk_backend_job_details_full()` into `pk_backend_job_details()`.

sidt4:
> We are using this patch for our out of tree backend: https://github.com/getsolus/PackageKit/commit/fc7579d760ad7632c234ef9312fb8ac35bcc63fb. This then allows the eopkg backend to report the download-size such that G-S can pick it up.

Got it.

> Obviously, the problem with this is it breaks all existing python spawned backends. With the way spawned backends is architectured it seems harder for spawned backends to opt-in to pk_backend_job_details_full() compared to compiled backends. It may be easier to break the API and fold pk_backend_job_details_full() into pk_backend_job_details().

Breaking API is not good practice. Major API breaks are planned in `PackageKit 2.0` (which might happen at some point in future). That will be Matt's call though.

I think once I submit fix for https://github.com/PackageKit/PackageKit/issues/781 ( with added `/src/pk-backend-spawn.c` now ), you can use it with `eopkg` backend. There would be no need to break API.

sidt4:
@ximion: Does the patch in https://github.com/getsolus/PackageKit/commit/fc7579d760ad7632c234ef9312fb8ac35bcc63fb sound fine to you?

I'm not sure how many out of tree backends exist for `PackageKit`. But, if we want to play it safe, the following solution might be backward compatible.

For `get-detail` response:

if array size is 8, we can use the existing code `pk_backend_job_details (..., size,)`).

[...]

PR #780, opened 2024-08-14
apt: Use 'pk_backend_job_details_full()' so download size is reported properly

sidt4:
This is needed for download size reporting to work in PK clients which use the `download-size` property (introduced in 3741c6842b) in `get-details` response. `download-size` property is available only when the PK backend uses `pk_backend_job_details_full()`, not otherwise.

Fixes: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1026076

I'll submit changes for other backends in separate pull requests.

### DIFF
--- a/backends/apt/apt-job.cpp
+++ b/backends/apt/apt-job.cpp
@@ -825,14 +825,15 @@ void AptJob::emitPackageDetail(const pkgCache::VerIterator &ver)
     }
 
     g_autofree gchar *package_id = m_cache->buildPackageId(ver);
-    pk_backend_job_details(m_job,
-                           package_id,
-                           m_cache->getShortDescription(ver).c_str(),
-                           "unknown",
-                           get_enum_group(section),
-                           m_cache->getLongDescriptionParsed(ver).c_str(),
-                           rec.Homepage().c_str(),
-                           size);
+    pk_backend_job_details_full (m_job,
+                                 package_id,
+                                 m_cache->getShortDescription(ver).c_str(),
+                                 "unknown",
+                                 get_enum_group(section),
+                                 m_cache->getLongDescriptionParsed(ver).c_str(),
+                                 rec.Homepage().c_str(),
+                                 ver->InstalledSize,
+                                 ver->Size);
 }
 
 void AptJob::emitDetails(PkgList &pkgs)


sidt4:
Opened issue for pending backends - https://github.com/PackageKit/PackageKit/issues/781

sidt4:
Size details are for GNOME Builder app in GNOME Software.

### Before fix:

bug-download
bug-installed

### After fix:

fixed-download
fixed-installed

sidt4:
```
$ apt show gnome-builder | grep Size
Installed-Size: 12.4 MB
Download-Size: 2,560 kB
```

mcrha:
> I guess [3741c68](https://github.com/PackageKit/PackageKit/commit/3741c6842b676d33e235a43098a4647cae84d4e1) broke the size reporting for backends other than `dnf` in GNOME Software.

That change could not break anything, because it adds an argument which was not present in the previous versions, thus nothing knew the download size, because this particular call/API did not have the value before that change.

sidt4:
IIUC  `pk_backend_job_details_full ()` was added and used in PK `dnf` backend which was fine. 

But, https://github.com/GNOME/gnome-software/commit/a2789a905 made changes to GNOME Software to supported only `pk_backend_job_details_full ()`  and not `pk_backend_job_details ()` (which was still used by PK backends other than `dnf`). So, even https://github.com/PackageKit/PackageKit/issues/781 is fixed by switching to `pk_backend_job_details_full ()`, all PK backends should work fine in GNOME Software, and the fix should not affect other PK clients (since they aren't aware of the extra `DBus` argument).

Does that sound right ?

mcrha:
Partly.

I hope we agree the PackageKit change you referenced and blamed at the description did not break anything. You corrected yourself that the change on the gnome-software side changed the things, which I agree with. Specifically when gnome-software knows the new object property is available it expects the PackageKit reports the download size properly, but, as you figured, not all backends do that. The old backends did not have any other option than to report only one size.

A workaround would be, if the old API is used, to set both sizes the same, but that's ugly and can be misleading. Or the old API could set some sort of "the size is unknown" in the download size property and then the gnome-software (and all other clients) would need to count with it, which is not great.

The best is to avoid the use of the old API on the PackageKit's backend side, as you suggested. It should be done at the beginning, but it was not, because I could not test with other than the dnf PackageKit backend.

sidt4:
> I hope we agree the PackageKit change you referenced and blamed at the description did not break anything. You corrected yourself that the change on the gnome-software side changed the things, which I agree with. 

It was a guess (which was not accurate). I've updated the description.

sidt4:
> The best is to avoid the use of the old API on the PackageKit's backend side, as you suggested. 

Agree.

> It should be done at the beginning, but it was not, because I could not test with other than the dnf PackageKit backend.

If it doesn't impact PK clients (since `download-size` is a new property addition) as mentioned above, then we can always switch to `pk_backend_job_details_full ()` for all PK backends (even without testing individual PK backends). Right ?

sidt4:
Any reason why `pk_backend_job_details_full ()` was added instead of modifying `pk_backend_job_details ()` considering that it's not public API ?

mcrha:
It's a public API for the backends. I've no idea whether there are any 3rd-party PackageKti backends, but being on the safe side is a good thing in general.

For the "not using the new API in all builtin backends", it did not feel right to do so, due to lack of testing. The change is straightforward, sort of, but it could provide inaccurate information if done wrong, which is easy to do without testing.

sidt4:
> It's a public API for the backends. I've no idea whether there are any 3rd-party PackageKti backends, but being on the safe side is a good thing in general.

Interesting point. 

@ximion: should we maintain `pk_backend_job_details_full ()` / `pk_backend_job_details ()` or merge them into one. There are backends which [fork PackageKit and use oot backends](https://github.com/PackageKit/PackageKit/issues/771), which I guess are rare. I will submit PR to https://github.com/PackageKit/PackageKit/issues/771 accordingly.

> For the "not using the new API in all builtin backends", it did not feel right to do so, due to lack of testing. The change is straightforward, sort of, but it could provide inaccurate information if done wrong, which is easy to do without testing.

Fair enough.

sidt4:
I'll try to test this change on all (`-dnf`) backends, and update this MR.

Based on https://github.com/PackageKit/PackageKit/pull/780#issuecomment-2315175264, this change should not affect any PK clients which depend on the output of `pk_backend_job_details ()`.

sidt4:
> I'll try to test this change on all (-dnf) backends, and update this MR.

Will take a look sometime this week.

sidt4:
> @ ximion: should we maintain pk_backend_job_details_full () / pk_backend_job_details () or merge them into one. There are backends which https://github.com/PackageKit/PackageKit/issues/771, which I guess are rare. I will submit PR to https://github.com/PackageKit/PackageKit/issues/771 accordingly.

@ximion: ping.

sidt4:
I can work on this if there is some clarification. Thanks!

ximion:
First of all, thanks for the patch, this change is fine (and sorry for the delay!). Changing backend API is generally easier as all backends are maintained in-tree, but every backend has their own maintainer and if I make any global change breaking API I'd have to test all backends to make sure they work alright.
That's why the two methods exist, it's for safety and the keep the backend API stability promise.

Let's discuss changing that on the other referenced issues, that's the better place for it :-)

PR #786, opened 2024-08-23
Fix zypp compiler errors and warnings

sidt4:
Refer commit log.

Closes #692

PR #787, opened 2024-08-24
Add alpm and zypp backends to Fedora CI

sidt4:
Refer commit logs.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -25,6 +25,11 @@ jobs:
              ./tests/ci/build-and-test.sh -Dpackaging_backend=dnf
         if: ${{ matrix.distro == 'fedora' }}
 
+      - name: Build & Test [ALPM]
+        run: podman run -t -v `pwd`:/build packagekit-${{ matrix.distro }}
+             ./tests/ci/build-and-test.sh -Dpackaging_backend=alpm
+        if: ${{ matrix.distro == 'fedora' }}
+
       - name: Build & Test [APT]
         run: podman run -t -v `pwd`:/build packagekit-${{ matrix.distro }}
              ./tests/ci/build-and-test.sh -Dpackaging_backend=apt
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -30,6 +30,11 @@ jobs:
              ./tests/ci/build-and-test.sh -Dpackaging_backend=alpm
         if: ${{ matrix.distro == 'fedora' }}
 
+      - name: Build & Test [ZYPP]
+        run: podman run -t -v `pwd`:/build packagekit-${{ matrix.distro }}
+             ./tests/ci/build-and-test.sh -Dpackaging_backend=zypp
+        if: ${{ matrix.distro == 'fedora' }}
+
       - name: Build & Test [APT]
         run: podman run -t -v `pwd`:/build packagekit-${{ matrix.distro }}
              ./tests/ci/build-and-test.sh -Dpackaging_backend=apt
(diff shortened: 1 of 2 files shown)


sidt4:
Failing zypp  job https://github.com/sidt4/PackageKit/actions/runs/10537898788/job/29199803510 ( as expected ). Failures should be fixed in https://github.com/PackageKit/PackageKit/pull/786.

Passing zypp job https://github.com/sidt4/PackageKit/actions/runs/10537935119/job/29199883815 (after temporarily applying patch from https://github.com/PackageKit/PackageKit/pull/786).

sidt4:
Depends on https://github.com/PackageKit/PackageKit/pull/786 and https://github.com/PackageKit/PackageKit/pull/789.

sidt4:
We can possibly supply multiple backend choices as `-Dpackaging_backend=alpm,apt,zypp`, but one would have to look into the logs every time to check which backend failed (when there is one).

Also, each backend job `[APT, ZYPP, ALPM]` runs for 1.5 minutes. So, I guess separate options would be cleaner to debug.

But, if we want to save time & resources, we can go with multiplexing backends.

Conan-Kudo:
We probably don't want to do this on Debian since they would be very old versions. Fedora would probably be a better choice here.

Conan-Kudo:
Also, I don't really want to add CI jobs for backends that people aren't actively maintaining. ALPM is actively maintained, but we don't really get much from the Zypp folks.

sidt4:
>  Fedora would probably be a better choice here.

This MR upgrades from `bookworm` / 12 to `trixie` / `testing`. Does that work ?

sidt4:
> ALPM is actively maintained, but we don't really get much from the Zypp folks.

From what I see, Zypp is being actively maintained by @JonathanKang from SUSE.

sidt4:
> We probably don't want to do this on Debian since they would be very old versions. Fedora would probably be a better choice here.

I agree Fedora is a better choice here (as alpm is not recent enough in Debian)

### Fedora

- [libzypp](https://src.fedoraproject.org/rpms/libzypp)
- [pacman](https://src.fedoraproject.org/rpms/pacman)

### Debian

- [libzypp-dev](https://packages.debian.org/search?keywords=libzypp-dev&searchon=names&suite=all&section=all)
- [pacman-package-manager](https://packages.debian.org/search?keywords=pacman-package-manager&searchon=names&suite=all&section=all)

Should we use `Fedora 40` ?

sidt4:
> Also, I don't really want to add CI jobs for backends that people aren't actively maintaining. ALPM is actively maintained, but we don't really get much from the Zypp folks.

One other reason for having `zypp` in CI is that most backends are in python with PK wrappers in C (so we can't any way test them on non-native distros live Fedora / Debian, IIUC), while zypp is C++. Refer https://github.com/PackageKit/PackageKit/pull/786.

ximion:
Why don#t we just test those backends on their native distributions, so, zypp in OpenSUSE?

ximion:
Why don't we just test those backends on their native distributions, so, zypp in OpenSUSE?

sidt4:
Sure, that would be the right way to go.

My suggestions were to get maximum CI coverage with minimal time and effort, given the resource constraints in PackageKit.

sidt4:
I just checked the links in https://github.com/PackageKit/PackageKit/pull/787#issuecomment-2408333983. It appears that Debian keeps alpm / zypp versions in sync with Fedora (infact Debian has more recent zypp than Fedora as of today).

Conan-Kudo:
I would prefer to use Fedora for this as I maintain most of the alternative package managers there.

sidt4:
> I would prefer to use Fedora for this as I maintain most of the alternative package managers there.

Moved to Fedora. Thanks!

Conan-Kudo (approved):

PR #788, opened 2024-08-24
Decouple offline_update and systemd Meson features

arrowd:


arrowd:
Bump, can we get this in?

meson.build
```diff
@@ -37,16 +37,14 @@ if get_option('systemd')
...
-    error('Offline updates requires Systemd, use -Dsystemd=true to enable it or -Doffline_update=false to disable offline updates')
-  endif
+  add_project_arguments ('-DHAVE_SYSTEMD_SD_JOURNAL_H=1', language: 'c')
 endif
 
 elogind = []
 if get_option('elogind')
   elogind = dependency('elogind', version: '>=229.4')
   add_project_arguments ('-DHAVE_SYSTEMD_SD_LOGIN_H=1', language: 'c')
+  add_project_arguments ('-DHAVE_SYSTEMD_SD_JOURNAL_H=1', language: 'c')
```
Conan-Kudo, line 47: But elogind doesn't provide sd-journal, so why would we do that?

Conan-Kudo (changes requested):

meson.build
arrowd: Oh, it doesn't? I had no idea, I will fix that.

Conan-Kudo: Hmm, it seems it does? https://github.com/elogind/elogind/tree/main/src/libelogind/sd-journal

arrowd: image

client/pk-offline-update.c
```diff
@@ -32,7 +32,13 @@
 #include <packagekit-glib2/pk-offline-private.h>
 #include <stdlib.h>
 #include <unistd.h>
+#ifdef HAVE_SYSTEMD_SD_JOURNAL_H
 #include <systemd/sd-journal.h>
+#else
+#define LOG_INFO STDERR_FILENO
+#define LOG_WARNING STDERR_FILENO
+#define sd_journal_print dprintf
+#endif
```
Conan-Kudo, line 41: Does this actually work? This seems a bit sketchy...

arrowd: Yes, it prints correctly here on FreeBSD.

Conan-Kudo: Huh, okay...

Conan-Kudo (approved):

PR #789, opened 2024-08-24
Fixes alpm compiler errors (supports alpm 13 and 14)

sidt4:
Refer commit log.


sidt4:
Mostly addressed in https://github.com/PackageKit/PackageKit/pull/793. Closing.

Issue #659, opened 2023-08-24
Include previous version in /var/lib/PackageKit/offline-update-competed


mcrha:
Just something to think of. Feel free to reject it.

Currently, the `/var/lib/PackageKit/offline-update-competed` contains versions of the packages which had been installed, not versions of the packages, which  had been updated, aka the version before the update. It would be good to know the previous versions, thus one can show to the user what change had been done (like in the GNOME Software). Getting the new installed version is easy, it's just the installed package in the system.

I can understand other software may depend on the new versions, thus maybe add a new key into the file with the same list as in the `Packages` key, only with the previous versions. Something like `UpdatedPackages` or any better name.

ximion:
Maybe that's an oportunity to create a new file and fix that typo too, without breaking compatibility... :-P

sidt4:
Is it possible to move away from `/var/lib/PackageKit/offline-update-competed ` and maintain the details in transaction.db (maybe, new `offline-updates`table), and abstract the query via `libpackagekit-glib2` API ?

With this, we can add new features (new tables / columns etc) without breaking backward compatibility in future. We can also make the `offline-updates` include previous updates and have `boot-index` column (similar to `journalctl`), so PK clients (like GNOME Software etc) can have a complete `Update History` feature in their UI as well.

PR #795, opened 2024-09-24
Use '|' as package-id separator when writing offline updates trigger file

arrowd:
The ',' separator is unsuitable because FreeBSD packages may contain it in version strings.

### DIFF
--- a/docs/offline-updates.txt
+++ b/docs/offline-updates.txt
@@ -47,7 +47,7 @@ For success, the file will contain the following:
 [PackageKit Offline Update Results]
 Success=true
 Role=update-packages
-Packages=gimp;2.10.34-1.fc38;x86_64;updates,bash;5.2.15-3.fc38;x86_64;updates
+Packages=gimp;2.10.34-1.fc38;x86_64;updates|bash;5.2.15-3.fc38;x86_64;updates
 """
 
 ________________________________________________________________________
--- a/lib/packagekit-glib2/pk-offline-private.c
+++ b/lib/packagekit-glib2/pk-offline-private.c
@@ -310,7 +310,7 @@ pk_offline_auth_set_prepared_ids (gchar **package_ids, GError **error)
 
 	g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
 
-	data = g_strjoinv (",", package_ids);
+	data = g_strjoinv ("|", package_ids);
 	keyfile = g_key_file_new ();
 	g_key_file_set_string (keyfile, "update", "prepared_ids", data);
 	return g_key_file_save_to_file (keyfile, PK_OFFLINE_PREPARED_FILENAME, error);
@@ -475,7 +475,7 @@ pk_offline_auth_set_results (PkResults *results, GError **error)
 
 				/* deduplicate entries in case the backend has emitted them multiple times */
 				if (g_hash_table_add (known_pkgids, (gpointer) pkgid))
-					g_string_append_printf (string, "%s,", pkgid);
+					g_string_append_printf (string, "%s|", pkgid);
 				break;
 			default:
 				break;
--- a/lib/packagekit-glib2/pk-offline.c
+++ b/lib/packagekit-glib2/pk-offline.c
@@ -483,7 +483,7 @@ pk_offline_get_prepared_ids (GError **error)
 		return NULL;
 
 	/* return raw package ids */
-	return g_strsplit (prepared_ids, ",", -1);
+	return g_strsplit (prepared_ids, "|", -1);
 }
 
 /**
@@ -725,7 +725,7 @@ pk_offline_get_results (GError **error)
 	data = g_key_file_get_string (file, PK_OFFLINE_RESULTS_GROUP,
 				      "Packages", NULL);
 	if (data != NULL) {
-		package_ids = g_strsplit (data, ",", -1);
+		package_ids = g_strsplit (data, "|", -1);
 		for (i = 0; package_ids[i] != NULL; i++) {
 			g_autoptr(PkPackage) pkg = NULL;
 			pkg = pk_package_new ();
--- a/lib/packagekit-glib2/pk-test-private.c
+++ b/lib/packagekit-glib2/pk-test-private.c
@@ -623,7 +623,7 @@ pk_test_offline_func (void)
 	const gchar *results_success =
 			"[PackageKit Offline Update Results]\n"
 			"Success=true\n"
-			"Packages=upower;0.9.16-1.fc17;x86_64;updates,"
+			"Packages=upower;0.9.16-1.fc17;x86_64;updates|"
 				 "zif;0.3.0-1.fc17;x86_64;updates\n";
 
 	/* cleanup */


Conan-Kudo:
Uhh, this might be significant. Changing these things is effectively an API break.

@mcrha, @aleixpol: Can you please look at this and see what you think?

arrowd:
How is so? This file is a private implementation detail of the pk-offline machinery.

sidt4:
The filename rename was planned for `PackageKit-2.0` by Matt. I'll link the discussion here.

### Few points:

1. Missing changes for reading the results.

```diff
--- a/lib/packagekit-glib2/pk-offline.c
+++ b/lib/packagekit-glib2/pk-offline.c
@@ -725,7 +725,7 @@ pk_offline_get_results (GError **error)
 	data = g_key_file_get_string (file, PK_OFFLINE_RESULTS_GROUP,
 				      "Packages", NULL);
 	if (data != NULL) {
-		package_ids = g_strsplit (data, ",", -1);
+		package_ids = g_strsplit (data, "|", -1);
 		for (i = 0; package_ids[i] != NULL; i++) {
 			g_autoptr(PkPackage) pkg = NULL;
 			pkg = pk_package_new ();
```

2. This should mostly work, but some PK clients might be accessing the file directly. 

     E.g. `libdiscover` uses the filename. Refer [https://codesearch.debian.net/search?q=offline-update-competed](https://codesearch.debian.net/search?q=offline-update-competed).

3. `docs/offline-updates.txt` needs to be updated as well.

sidt4:
Also, I guess we might end up when the old file (old name + old format) is present, but new `PackageKit` ( with this patch ) is installed. It wouldn't work in that case.

So, we might want to support both filenames and format for a couple of releases.

arrowd:
Ugh, it sounds like we won't get this in quickly.

sidt4:
> The filename rename was planned for PackageKit-2.0 by Matt. I'll link the discussion here.

Refer https://github.com/PackageKit/PackageKit/issues/659.

If we're modifying the current format (`v1`) of the `offline-update-competed` we should define a new format (`v2`) supporting features like https://github.com/PackageKit/PackageKit/issues/659 (so `v2` format holds for a few years) etc, and add backward compatibility to support `v1`.

> gnome-software --mode=updated prints the contents of the offline-update-competed file.

By this I meant `gnome-software` displays the contents of the `offline-update-competed` file via `PackageKit` (and not direct file access as done by `libdiscover`), so we're good here. I added it to showcase the scenario where having old format (`v1`) with new PackageKit (supporting `v2` only) would break this use case.

aleixpol:
Seems like this would break discover. It's not a huge deal as it just goes into the notification but something to keep in mind.

https://invent.kde.org/plasma/discover/-/blob/master/libdiscover/backends/PackageKitBackend/PackageKitNotifier.cpp#L115

If there's going to be a v2, we might as well target that.

mcrha:
The filenme is not exposed in the public API (`/usr/include/PackageKit/...`), thus gnome-software does not reference it. The gnome-software uses the PackageKit API to access it (`pk_offline_get_results_mtime`) and its content (`pk_offline_get_results`). Whatever you do to the file, being it a rename and/or data change format, won't influence the gnome-software by any means, as long as the public PackageKit API will still work the same as before the change.

arrowd:
> Seems like this would break discover. It's not a huge deal as it just goes into the notification but something to keep in mind.
> 
> https://invent.kde.org/plasma/discover/-/blob/master/libdiscover/backends/PackageKitBackend/PackageKitNotifier.cpp#L115

I took a stab on this which quickly lead me to a realization on why Discover had this code in the first place.

The PackageKit code that obtains the result of an offline action does not use D-Bus (intentionally, most likely) and accesses files from `/var/lib/PackageKit/...` directly. This code is present in `packagekit-glib` (pk-offline.c), but not in PackageKit-Qt, which is a D-Bus only bindings library.

To solve this problem we should either
1. Make Discover link to `packagekit-glib`, allowing us to call `pk_offline_get_results()`.
2. Duplicate `pk_offline_get_results()` code in PackageKit-Qt
3. Export `pk_offline_get_results()` from the D-Bus surface somehow and let PackageKit-Qt pick it up.

I can do the work, but need suggestions on how to proceed.

arrowd:
We can pass a pipe fd, from which the client would read the ini- or json- formatted data.

Conan-Kudo:
Because the dbus broker service (dbus-daemon in particular) doesn't necessarily handle that extremely well. It's probably fine with systems that use dbus-broker, but non-Linux systems use dbus-daemon.

arrowd:
> It's probably fine with systems that use dbus-broker, but non-Linux systems use dbus-daemon.

This is exactly my case with FreeBSD. But do you have more details on this problem? This is the first time I hear about that.

Conan-Kudo:
This was outlined in the announcement of dbus-broker: https://dvdhrm.github.io/rethinking-the-dbus-message-bus/

That said, it's possible that PackageKit's messages won't be too bad for even dbus-daemon. I just don't know.

arrowd:
This is a very interesting read, thank you.

Back to the topic, since we're going to wrap this D-Bus method into a C API anyways, let's just pass a read-only file fd? This should address all the concerns.

Conan-Kudo:
Works for me.

sidt4:
Now that https://github.com/PackageKit/PackageKit/pull/821 is merged, I guess `Discover` should be updated to use the DBus API than directly accessing `/var/lib/PackageKit/offline-update-competed` as the next step.

Also, would it make sense to make the package-id separator character configurable (e.g. `meson_options.txt`). We don't want another os having issues with `|` separator in future.

arrowd:
Actually, I was planning to reuse `pk_package_ids_to_string ()` from https://github.com/PackageKit/PackageKit/blob/e9c92a506749b5bde04fa047824f4a87ed9932ff/lib/packagekit-glib2/pk-package-ids.h#L44 that uses `&` symbol.

ximion:
Uhhh, I really don't like any of this. How does FreeBSD use `,` in version strings? The only way to make this somewhat compatible would be to define an escape sequence for `,` - that way, only FreeBSD would be affected by the change (ideally), and consumers of the file would continue to work everywhere else, but would need to adapt their parsing to support FreeBSD.

But if we do change the file in an incompatible way (and/or make it an implementation detail ideally in a future release, ideally), we might as well use a format that natively supports arrays (could be anything, JSON/YAML/TOML/SQLite/LMDB/etc. - but JSON or TOML is likely most sensible). That would sidestep the problem, no escaping and complicated additional rules or compat code needed.

arrowd:
>  How does FreeBSD use , in version strings?

It is called "epoch" and we use it when upstream suddenly changes the versioning scheme in a such way that a new version looks lesser than an old one. Like, after having snapshot versions in form of `2025.05.03` a project makes its first stable release `1.0.0`. But `1.0.0` < `2025.05.03`, so we make it `1.0.0,1` so that it is actually larger.

> The only way to make this somewhat compatible would be to define an escape sequence for , - that way, only FreeBSD would be affected by the change (ideally), and consumers of the file would continue to work everywhere else, but would need to adapt their parsing to support FreeBSD.

I'd argue that using `pk_package_ids_*` there (or in other words, switching to '&' as separator) is a good solution, because:
- We're storing arrays of not some arbitrary data, but package IDs.
- `pk_package_ids_` is the intended way to work with arrays of package IDs including their (de)serialization

Anyways, how do we move this forward now. Are you basically saying "handle this on FreeBSD side yourself" or switch this MR to JSON and then wait unknown amount of time for the next major release?

mcrha:
There is `g_key_file_set_string_list()`

ximion:
The problem is that the `data` part of a package-ID can contain arbitrary data, and I am not sure if some distro isn't using all kinds of characters (except `;`) in there. Since this is an API break, I really wonder whether we shouldn't just write this file as JSON data and avoid all the escaping problems, keep writing the old file in the exact same format for quite a while, and once everyone has migrated to the new file we stop writing the old compat file.

That way we avoid any future issues of this kind as well.

arrowd:
Well, we can modify `pk_package_id_build()` and `pk_package_id_split()` to perform appropriate escaping/unescaping of the data part.

ximion:
@mcrha 
> There is `g_key_file_set_string_list()`

Actually yes! And that one performs the necessary string (un)escaping! - Why aren't we just using this, that would resolve all problems, as far as I can see (and would be backwards-compatible, for the most part at least...)?

mcrha:
No idea why the `g_key_file_set_string_list` is not used, maybe just an overlook. As long as the other file readers use the PacakgeKit API, which they should, it's transparent and easy to change for everyone, no need to modify the PackageKit API at all, neither the internal formats (a test what it does during the transition period, when reading what had been stored before using the `g_key_file_set_string_list` is required). That was my point, even not spoken explicitly.

PR #797, opened 2024-10-01
pk-client: Allow debconf helper to display Wayland windows

aleasto:
Set both XDG_RUNTIME_DIR and WAYLAND_DISPLAY in the helper's environment.

Addittionally set XAUTHORITY which is necessary for connecting to XWayland, in case Wayland wasn't available.

### DIFF
--- a/lib/packagekit-glib2/pk-client.c
+++ b/lib/packagekit-glib2/pk-client.c
@@ -2159,6 +2159,8 @@ pk_client_create_helper_argv_envp (gchar ***argv,
 {
 	const gchar *dialog = NULL;
 	const gchar *display;
+	const gchar *wayland_display;
+	const gchar *xauthority;
 	const gchar *term;
 	gboolean ret;
 	guint envpi = 0;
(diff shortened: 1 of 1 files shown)


aleasto:
@ximion @hughsie You previously dismissed exposing the HOME variable in #609, but I believe that was done with incorrect arguments.

The helper process always runs as a user process on the client side. In this case, graphical helpers need to learn about the Wayland socket for showing graphics in a Wayland session. That also requires exposing the user's XDG_RUNTIME_DIR. It should be perfectly safe as we're just forwarding one variable from user-process to a forked user-process.

I've verified this works in Ubuntu 24.10 (*minus a new regression in 1.3.0 which causes `pkcon` to segfault when finishing up the transaction)

Conan-Kudo (approved):

julian-klode (approved):
This sounds correct to me. Basically yes the helper can be systemd-activated these days, but it's always a user process. It used to be started by the frontend but I moved it to a systemd *user* service in d57588f2b723e36c842c453810f81025285ef2bd

Either way it's a user service :D

PR #800, opened 2024-10-03
pk-task: Restore GTask owner after simulate_question

aleasto:
After simulate_question the only reference to the GTask is in the hash table of the PkTask.
Make the user_{accepted,declined} callbacks a new owner of the GTask, as those functions will eventually unref the object themselves.

Fixes commit 9dfaf0dd653f30edc80ed9a650d674440581cf29

### DIFF
--- a/lib/packagekit-glib2/pk-task.c
+++ b/lib/packagekit-glib2/pk-task.c
@@ -759,7 +759,7 @@ pk_task_user_accepted (PkTask *task, guint request)
 	}
 
 	idle_source = g_idle_source_new ();
-	g_source_set_callback (idle_source, G_SOURCE_FUNC (pk_task_user_accepted_idle_cb), g_object_ref (gtask), g_object_unref);
+	g_source_set_callback (idle_source, G_SOURCE_FUNC (pk_task_user_accepted_idle_cb), g_object_ref (gtask), NULL);
 	g_source_set_name (idle_source, "[PkTask] user-accept");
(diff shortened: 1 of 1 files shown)


aleasto:
@tintou 

Fixes
`GLib-GObject-CRITICAL **: 10:41:15.858: g_object_unref: assertion 'G_IS_OBJECT (object)' failed`
or SIGSEGV in `g_type_check_instance_is_fundamentally_a`
when running, for example, `pkcon install <>`

aleasto:
> Nice catch but that's not the right fix, you'd better change the called function to not take ownership of the given gtask

That's the expected owner when *not* going through simulate_question.
You create a GTask in https://github.com/PackageKit/PackageKit/blob/19432b1a1d9f075e37557cd63d4043bce5b2d371/lib/packagekit-glib2/pk-task.c#L1048 at `ref_count == 1` and pass it to either `pk_task_do_async_action` or `pk_task_do_async_simulate_action` which becomes the owner.
https://github.com/PackageKit/PackageKit/blob/19432b1a1d9f075e37557cd63d4043bce5b2d371/lib/packagekit-glib2/pk-task.c#L1055-L1058
In the latter case though, after the simulate action is finished the ref is dropped, and it's eventually retrieved again from the hash table eventually calling `pk_task_do_async_action` which expects to be owner as if it was called directly after creating the GTask.

tintou:
This is what I meant https://github.com/PackageKit/PackageKit/pull/801

PR #807, opened 2024-11-06
PkClient: Remove the proxy_props field

tintou:
It is not used anywhere.


Conan-Kudo:
Isn't this an API and ABI break?

tintou:
it is not, it is an internal structure

Conan-Kudo (approved):

PR #808, opened 2024-11-06
PkClient: Use g_atomic_int for the copy count

tintou:
Rename refcount to remaining_files_to_copy to avoid mixing reference counts with time to live.

Change the documentation around pk_client_state_finish as it is really not derefencing PkClientState.

Take a reference on PkClientState while copying a file to make sure that it always live long enough.


ximion:
Really nice, thank you!

PR #810, opened 2024-11-13
apt: Fix matching gstreamer pkgs where the only modifier is ()(64bit)

aleasto:
When the list of GstCaps specified is empty, and the only modifier we're given is ()(64bit), our regex eats the outer parenthesis so we're left with ")(64bit" while trying to match "()(64bit".

This change would introduce a trailing comma when the list of specified GstCaps is instead not empty, which gst_caps_from_string today can handle, but for abundance of caution let's avoid it.

Test cases:
```
pkcon what-provides "gstreamer1(decoder-audio/mpeg)(mpegversion=4)()(64bit)"
pkcon what-provides "gstreamer1(decoder-audio/mpeg)(mpegversion=4)"
pkcon what-provides "gstreamer1(decoder-video/x-h265)()(64bit)"
pkcon what-provides "gstreamer1(decoder-video/x-h265)"
```

### DIFF
--- a/backends/apt/gst-matcher.cpp
+++ b/backends/apt/gst-matcher.cpp
@@ -71,17 +71,21 @@ GstMatcher::GstMatcher(gchar **values)
                 if (!opt.empty()) {
                     size_t start_pos = 0;
                     // This is hardcoded in pk-gstreamer-install, so we also hardcode it here
-                    const string x86_64 = "()(64bit";
+                    const string x86_64 = ")(64bit";
 
                     if (ends_with(opt.c_str(), x86_64.c_str())) {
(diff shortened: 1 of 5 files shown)


3v1n0:
Looks good to me. Code seems a bit fragile though (not as per your change), so wondering if we can add some testing to ensure we're not breaking anything now or in the future.

aleasto:
Added tests

ximion:
Very nice, I am incredibly glad for the tests you added! We may need to integrate the tests better in future with the rest of PKs testsuite, but for now this is good to go!

Thank you for the work!

PR #816, opened 2024-12-16
freebsd backend: Make types of {compressed,flat}size() match to what pk_backend_job_details_full() expect

arrowd:


Conan-Kudo:
Can you fix the commit message so the title (first line) is under 60 characters? You can put additional information in the body of the commit message.

arrowd:
Done.

Conan-Kudo (approved):

Issue #709, opened 2024-01-09
pkmon: GTask finalized without ever returning (using g_task_return_*())


sidt4:
```sh
$ pkmon
...
...
(pkmon:4086): GLib-GIO-CRITICAL **: 17:29:46.168: GTask 0x55aa9147cf60 (source object: 0x55aa9147aa30, source tag: 0x7f1b39a3b1c0) finalized without ever returning (using g_task_return_*()). This potentially indicates a bug in the program.
/8_edaceddb	allow_cancel 1
/8_edaceddb	percentage   -1
/8_edaceddb	role         search-file
/8_edaceddb	sender       /usr/bin/gnome-software
/8_edaceddb	status       wait
```

sidt4:
Still getting these, making `pkmon` a bit annoying to use.

```
(pkmon:78935): GLib-GIO-CRITICAL **: 15:57:17.241: GTask 0x55bec43c7bf0 (source object: 0x55bec30a0a30, source tag: 0x7fa2f7dba1c0) finalized without ever returning (using g_task_return_*()). This potentially indicates a bug in the program.

(pkmon:78935): GLib-GIO-CRITICAL **: 15:57:17.241: GTask 0x55bec43ecbe0 (source object: 0x55bec30a0a30, source tag: 0x7fa2f7dba1c0) finalized without ever returning (using g_task_return_*()). This potentially indicates a bug in the program.

(pkmon:78935): GLib-GIO-CRITICAL **: 15:57:17.241: GTask 0x55bec43a1ba0 (source object: 0x55bec30a0a30, source tag: 0x7fa2f7dba1c0) finalized without ever returning (using g_task_return_*()). This potentially indicates a bug in the program.

(pkmon:78935): GLib-GIO-CRITICAL **: 15:57:17.241: GTask 0x55bec4302530 (source object: 0x55bec30a0a30, source tag: 0x7fa2f7dba1c0) finalized without ever returning (using g_task_return_*()). This potentially indicates a bug in the program.
```

 @tintou: can you please take a look ?

tintou:
I've looked at it and is not really a big issue per-se (other than producing a verbose message) but requires a bit refactoring

sidt4:
Some downstream bugs:

- https://bugzilla.opensuse.org/show_bug.cgi?id=1236037
- https://bugzilla.altlinux.org/53178

PR #818, opened 2024-12-20
pk-task: Modernize the codebase

tintou:
No functional change but use the latest GObject coding standards.

### DIFF
--- a/lib/packagekit-glib2/pk-task.c
+++ b/lib/packagekit-glib2/pk-task.c
@@ -36,8 +36,6 @@
 
 static void     pk_task_finalize	(GObject     *object);
 
-#define PK_TASK_GET_PRIVATE(o) (G_TYPE_INSTANCE_GET_PRIVATE ((o), PK_TYPE_TASK, PkTaskPrivate))
-
 #define PK_TASK_TRANSACTION_CANCELLED_RETRY_TIMEOUT	2000 /* ms */
 
 /**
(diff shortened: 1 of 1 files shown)


sidt4:
Thanks for this PR!

I'm not sure, but I guess we should resolve https://github.com/PackageKit/PackageKit/issues/709 before making new code changes in that area.

tintou:
> Does this raise the GLib minimum version?

It doesn't

tintou:
> Thanks for this PR!
> 
> I'm not sure, but I guess we should resolve #709 before making new code changes in that area.

I'll try to look into this but this shouldn't block any change

lib/packagekit-glib2/pk-task.c
```diff
@@ -58,13 +56,15 @@ struct _PkTaskPrivate
 enum {
 	PROP_0,
 	PROP_SIMULATE,
-	PROP_ONLY_PREPARE,
+	PROP_ONLY_DOWNLOAD,
```
Conan-Kudo, line 59: Why was this renamed?

Conan-Kudo (changes requested):

lib/packagekit-glib2/pk-task.c
tintou: The property is actually named "only-download" and the setter/getter are named set/get_only_download so It makes sense to have it consistent here too

tintou: This has no implication on API/ABI

Conan-Kudo (approved):

lib/packagekit-glib2/pk-task.c
Conan-Kudo: I didn't think it did, since you didn't change the order in the enumeration.

PR #821, opened 2025-01-17
Add GetResult{,Packages,Error} methods to the Offline interface

arrowd:
Continuing from #795 , this change adds methods for accessing the `PK_OFFLINE_RESULTS_FILENAME` data.

The `GetResultPackages` method returns a plain strings array because
1. The `PK_OFFLINE_RESULTS_FILENAME` contents is still a private implementation detail of the `packagekit-glib` library.
2. The `GetPrepared` method already returns a string array, which might be as large as the resulting array.

### DIFF
--- a/src/org.freedesktop.PackageKit.xml
+++ b/src/org.freedesktop.PackageKit.xml
@@ -592,6 +592,75 @@
     </property>
 
     <!--*********************************************************************-->
+    <method name="GetResult">
+      <doc:doc>
+        <doc:description>
+          <doc:para>
+            Returns if the last offline action was completed successfully.
+          </doc:para>
+        </doc:description>
+      </doc:doc>
+      <arg type="b" name="success" direction="out">
+        <doc:doc>
+          <doc:summary>
+            <doc:para>
+              A boolean value representing success.
+            </doc:para>
+          </doc:summary>
+        </doc:doc>
+      </arg>
+    </method>
+
+    <!--*********************************************************************-->
+    <method name="GetResultPackages">
+      <doc:doc>
+        <doc:description>
+          <doc:para>
+            Returns the list of packages that were updated during the last offline action.
+          </doc:para>
+        </doc:description>
+      </doc:doc>
+      <arg type="as" name="package_ids" direction="out">
+        <doc:doc>
+          <doc:summary>
+            <doc:para>
+              An array of package IDs.
+            </doc:para>
+          </doc:summary>
+        </doc:doc>
+      </arg>
+    </method>
+
+    <!--*********************************************************************-->
+    <method name="GetResultError">
+      <doc:doc>
+        <doc:description>
+          <doc:para>
+            Returns the error if the last offline action has failed.
+          </doc:para>
+        </doc:description>
+      </doc:doc>
+      <arg type="s" name="error_code" direction="out">
+        <doc:doc>
+          <doc:summary>
+            <doc:para>
+              An stringified PackageKit error code.
+            </doc:para>
+          </doc:summary>
+        </doc:doc>
+      </arg>
+      <arg type="s" name="error_description" direction="out">
+        <doc:doc>
+          <doc:summary>
+            <doc:para>
+              A detailed error description.
+            </doc:para>
+          </doc:summary>
+        </doc:doc>
+      </arg>
+    </method>
+
+    <!--*********************************************************************-->
     <method name="ClearResults">
       <doc:doc>
         <doc:description>
--- a/src/pk-engine.c
+++ b/src/pk-engine.c
@@ -1755,6 +1755,63 @@ pk_engine_offline_method_call (GDBusConnection *connection_, const gchar *sender
 		g_dbus_method_invocation_return_value (invocation, value);
 		return;
 	}
+	if (g_strcmp0 (method_name, "GetResult") == 0) {
+		g_autoptr(PkResults) results = NULL;
+
+		results = pk_offline_get_results (&error);
+		if (results == NULL) {
+			g_dbus_method_invocation_return_error (invocation,
+							       PK_ENGINE_ERROR,
+							       PK_ENGINE_ERROR_INVALID_STATE,
+							       "%s", error->message);
+			return;
+		}
+
+		g_dbus_method_invocation_return_value (invocation, g_variant_new ("(b)", pk_results_get_exit_code (results) == PK_EXIT_ENUM_SUCCESS));
+		return;
+	}
+	if (g_strcmp0 (method_name, "GetResultPackages") == 0) {
+		g_autoptr(PkResults) results = NULL;
+		GVariant *value = NULL;
+
+		results = pk_offline_get_results (&error);
+		if (results == NULL) {
+			g_dbus_method_invocation_return_error (invocation,
+							       PK_ENGINE_ERROR,
+							       PK_ENGINE_ERROR_INVALID_STATE,
+							       "%s", error->message);
+			return;
+		}
+
+		value = g_variant_new ("(^as)", pk_package_sack_get_ids (pk_results_get_package_sack (results)));
+
+		g_dbus_method_invocation_return_value (invocation, value);
+		return;
+	}
+	if (g_strcmp0 (method_name, "GetResultError") == 0) {
+		g_autoptr(PkResults) results = NULL;
+		g_autoptr(PkError) pk_error = NULL;
+		GVariant *value = NULL;
+
+		results = pk_offline_get_results (&error);
+		if (results == NULL) {
+			g_dbus_method_invocation_return_error (invocation,
+							       PK_ENGINE_ERROR,
+							       PK_ENGINE_ERROR_INVALID_STATE,
+							       "%s", error->message);
+			return;
+		}
+
+		pk_error = pk_results_get_error_code (results);
+
+		if (pk_error != NULL) {
+			value = g_variant_new ("(ss)", pk_error_enum_to_string (pk_error_get_code (pk_error)), pk_error_get_details (pk_error));
+		} else {
+			value = g_variant_new ("(ss)", "", "");
+		}
+		g_dbus_method_invocation_return_value (invocation, value);
+		return;
+	}
 }
 
 static void


Conan-Kudo (approved):

ximion:
I generally like this, because it allows some future improvements (and encoding the specific file path as basically an element of the API was always odd, even though I wouldn't be at all opposed to mandating the -glib interface as our API (but that hasn't been done traditionally)).

Please give me a little bit of time to think through the ramifications and review the API additions properly though - my initial opinion of this change is very positive though :-)

arrowd:
If there is a high chance for this change getting in, maybe I should start adapting PackageKitQt already while waiting on this PR to get merged?

Conan-Kudo:
Yes, it'd be a good idea to do that.

ximion (commented):
Using three DBus methods feels pretty excessive to me, especially since they all return the same state, just different aspects of it. Why not make one `GetResults` method that returns the success value, packages (if any) and error (if any) all at once?

arrowd:
I removed only `GetResult`, but preserved `GetResultPackage` and `GetResultError`, because I see no way to convey the PackageKit error code and description via D-Bus error mechanism. Or maybe you want me to make them an output arguments?

ximion:
My preference would be to have those as output arguments.

I was briefly thinking that it could make sense to actually have this as Transaction, so we can just reuse the existing machinery of a Transaction to emit packages, error information and everything else.
However, there's also a good point to be made to just have this in the offline-updates interface, as it's fairly logical to put it there.

src/org.freedesktop.PackageKit.xml
```diff
@@ -591,6 +591,46 @@
       </doc:doc>
     </property>
 
+    <!--*********************************************************************-->
+    <method name="GetResults">
+      <doc:doc>
+        <doc:description>
+          <doc:para>
+            If the last offline action was completed successfully, returns a list
+            of packages that were actioned on. Otherwise return an error code and
```
ximion, line 600:
```suggestion
            of packages that were processed. Otherwise returns an error code and
```

src/org.freedesktop.PackageKit.xml
```diff
@@ -591,6 +591,46 @@
...
+      <arg type="as" name="package_ids" direction="out">
+        <doc:doc>
+          <doc:summary>
+            <doc:para>
+              An array of package IDs.
+            </doc:para>
+          </doc:summary>
+        </doc:doc>
+      </arg>
+      <arg type="s" name="error_code" direction="out">
```
ximion, line 614: This should be `u`, we should return the actual error code here. We should also explicitly state how the "no error" condition looks like if we use an unsigned integer  (which could be returning `PK_ERROR_ENUM_UNKNOWN` or MAXUINT).
TBH, I think in this case the most straightforward thing to do would be to also return a "success" boolean that is false if an error is set, and write in the documentation that error code and description are only valid if success is false :-)

ximion (changes requested):

arrowd:
Ping, can we get this in?

src/pk-engine.c
```diff
@@ -1755,6 +1755,38 @@ pk_engine_offline_method_call (GDBusConnection *connection_, const gchar *sender
...
+		pk_error = pk_results_get_error_code (results);
+		success = pk_error == NULL;
+		if (success)
+			pk_error = pk_error_new ();
+		else
+			err_descr = pk_error_get_details (pk_error);
+
+		value = g_variant_new ("(b^asus)",
+				       success,
+				       pk_package_sack_get_ids (pk_results_get_package_sack (results)),
```
ximion, line 1783: You are leaking memory here, the returned PkPackageSack is never free'd...

src/pk-engine.c
```diff
@@ -1755,6 +1755,38 @@ pk_engine_offline_method_call (GDBusConnection *connection_, const gchar *sender
...
 	}
+	if (g_strcmp0 (method_name, "GetResults") == 0) {
+		g_autoptr(PkResults) results = NULL;
+		gboolean success = FALSE;
+		g_autoptr(PkError) pk_error = NULL;
+		const gchar *err_descr = "";
+		GVariant *value = NULL;
+
+		results = pk_offline_get_results (&error);
+		if (results == NULL || pk_results_get_exit_code (results) != PK_EXIT_ENUM_SUCCESS) {
```
ximion, line 1766:
```suggestion
		if (results == NULL) {
```
This should probably only throw an error if there was one returned...

src/pk-engine.c
```diff
@@ -1755,6 +1755,38 @@ pk_engine_offline_method_call (GDBusConnection *connection_, const gchar *sender
...
+			g_dbus_method_invocation_return_error (invocation,
+							       PK_ENGINE_ERROR,
+							       PK_ENGINE_ERROR_INVALID_STATE,
+							       "%s", error->message);
+			return;
+		}
+
+		pk_error = pk_results_get_error_code (results);
+		success = pk_error == NULL;
```
ximion, line 1775:
```suggestion
		success = pk_results_get_exit_code (results) == PK_EXIT_ENUM_SUCCESS;
```

src/pk-engine.c
```diff
@@ -1755,6 +1755,38 @@ pk_engine_offline_method_call (GDBusConnection *connection_, const gchar *sender
...
+		results = pk_offline_get_results (&error);
+		if (results == NULL || pk_results_get_exit_code (results) != PK_EXIT_ENUM_SUCCESS) {
+			g_dbus_method_invocation_return_error (invocation,
+							       PK_ENGINE_ERROR,
+							       PK_ENGINE_ERROR_INVALID_STATE,
+							       "%s", error->message);
+			return;
+		}
+
+		pk_error = pk_results_get_error_code (results);
```
ximion, line 1774: Check if "success" is set as below, and if "success" is FALSE, do this:

```c
PkErrorEnum error_code = PK_ERROR_ENUM_UNKNOWN;
g_autofree gchar *error_msg = NULL;
g_autoptr(PkError) pk_error = NULL;
pk_error = pk_results_get_error_code (results);
if (pk_error == NULL) {
        error_code = PK_ERROR_ENUM_UNKNOWN
	error_msg = g_strdup("Offline update failed without error_code set");
} else {
        error_code = pk_error_get_code (pk_error),
	error_msg = g_strdup( pk_error_get_details (error_code));
}
```
And then return `success`, `error_code` and `error_msg` from the DBus call. This will make sure we handle even odd error corner cases.

ximion (changes requested):
Unfortunately, there's a few more things...
Also, I think this API should return the `Role` of the results as well, to be future proof, so the result of `pk_results_get_role()` on the results object as enum.

I really wonder whether we should make this into a transaction, that would make it easier to extend in future, and we could just reuse existing signals instead of making this a specialty method. But this may be pragmatic and good enough...

I think this DBus call should also return the file's mtime via `pk_offline_get_results_mtime` or there should be a separate call which does that, as frontends will want that value as well and will usually call both `pk_offline_get_results_mtime` and `pk_offline_get_results` back to back.

arrowd:
> I think this DBus call should also return the file's mtime via pk_offline_get_results_mtime or there should be a separate call which does that, as frontends will want that value as well and will usually call both pk_offline_get_results_mtime and pk_offline_get_results back to back.

Hm, I started this PR because of Plasma Discover and it doesn't seem to be interested in mtime. I added it to the API anyways.

src/pk-engine.c
```diff
@@ -1755,6 +1755,50 @@ pk_engine_offline_method_call (GDBusConnection *connection_, const gchar *sender
...
+		success = pk_results_get_exit_code (results) == PK_EXIT_ENUM_SUCCESS;
+		if (!success) {
+			pk_error = pk_results_get_error_code (results);
+			if (pk_error == NULL) {
+				error_code = PK_ERROR_ENUM_UNKNOWN;
+				error_msg = g_strdup("Offline update failed without error_code set");
+			} else {
+				error_code = pk_error_get_code (pk_error),
+				error_msg = g_strdup( pk_error_get_details (pk_error));
```
ximion, line 1785:
```suggestion
				error_msg = g_strdup (pk_error_get_details (pk_error));
```

src/pk-engine.c
```diff
@@ -1755,6 +1755,50 @@ pk_engine_offline_method_call (GDBusConnection *connection_, const gchar *sender
...
+			return;
+		}
+
+		pk_error = pk_results_get_error_code (results);
+		success = pk_results_get_exit_code (results) == PK_EXIT_ENUM_SUCCESS;
+		if (!success) {
+			pk_error = pk_results_get_error_code (results);
+			if (pk_error == NULL) {
+				error_code = PK_ERROR_ENUM_UNKNOWN;
+				error_msg = g_strdup("Offline update failed without error_code set");
```
ximion, line 1782:
```suggestion
				error_msg = g_strdup ("Offline update failed without error_code set");
```

ximion (changes requested):
Just some really minor style nitpicks, and then this is good to go I think :-)

src/pk-engine.c
```diff
@@ -1755,6 +1755,50 @@ pk_engine_offline_method_call (GDBusConnection *connection_, const gchar *sender
...
+
+		results_sack = pk_results_get_package_sack (results);
+
+		value = g_variant_new ("(b^asutus)",
+				       success,
+				       pk_package_sack_get_ids (results_sack),
+				       pk_results_get_role (results),
+				       pk_offline_get_results_mtime (&error),
+				       error_code,
+				       error_msg);
```
ximion, line 1797: Actually, you can not pass NULL to a `s` parameter in DBus. You most likely have to add a
```c
if (error_msg == NULL)
	error_msg = g_strdup ("");
``` 
block before constructing the variant, to avoid odd behavior.

arrowd: Sorry for that silly mistake. My experience with GDBus is somewhere near zero.

arrowd:
Looks correct:

```
# cat /var/lib/PackageKit/offline-update-competed 
[PackageKit Offline Update Results]
Success=true
Role=update-packages
Packages=atf;0.23;amd64;FreeBSD
```

```
# qdbus-qt5 --system --literal org.freedesktop.PackageKit /org/freedesktop/PackageKit org.freedesktop.PackageKit.Offline.GetResults
true
{"atf;0.23;amd64;FreeBSD"}
22
1744877464
0
""
```

PR #823, opened 2025-02-05
apt: Use shared_module instead of shared_library as this is a loaded module

tintou:
We are not exactly creating a library here.


ximion:
It shouldn't change any behavior, but is indeed a little more correct semantically.
Thanks!

PR #824, opened 2025-02-15
apt: Adjust for libapt-pkg7.0 API

julian-klode:
These functions return std::string_view now; adjust the callers to wrap them in std::string{}. An alternative would be to use append() instead of a temporary string object but this is not performance critical anyway.

I'm cherry-picking this in the package, I already committed the patch in salsa, and will upload once APT transition starts and it has been built everywhere.


ximion:
Oh, very nice - I was looking into that yesterday but didn't have the time yet to work on it, thank you for taking care! :-)

The patch looks fine, "append" may indeed be cleaner, but this works and should be compatible in all directions.
Thanks!

PR #829, opened 2025-03-08
apt: Add initial support for deb822

### LINKED ISSUE #763: PackageKit not parsing *.sources files
Upgraded to Kubuntu 24.04 LTS this morning from Kubuntu 23.10. In Kubuntu 24.04, Ubuntu changed from using /etc/apt/sources.list for its software repositories data, to /etc/apt/sources.list.d/ubuntu.sources. There is a sources.list file in /etc/apt but it just contains:

# Ubuntu sources have moved to /etc/apt/sources.list.d/ubuntu.sources

Launching Discover and clicking on Settings now doesn't show any of the Ubuntu software sources as it did, and the button to access Software Sources is also missing, so no way, through Discover, to access/manage Software Sources.

I copied the ubuntu.sources file to /etc/apt to see if Discover would again, display Ubuntu Software Sources data, and present the Software Sources button. Doing so does not.

I was told by Harold Sitter to file a bug against PackageKit (I had filed a bug against Discover on bugs.kde.org (https://bugs.kde.org/show_bug.cgi?id=487453).

tintou:
This is the new format to specify Debian source list.
Closes #763

### DIFF
--- a/backends/apt/apt-sourceslist.cpp
+++ b/backends/apt/apt-sourceslist.cpp
@@ -34,12 +34,26 @@
 #include <apt-pkg/sourcelist.h>
 #include <apt-pkg/strutl.h>
 #include <apt-pkg/error.h>
+#include <apt-pkg/tagfile.h>
+#include <apt-pkg/indexfile.h>
+#include <apt-pkg/metaindex.h>
 #include <algorithm>
 #include <fstream>
+#include <fcntl.h>
 
 #include "apt-utils.h"
 
 #include "config.h"
+
+static std::vector<std::string> FindMultiValue(pkgTagSection &Tags, char const *const Field)
+{
+   auto values = Tags.FindS(Field);
+   // we ignore duplicate spaces by removing empty values
+   std::replace_if(values.begin(), values.end(), isspace_ascii, ' ');
+   auto vect = VectorizeString(values, ' ');
+   vect.erase(std::remove_if(vect.begin(), vect.end(), [](std::string const &s) { return s.empty(); }), vect.end());
+   return vect;
+}
 
 SourcesList::~SourcesList()
 {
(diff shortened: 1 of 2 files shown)


sidt4:
Great!

I'll give it some testing when I find time. Thanks!

ximion:
Nice! Is there a reason why you wouldn't use `pkgTagFile` from APT for parsing? I'd prefer that if possible, so we don't duplicate parsing logic...

sidt4:
**Quick question:**

Can't we use the sources parser provided by `libapt` than manually parsing ourselves. 

Refer: https://salsa.debian.org/apt-team/apt/-/blob/main/apt-pkg/sourcelist.h

sidt4:
Looks like Matthias already commented on this few hours ago (which I see in my email), but not in this PR.

**Matthias wrote:**

_Nice! Is there a reason why you wouldn't use pkgTagFile from APT for parsing? I'd prefer that if possible, so we don't duplicate parsing logic..._

tintou:
I first tried with pkgSourceList and gave up, using pkgTagFile should be fine then :slightly_smiling_face: I wish there was a class that would just parse a file for us directly though :sweat_smile:

tintou:
Here is the new version that uses `pkgTagFile` I wish we could reuse more parsing logic from libapt-pkg but that's not possible

sidt4:
@ximion: Should we ping Julian?

ximion:
@sidt4 
> Looks like Matthias already commented on this few hours ago (which I see in my email), but not in this PR.

Yeah, I deleted that because that comment was from a  very quick glance at the patch, and after looking at it more I decided I needed to give it a much closer look for a more qualified comment, especially after looking at the `pkgSourceList` API which would make the most sense to use here.

> @ximion: Should we ping Julian?

We absolutely should ;-) @tintou what was the issue with `pkgSourceList`, and @julian-klode, if you have time, could you have a look at this as well?

tintou:
They are not targeting the same audience, `pkgSourceList` is made for apt and thus directly parses the source list into the sources indexes to fetch so it's not what we want here. Here we mostly want to have access to (for the user to be able to understand it):
 * Type
 * URIs
 * Components
 * Suites
 * Enabled

But we also need a way to enable/disable a selected source

julian-klode:
The information requested is available from the index targets and you can get back from the index target to the stanza it's from.

I don't quite understand the point of this. If you want to edit sources you don't need the parsing logic in the first place. Why would you card about the whole .list to deb822 option mapping here? This is just wrong.

You use pkgSourcesList to parse sources, for editing you can use pkgTagFile; parse the given file and iterate to the specified section in the location, present the section to the user to edit; let the user edit it, parse it using pkgSourcesList to see if it's correct, and then use pkgTagSection::Rewrite in the original section to rewrite the modified values?

That all being said pkgTagFile also doesn't retain comments, so editing via any of those APIs is a strong no-go.

The best course of action is to fail any such requests as unsupported.

ximion:
> That all being said pkgTagFile also doesn't retain comments, so editing via any of those APIs is a strong no-go.

It's part of the PackageKit API though and we currently support it. And I think we *at least* need to support it for enabling/disabling sources, but I guess if we remove this feature now someone out there will complain for sure.

So yeah, tbh, I think we actually do need a way to read **and** edit the file. And we really need to address this for Debian 13 (I will make a PK release soon).

ximion:
Currently with this patch applied, once you disable a software source, PackageKit will write a classic sources.list entry into the deb822-style entry, breaking APT updates.

ximion:
Currently with this patch applied, once you disable a software source, PackageKit will write a classic sources.list entry into the deb822-style entry, breaking APT updates.

We also need to write these files properly, *and* ideally also preserve all comments in them. At least enabling(disabling sources is now fairly easy with the `Enabled` field in the deb822 entries :-)

ximion (changes requested):
Currently with this patch applied, once you disable a software source, PackageKit will write a classic sources.list entry into the deb822-style entry, breaking APT updates.

We also need to write these files properly, and ideally also preserve all comments in them. At least enabling/disabling sources is now fairly easy with the Enabled field in the deb822 entries :-)
(And we can likely have a comment-preserving editor for that, because all features PK needs to support is to add sources, remove them, enable/disable them and list them - the latter this PR already implements, the former still need work).

tintou:
Leaving it to you 😉

PR #830, opened 2025-03-12
meson: Add "dbdir" option allowing to specify where to install transactions.db

arrowd:
The main rationale for this change is that FreeBSD prefers /var/db for keeping
programs state and does not have /var/lib

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -81,5 +81,5 @@ endif
 
 install_data(
   'transactions.db',
-  install_dir: join_paths(get_option('localstatedir'), 'lib', 'PackageKit'),
+  install_dir: pk_db_dir,
 )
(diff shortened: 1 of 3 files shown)


sidt4:
Explaining why this change is needed would be helpful.

arrowd:
Yes, you're right. I updated the commit's description.

sidt4:
`/var/lib/PackageKit/` is home to lot of other PK files (`offline-update-competed`, `prepared-update` etc) and not just `transactions.db`.

arrowd:
Ugh, right, and the "lib" part is hardcoded there.

Whelp, this requires a bit more work then, for which I don't have enough time ATM.

Thanks for the review!

ximion:
This is also kind of part of PackageKit's API, so changing it may not be wise...

PR #831, opened 2025-03-12
FreeBSD CI: Switch image to FreeBSD 14

arrowd:


arrowd:
Not sure why tests started to fail on FreeBSD, will research it later.

PR #834, opened 2025-03-17
PkProgress: Move the progress_callback from PkClientState to PkProgress

tintou:
PkProgress is actually the member that get updated on progress so use it to notify on any new progress.

### DIFF
--- a/lib/packagekit-glib2/pk-client.c
+++ b/lib/packagekit-glib2/pk-client.c
@@ -774,8 +774,8 @@ pk_client_set_property_value (PkClientState *state,
 
 	/* remaining-time */
 	if (g_strcmp0 (key, "RemainingTime") == 0) {
-		ret = pk_progress_set_elapsed_time (state->progress,
-						    g_variant_get_uint32 (value));
+		ret = pk_progress_set_remaining_time (state->progress,
+						      g_variant_get_uint32 (value));
 		if (ret && state->progress_callback != NULL) {
 			state->progress_callback (state->progress,
 						  PK_PROGRESS_TYPE_REMAINING_TIME,
--- a/lib/packagekit-glib2/pk-client.c
+++ b/lib/packagekit-glib2/pk-client.c
@@ -44,6 +44,7 @@
 #include <packagekit-glib2/pk-enum.h>
 #include <packagekit-glib2/pk-package-id.h>
 #include <packagekit-glib2/pk-package-ids.h>
+#include <packagekit-glib2/pk-progress-private.h>
 
 static void     pk_client_finalize	(GObject     *object);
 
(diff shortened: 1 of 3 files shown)


tintou:
This also fixes remaining time being emitted as elapsed time…

sidt4:
> This also fixes remaining time being emitted as elapsed time…

Can you make this bug fix a separate commit (first commit) ?

Will take a detailed look this weekend. Thanks!

sidt4:
Thanks for looking into this!

The right way to do things here is for clients of `PkProgress` API to connect to the properties (`"notify::percentage"` etc) using `g_signal_connect*` functions and get the progress callback invoked automatically when the property value changes and not invoke the callback manually (current code and this PR). All `set` functions have a convention to return `void`, but here we're returning `boolean` which is not right to start with. Though the code looks cleaner with the new `pk_progress_new_with_callback`, I don't think this code change is the right fix. The naming of `pk_progress_emit_callback` also hints to this (we emit a signal not a `callback`).

Any reason not to use `g_signal_connect` here?

@ximion: Any reason why we didn't use the `void` return setter convention here?

tintou:
Ideally the callback should go away in the next API break (if any!) as it is really equivalent to a `notify::foo`. I agree that connecting to `notify::foo` is also a solution but that would mean that we would connect to 13 properties, create 13 callback functions or have a multiplexer if we only connect to `notify` and that would not make that code cleaner :sweat_smile: so I think that here is the best compromise for now

ximion:
I originally just cherry-picked the time-remaining bugfix, but the more I looked over the other patch, I agree with @tintou that this is the way to go - for now, until we want to break the API.
This is a pretty nice improvement of the existing code :-)

> Any reason why we didn't use the void return setter convention here?

Well, in C it doesn't really matter and I bet at the time this code was quite an efficiency improvement (as you immediately know if setting a value actually has changed state). But I do not know for certain, this code is more than 15 years old at this point :-)

sidt4:
Okay. I think we can rename `pk_progress_emit_callback` into something better (`pk_progress_invoke_callback` ?).

Also, during my initial review, I vaguely remember there were 16 entries, and some 3 entries were not accounted for (??). I've lost track of them, Once that's sorted out, we can merge this.

ximion:
Whoops, a bit too late, I already merged this :sweat_smile: 
But the helper function rename is a good idea, in fact, I thought the same ^^

> Also, during my initial review, I vaguely remember there were 16 entries, and some 3 entries were not accounted for

Hmm, I actually checked for this and just found one such instance and also found no issues so far during testing.
Can you file a bug or new PR in case I missed anything?

@tintou Thank you for the patch, I very much appreciate these code maintenance patches :-)

PR #841, opened 2025-04-16
apt: Define an apt backend static library

tintou:
Include it in the module and in the test executable. No behavioral changes.


ximion:
Yes, I think we can do this :-)

PR #843, opened 2025-04-17
Add the 'offline-clear' subcommand to pkcon

arrowd:


ximion:
Why would one use this? (except for debugging, of course)

arrowd:
I used it exactly when debugging offline stuff on FreeBSD.

sidt4:
`pkcon offline-trigger` already does this internally. So, why do we need this as a standalone command?

arrowd:
If I read the code correctly, is is not possible to clear `PK_OFFLINE_RESULTS_FILENAME` using "offline-trigger" :

```
// prepared_file = PK_OFFLINE_PREPARED_FILENAME
	if (!g_file_test (prepared_file, G_FILE_TEST_EXISTS)) {
		g_set_error (error,
			     PK_OFFLINE_ERROR,
			     PK_OFFLINE_ERROR_NO_DATA,
			     "Prepared update not found: %s",
			     prepared_file);
		return FALSE;
	}

	/* triggering a new update clears the status from any previous one */
	if (!pk_offline_auth_clear_results (error))
```

sidt4:
What I meant was, in a normal offline update workflow (debugging workflow too), this will be done automatically by `pkcon offline-trigger` command.

### Step 1: Prepare update:

`$ pkcon update --only-download <packages-to-update>`

### Step 2: Trigger update:

`$ pkcon offline-trigger`

### Step 3: Perform offline update:

`$ PK_OFFLINE_UPDATE_TEST=1 /usr/libexec/pk-offline-update`

 So, having `pkcon offline-clear` will be confusing from a `pkcon` command point of view. There is no use for the command in the normal flow of offline update.

arrowd:
Sorry, maybe I'm missing something. At the end of step 3 I ended up with `PK_OFFLINE_RESULTS_FILENAME` created that contains a list of packages updated. And "offline-clear" allows me to remove it. What would otherwise remove this file?

sidt4:
> What would otherwise remove this file?

It will persist until the next update. Why do you want it removed ?

arrowd:
Frankly, I'm not sure. I just thought that if there is an API to remove it, then it might sense to export it as `pkcon` subcommand. All right, feel free to close this PR, then.

sidt4:
PackageKit clients read the contents of `PK_OFFLINE_RESULTS_FILENAME` to display list of packages last updated in the UI.

Example, try running `gnome-software --mode=updated` after an offline update.

So, it's not a left over file, but a update transaction log :).

arrowd:
Yes, but don't GUI clients call to clear these results at some point?

PR #842, opened 2025-04-16
Fix clang warnings

sidt4:
Please refer commit logs.

PR #844, opened 2025-04-17
ci: Reduce verbose logs in freebsd build

sidt4:
Removes the `-v` ninja build flag, as there is just too much build output. 

https://github.com/PackageKit/PackageKit/pull/842 fixes all build issues related to `FreeBSD` (which uses `Clang`).

@arrowd: Thoughts?


arrowd:
Sure, I'm fine with this change.

Issue #839, opened 2025-04-13
Zypp -Wdeprecated-declarations warnings


sidt4:
From: https://github.com/PackageKit/PackageKit/actions/runs/14432889256/job/40469819487?pr=787

```
[231/232] Compiling C++ object backends/zypp/libpk_backend_zypp.so.p/pk-backend-zypp.cpp.o
../backends/zypp/pk-backend-zypp.cpp:364:78: warning: ‘RepoReport’ is deprecated [-Wdeprecated-declarations]
  364 | struct RepoReportReceiver : public zypp::callback::ReceiveReport<zypp::repo::RepoReport>, ZyppBackendReceiver
      |                                                                              ^~~~~~~~~~
In file included from ../backends/zypp/pk-backend-zypp.cpp:72:
/usr/include/x86_64-linux-gnu/zypp/ZYppCallbacks.h:262:37: note: declared here
  262 |     struct ZYPP_API ZYPP_DEPRECATED RepoReport : public callback::ReportBase
      |                                     ^~~~~~~~~~
../backends/zypp/pk-backend-zypp.cpp:379:142: warning: ‘RepoReport’ is deprecated [-Wdeprecated-declarations]
  379 |         virtual void finish (zypp::Repository source, const std::string &task, zypp::repo::RepoReport::Error error, const std::string &reason)
      |                                                                                                                                              ^
/usr/include/x86_64-linux-gnu/zypp/ZYppCallbacks.h:262:37: note: declared here
  262 |     struct ZYPP_API ZYPP_DEPRECATED RepoReport : public callback::ReportBase
      |                                     ^~~~~~~~~~
../backends/zypp/pk-backend-zypp.cpp:379:142: warning: ‘RepoReport’ is deprecated [-Wdeprecated-declarations]
  379 |         virtual void finish (zypp::Repository source, const std::string &task, zypp::repo::RepoReport::Error error, const std::string &reason)
      |                                                                                                                                              ^
/usr/include/x86_64-linux-gnu/zypp/ZYppCallbacks.h:262:37: note: declared here
  262 |     struct ZYPP_API ZYPP_DEPRECATED RepoReport : public callback::ReportBase
      |                                     ^~~~~~~~~~
../backends/zypp/pk-backend-zypp.cpp: In function ‘zypp::target::rpm::RpmHeader::constPtr zypp_get_rpmHeader(const std::string&, zypp::Edition)’:
[...]

PR #845, opened 2025-04-17
ci: Disable deprecation errors in zypp backend temporarily

sidt4:
This is just an interim fix until the deprecation warnings are fixed.

Refer: https://github.com/PackageKit/PackageKit/issues/839

### DIFF
--- a/backends/zypp/meson.build
+++ b/backends/zypp/meson.build
@@ -22,10 +22,10 @@ shared_module(
     '-Wall',
     '-Woverloaded-virtual',
     '-Wnon-virtual-dtor',
+    '-Wno-error=deprecated-declarations',
     '-std=c++1z'
   ],
   c_args: [
-    '-Wno-deprecated',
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -183,6 +183,15 @@ if get_option('maintainer')
 
         language: ['c', 'cpp']
     )
+
+    # Refer https://github.com/PackageKit/PackageKit/issues/839
+    backend = get_option('packaging_backend')
+    if backend.length() == 1 and backend[0] == 'zypp'
+      add_project_arguments(
```
ximion, line 190: Please don't add these project-wide, please only add the flag to the Zypp targets specifically :-)

ximion (changes requested):

meson.build
sidt4: Right. Thanks!

ximion:
Thanks!

PR #850, opened 2025-04-25
pk-control: Use g_debug() rather than g_warning() for proxy error

pwithnall:
The error is already propgated programmatically as a `GError`, which allows the caller to handle it as they see fit.

Additionally printing a warning message slightly undermines the caller’s ability to ignore errors here, as it will cause an abort if running with `G_DEBUG=fatal-warnings` (for example).

This is causing problems when running PackageKit in gnome-software’s CI, which doesn’t support proxy settings (as no logind session is running) — the tests are run with `G_DEBUG=fatal-warnings`, and this warning then becomes fatal.


ximion:
You're right, I don't know why we also logged a warning here in addition to returning an error...
Thanks for the patch!

PR #852, opened 2025-05-04
Support reading package IDs separated by &

arrowd:
This is a first step to properly address #795 

Trying to maintain backwards compatibility, the code first checks if package IDs are separated the old way. If no commas are found, we go the new way by reusing `pk_package_ids_*` API.


arrowd:
Applied the `g_key_file_set_string_list` suggestion.

sidt4:
I think a change in this code path would need to have lot more unit tests (adding new tests if needed).

arrowd:
What tests do you have in mind? It feels to me that these tests would be actually testing `g_key_file_*` API, which is not relevant for PackageKit, but maybe I'm missing something.

sidt4:
Also, I think we can introduce a new `version` key to have more control on parsing code (than guessing the format used from parsing text). This should also allow us to add new features more easily, going ahead.

### Current: 

No version can be taken as `version=1`

```
[update]
prepared_ids=prepared_ids=code;1.103.2-1755709794;amd64;code_stable-stable-main,dirmngr;2.4.8-3;amd64;debian-unstable-main
```

### New:

```
[update]
version=2
prepared_ids=prepared_ids=code;1.103.2-1755709794;amd64;code_stable-stable-main|dirmngr;2.4.8-3;amd64;debian-unstable-main
```

### Future:

```
[update]
version=3
prepared_ids=prepared_ids=code;1.103.2-1755709794;amd64;code_stable-stable-main|dirmngr;2.4.8-3;amd64;debian-unstable-main
new_key=....
```

sidt4:
>  Do you mean that using non ; separator is better just because there would be less escaping performed?

Yes. Too many `\` (as shown in https://github.com/PackageKit/PackageKit/pull/852#issuecomment-3213761411) is distracting and less legible to read.

lib/packagekit-glib2/pk-offline.c
```diff
@@ -722,18 +725,19 @@ pk_offline_get_results (GError **error)
...
-			if (!pk_package_set_id (pkg, package_ids[i], error))
-				return NULL;
-			pk_results_add_package (results, pkg);
-		}
+	package_ids = g_key_file_get_string_list (file, PK_OFFLINE_RESULTS_GROUP,
+						  "Packages", &package_ids_size, NULL);
+	if (package_ids == NULL || package_ids_size == 0)
+		goto end;
+
+	for (i = 0; package_ids[i] != NULL; i++) {
```
ximion, line 733: nit: (we can modernize this, while we're at it, since we can use a newer C standard now)
```suggestion
	for (guint i = 0; package_ids[i] != NULL; i++) {
```

lib/packagekit-glib2/pk-offline.c
```diff
@@ -722,18 +725,19 @@ pk_offline_get_results (GError **error)
...
+		g_autoptr(PkPackage) pkg = NULL;
+		pkg = pk_package_new ();
+		pk_package_set_info (pkg, PK_INFO_ENUM_UPDATING);
+		if (!pk_package_set_id (pkg, package_ids[i], error))
+			return NULL;
+		pk_results_add_package (results, pkg);
 	}
+end:
```
ximion, line 741: nit: We commonly call this `out` or `finish` in other parts of the code - definitely a nitpick though.

lib/packagekit-glib2/pk-package-ids.h
```diff
@@ -36,6 +36,7 @@ G_BEGIN_DECLS
  * '|': used as the filename separator
  * '~': conary
  * '@': conary
+ * ',': may be a part of version string in some OSes
```
ximion, line 39: I think we can delete `PK_PACKAGE_IDS_DELIM` entirely now? And maybe move that documentation to the function writing it, expressing why comma is preferred, and that it will be escaped by GKeyFile in places where it is part of the version string.

ximion (changes requested):
It's mostly nitpicks, except for the documentation update and `PK_PACKAGE_IDS_DELIM` removal - that needs to be addressed.

Sorry for the delay in reviewing, the rest of the code looks good to me!

lib/packagekit-glib2/pk-package-ids.h
ximion: Hmm, the annoying thing is that we can't easily change the delimiter as it is part of the API - I would really like to avoid it. Fortunately, I found zero users of it, but I still wonder if we could just use the same (un)escaping that GKeyFile uses in the helper routines, to solve this problem forever.

ximion: Just copy it, it should just be string-escaping, right? I can't imagine that that would be complex at all, you can probably write it quickly with a GString and going through input data char-by-char in one go.

lib/packagekit-glib2/pk-offline.c
```diff
@@ -473,17 +474,18 @@ pk_offline_get_prepared_ids (GError **error)
...
 		/* fall back to previous plain text file format for backwards compatibility */
 		return g_strsplit (data, "\n", -1);
 	}
 
-	prepared_ids = g_key_file_get_string (keyfile, "update", "prepared_ids", error);
-	if (prepared_ids == NULL)
+	prepared_ids = g_key_file_get_string_list (keyfile, "update", "prepared_ids", &prepared_ids_size, error);
+
+	if (prepared_ids == NULL || prepared_ids_size == 0)
 		return NULL;
```
mcrha, line 486: in case the `prepared_ids` is empty, but non-NULL, the pointer is leaked here.

arrowd: Of course, that's my fault. Will do a bit later today.

PR #857, opened 2025-05-14
Require polkit 0.114

tintou:
We can safely require this version from 2018 and remove our autoptr workarounds.


ximion:
Indeed! Thank you!

Issue #790, opened 2024-08-25
freebsd: pk_test_backend_func: assertion failed (number_packages == 1): (0 == 1)


sidt4:
From `Cirrus CI / Build & Test (freebsd)` job at https://github.com/PackageKit/PackageKit/pull/789/checks?check_run_id=29205364525

```
# DEBUG: transaction now finished
# DEBUG: 1 transactions in list, 0 committed but not finished
ok 5 /packagekit/scheduler-parallel
# slow test /packagekit/scheduler-parallel executed in 21.09 secs
# DEBUG: trying to open database './transactions.db'
# DEBUG: creating table to repair: Failed to execute statement 'SELECT * FROM transactions LIMIT 1': no such table: transactions
# DEBUG: adding last action details: Failed to execute statement 'SELECT * FROM last_action LIMIT 1': no such table: last_action
# DEBUG: adding config: Failed to execute statement 'SELECT * FROM config LIMIT 1': no such table: config
# DEBUG: adding table proxy: Failed to execute statement 'SELECT * FROM proxy LIMIT 1': no such table: proxy
# DEBUG: trying to open database './transactions.db'
# DEBUG: job count is now at 1
# DEBUG: job count now 2
# DEBUG: job count now 3
# DEBUG: emitting updates-changed
# DEBUG: set proxy 127.0.0.1:80, 127.0.0.1:21 for uid:500 and session:session1
# DEBUG: updated proxy 127.0.0.1:80, 127.0.0.1:21 for uid:500 and session:session1
ok 6 /packagekit/transaction-db
# slow test /packagekit/transaction-db executed in 2.02 secs
# GLib-GIO-DEBUG: _g_io_module_get_default: Found default implementation local (GLocalVfs) for ?gio-vfs?
# DEBUG: config file changed
# DEBUG: config file changed
# DEBUG: config file changed
# DEBUG: already added license_foo to accepted list
# DEBUG: Trying to load : invalid
# DEBUG: local backend not found 'backends/invalid/libpk_backend_invalid.so'
# DEBUG: dlopening '/usr/local/lib/packagekit-backend/libpk_backend_invalid.so'
# DEBUG: Trying to load : dummy
# DEBUG: dlopening 'backends/dummy/libpk_backend_dummy.so'
# DEBUG: already closed (nonfatal)
# DEBUG: Trying to load : dummy
# DEBUG: dlopening 'backends/dummy/libpk_backend_dummy.so'
# DEBUG: config file changed
[...]

PR #858, opened 2025-05-15
FreeBSD CI: Use latest package set

arrowd:


arrowd:
This might fix https://github.com/PackageKit/PackageKit/issues/790

arrowd:
All FreeBSD package repositories are rolling release. It is just `quarterly` makes a big roll once in 3 months while `latest` rolls constantly.
I try to use `latest` in FreeBSD CI's, because when I get to update a FreeBSD port to a new upstream release, it will be compiling against dependencies from `latest` and not `quarterly`.
This specific PR is needed, because `latest` ships GLib 2.84, that includes improvements to `GFileMonitor` which I suspect to be the culprit for the CI flakiness during testing stage.

ximion:
Okay, that makes sense!
Thanks! :-)

sidt4:
> This might fix https://github.com/PackageKit/PackageKit/issues/790

Great! The last few runs in https://cirrus-ci.com/github/PackageKit/PackageKit are green. We can keep a watch for the next runs.

PR #859, opened 2025-05-16
ci: Check ABI with last know good one

tintou:
Add automated ABI check with abigail.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -9,33 +9,23 @@ jobs:
 
     strategy:
       matrix:
-        distro:
-          - fedora
-          - debian
+        include:
+          - distro: debian
+            backend: apt
+          - distro: fedora
+            backend: dnf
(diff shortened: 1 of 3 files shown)


ximion:
Nice, this actually looks way easier than I thought it would be!

The only issue we will run into is that PK doesn't hide private headers from the ABI currently, so we will trigger this by accident quite a bit I guess. It's probably still worth it.

Could the ABI check be split out as separate job from the build tests, so it shows up as its own thing in the summary pages? (currently, if the Debian check fails, it could be for many reasons, so one always has to look at the logs instead of getting a quick idea what went wrong)

tintou:
The `--headers-dir1` allows abigail to only check the ABI from the public header so we aren't getting warned about private ABI

tintou:
Use a separate job for the ABI check

ximion:
Perfect, thank you! :-)

PR #867, opened 2025-06-12
pk-client: Refactor PkClientState lifecycle handling to fix leaks

### LINKED ISSUE #709: pkmon: GTask finalized without ever returning (using g_task_return_*())

pwithnall:
This refactors the lifecycle of `PkClientState` a bit to eliminate some leaks I was seeing in gnome-software. It potentially also fixes #709, although I’ve only briefly tested that.

The main changes are that now:
 * `PkClient` holds a strong ref on every `PkClientState`
 * `PkClientState` no longer holds a strong ref on `PkClient`
 * A few code paths were fixed where `pk_client_state_finish()` was not called to finish a `PkClientState`
 * Each `PkClientState` is now added to `client->calls` as soon as it’s created, rather than after two async calls, which fixes cancellation of operations while either of those two calls are in flight
 * Lifecycle management of the `PkClientState` for `pk_client_get_progress_async()` has been made consistent with the rest of the code (it previously had its own outdated copy of `pk_client_state_finish()`)

The rest of the changes are cleanups or smaller fixes.

I’ve tested this for a few hours with a copy of gnome-software `main` with asan enabled. Previously it would leak about 500KB (directly and indirectly) from `gs_plugin_packagekit_auto_prepare_update_thread()` when using the following test setup:
 1. `LSAN_OPTIONS=report_objects=1,suppressions=supp.txt gnome-software &> out.log`
 2. Wait for it to load, and wait for `stat /var/lib/PackageKit/prepared-update` to succeed
 3. `sudo rm /var/lib/PackageKit/prepared-update` and wait 30s to trigger a prepared-updates check in gnome-software
 4. Wait for `stat /var/lib/PackageKit/prepared-update` to succeed again
 5. `gnome-software --quit` to cause gnome-software to exit gracefully and allow lsan to assess the damage

With these changes (and https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/2296) I get no leaks.

`supp.txt` is the following, to minimise noise from unrelated leaks (and false positives):
```
leak:as_xml_add_text_prop
leak:gsk_gpu_node_processor_add_node
leak:pango_cairo_fc_font_map_fontset_key_substitute
leak:FcFontSetMatch
leak:_gtk_css_reference_value_new
```

I’ll continue to test this with my background instance of `gnome-software`, but it’s ready for review and there’s nothing (from my point of view) blocking merge.

### DIFF
--- a/lib/packagekit-glib2/pk-client.c
+++ b/lib/packagekit-glib2/pk-client.c
@@ -1012,6 +1012,8 @@ pk_client_copy_downloaded_file (PkClientState *state, const gchar *package_id, c
  * We have to copy the files from the temporary directory into the user-specified
  * directory. There should only be one file for each package, although this is
  * not encoded in the spec.
+ *
+ * This will eventually call pk_client_state_finish() on `state`.
  */
 static void
 pk_client_copy_downloaded (PkClientState *state)
@@ -1021,15 +1023,19 @@ pk_client_copy_downloaded (PkClientState *state)
 	guint len;
 	PkFiles *item;
 	g_autoptr(GPtrArray) array = NULL;
+	unsigned int n_files_to_copy;
 
 	/* get data */
 	array = pk_results_get_files_array (state->results);
 
 	/* get the number of files to copy */
+	n_files_to_copy = 0;
 	for (i = 0; i < array->len; i++) {
 		item = g_ptr_array_index (array, i);
-		g_atomic_int_add (&state->remaining_files_to_copy, g_strv_length (pk_files_get_files (item)));
+		n_files_to_copy += g_strv_length (pk_files_get_files (item));
 	}
+
+	g_atomic_int_set (&state->remaining_files_to_copy, n_files_to_copy);
 
 	/* get a cached value, as pk_client_copy_downloaded_file() adds items */
 	len = array->len;
@@ -1047,6 +1053,12 @@ pk_client_copy_downloaded (PkClientState *state)
 							pk_files_get_package_id (item),
 							files[j]);
 		}
+	}
+
+	/* Were there actually any files to copy? */
+	if (n_files_to_copy == 0) {
+		state->ret = TRUE;
+		pk_client_state_finish (state, NULL);
 	}
 }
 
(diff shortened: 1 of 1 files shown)


ximion:
I'll be at a week-long conference next week (and need to prepare the flight to it) - I'll try to review this sooner, but if that doesn't work out, apologies in advance for only reviewing in ~2 weeks!

sidt4:
I've applied `PackageKit` git + this MR on my system live.  `pkmon` is usable at last.

Thanks a lot @pwithnall!

tintou:
Thank you, this looks great!

lib/packagekit-glib2/pk-client.c
```diff
@@ -249,9 +255,11 @@ pk_client_state_finish (PkClientState *state, GError *error)
 
 	/* remove from list */
 	pk_client_state_remove (state->client, state);
+	/* `state_owned` may be the only strong ref to `state` remaining after this point */
 
 	/* mark the state as finished */
 	g_clear_object (&state->res);
+
```
ximion, line 262: nit: Extra linebreak

ximion (approved):
This looks and works great! Thank you so much for the detailed commit messages, they helped immensely to wrap my head around why certain changes were made - I wish all PRs were like this! ^^

I tested this for about a day as well, and noticed no issues, and my only note is a whitespace nitpick... Do you want to resolve that? In any case, I'd merge this soon and we should have a release to include the change soon as well, I think.

sidt4:
https://gitlab.gnome.org/GNOME/gnome-software/-/issues/2873 might be related to this change. Not sure.

PR #870, opened 2025-07-02
dnf: update ROSA config

mikhailnov:
restricted repository was removed some years ago in all dnf-based plaforms. i586 was renamed to i686 in all dnf-based platforms. Added new arches.

See: https://mirror.rosa.ru/rosa/rosa13 and https://abf-downloads.rosa.ru/rosa13


ximion:
Thank you! :-)

PR #875, opened 2025-07-15
apt: Handle gstreamer ()(64bit) suffix on any architecture

aleasto:
The ()(64bit) suffix is an RPM artifact, where it is used to distinguish between the primary and secondary architecture in a 2-arch system. It doesn't translate precisely to DPKG, where multi-arch is not limited to one secondary architecture. Let's use it the same way RPM would and only match the native arch.


ximion:
This looks good to me, but I haven't had time to test it yet.

PR #877, opened 2025-07-15
meson: resolve warnings `-Wall`/`-Wextra`/`-Werror`

osalbahr:
The warnings:

```
meson.build:167: WARNING: Consider using the built-in werror option instead of using "-Werror".
meson.build:167: WARNING: Consider using the built-in warning_level option instead of using "-Wall".
meson.build:167: WARNING: Consider using the built-in warning_level option instead of using "-Wextra".
```

See https://mesonbuild.com/Builtin-options.html

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -2,7 +2,7 @@ project('PackageKit', 'c',
   version : '1.3.2',
   license : 'LGPL-2.1+',
   meson_version : '>=1.0',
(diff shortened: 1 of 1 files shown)


ximion:
We do not want this, as we want to conditionally enable/disable stricter error rules. Meson currently has no way to do that. The warning will not be shown if you aren't building in maintainer mode (= when more relaxed rules are in effect, as they should be for distribution builds).

PR #851, opened 2025-05-03
Improvements to 'get-details' command output

sidt4:
Please refer commit log.

PR #878, opened 2025-07-15
backends: Add eopkg

joebonrichie:
Eopkg is a long-term fork of the PiSi package manager used in Solus. It has long deviated from the original code-base but, mostly contains compatibility with the original API calls used. It is not feasible to share the same codebase as Solus has largely rewritten most of the surrounding tooling used by the package manager as well as differences in eopkg itself.

Several fixes will be applicable to the PiSi backend as well if the maintainers wish to trawl through.

Needs #771 resolved first.
Depends on #851.


Conan-Kudo (approved):

sidt4:
> It reminds me that I should probably delete the PiSi backend since it's still rotting and nobody ever did anything with it after it was restored.

We can discuss it in https://github.com/PackageKit/PackageKit/issues/894.

ximion (changes requested):
@joebonrichie Please also add yourself as maintainer for this backend to https://github.com/PackageKit/PackageKit/blob/main/backends/MAINTAINERS.md (so we know who to ping in case bugs are reported against this)

@Conan-Kudo I will probably squash this simply because it makes my job easier at release time to distinguish between changes that were actually new / made post-merge, and development that happened before the PR was merged.
But I'll look at the commits before deciding that, sometimes it is useful to keep the history.

backends/eopkg/pk-backend-eopkg.c
```diff
@@ -0,0 +1,459 @@
...
+
+#include <pk-backend.h>
+#include <pk-backend-spawn.h>
+
+static PkBackendSpawn *spawn;
+
+static const gchar *eopkg_get_backend_filename (void);
+
+static const gchar
+*eopkg_get_backend_filename (void)
```
ximion: Style: Please put the pointer asterisk next to the type name, not in a new line before the function name.

ximion:
This looks good to me now - I can't comment on the Solus bits specifically, but everything I've reviewed looks sane now and all documentation is filed correctly.

So, in other words: Welcome aboard! :-)

PR #884, opened 2025-08-20
Update sl.po (Slovenian)

filmsi:
This make Slovenian 100% completed.

### DIFF
--- a/po/sl.po
+++ b/po/sl.po
@@ -1,15 +1,18 @@
 # SOME DESCRIPTIVE TITLE.
 # Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
 # This file is distributed under the same license as the PackageKit package.
-#
+# 
 # Translators:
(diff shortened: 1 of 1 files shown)


filmsi:
@ximion 
OK, the link you provided does not work (it reports Bad Gateway).
Could you please provide a working link?

Also, you still keep packagekit for translation at Transifex:
https://app.transifex.com/freedesktop/packagekit/dashboard/

Does this mean you dropped Transifex? If so, it is misleading to keep the project still appearing at Transifex.

ximion:
You must have been unlucky and hit downtime at a bad time.

The Transifex page is read-only and has an announcement to switch to Weblate, that should be enough. If it was deleted, all we get would be a bunch of dead links and we couldn't deliver information to the current translators.

ximion:
Yes, for fwupd, but not for PackageKit ;-) There it's almost complete: https://hosted.weblate.org/projects/packagekit/translations/sl/

ximion:
Excellent, thank you for translating PackageKit!

PR #889, opened 2025-08-26
Sanity check all object types in public API

sidt4:
Please refer commit log.

Helps: https://github.com/PackageKit/PackageKit/pull/855.


ximion:
> If this sounds ok, I'll submit a new MR for types in `src/` dir.

It's a bit of a weird kind of test, but in light of the currently ongoing modernization and refactoring, it does make sense to have it. Plus, all objects should be able to be constructed and destroyed without warnings or errors, that way we test for that.

In other words: Looks good to me!

PR #899, opened 2025-09-25
pk-offline-update: Properly print newlines on systems without sd_journal_print

arrowd:


arrowd:
This is how `sd_journal_print()` is used in the code:

```
	sd_journal_print (LOG_WARNING,
				  "failed to display message on splash: %s",
				  error->message);
```

While its signature matches the `dprintf` one, there is no '\n' in the input strings. `dprintf` obviosly does not add its own newlines either, as don't all printf-family functions. This results in log messages concatenated to each other.

arrowd:
It does an additional `puts("")`.

ximion:
Yes, but to `stdout` and not `fd` - that can make quite a massive difference...

PR #900, opened 2025-09-26
pk-offline.c: Plug a memleak I introduced earlier

arrowd:
From what I gather, we can't use `g_autofree` on this pointer, because it must be freed with `g_strfreev`. The `g_autoptr(Gstrv)` implementation doesn't exist either.

This is a follow-up to https://github.com/PackageKit/PackageKit/pull/852#pullrequestreview-3270248754

### DIFF
--- a/lib/packagekit-glib2/pk-offline.c
+++ b/lib/packagekit-glib2/pk-offline.c
@@ -446,7 +446,7 @@ pk_offline_get_prepared_ids (GError **error)
 {
 	g_autoptr(GError) error_local = NULL;
 	g_autofree gchar *data = NULL;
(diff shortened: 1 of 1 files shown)


arrowd:
> Why not use g_auto(GStrv) for this?

Ah, that's how it is called. All right, fixed, thanks for the hint.

ximion:
The CI disagrees ;-)

(you'll have to use g_steal_pointer on the return statement for this to work now)

arrowd:
Ugh, sorry, it is pretty late there and I'm brainfarting a lot.

ximion:
No worries, that's what reviews and CI are for.
Thank you for the quick fix!

PR #901, opened 2025-09-26
pk-package-ids: Properly escape the separator found in package IDs

arrowd:
While there, clarify why comma is a bad choice for separator

This is a follow-up for the request raised in https://github.com/PackageKit/PackageKit/pull/852#discussion_r2377188533

I'm a bit worried about the fact that it is now wrong to do `strcmp(some_id, package_ids[i])`, because `some_id` might be unescaped and `package_ids` elements are escaped. But we have `pk_package_ids_present_id()` for that, I guess.

### DIFF
--- a/lib/packagekit-glib2/pk-package-ids.c
+++ b/lib/packagekit-glib2/pk-package-ids.c
@@ -25,6 +25,50 @@
 #include <packagekit-glib2/pk-package-id.h>
 #include <packagekit-glib2/pk-package-ids.h>
 
+/* Rationale for choosing ',' as separator is to make it match with what
+ * pk-offline machinery does when storing offline update results.
+ *
+ * If this has to be changed, also change:
+ * - lib/python/packagekit/backend.py
+ */
+#define PK_PACKAGE_IDS_DELIM_STR	","
+#define PK_PACKAGE_IDS_DELIM_CHR	','
+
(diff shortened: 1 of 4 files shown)


ximion:
I don't understand, why is `PK_PACKAGE_IDS_DELIM_STR` still there? The list of package-IDs should still be comma-separated, just the internal commas should be escaped (basically the same logic that GKeyFile uses).

In this patch, you are somehow doing both, changing the delimiter *and* escaping it, which guarantees maximum incompatibility...

ximion:
If we do that, we'll get exactly the same behavior as before for most versions, properly escaped strings for package-IDs that have commas, and also behavior parity with what GKeyFile writes to a file.

It is consistent behavior while staying mostly compatible as well.

ximion:
I think that was me being stuck with the prior attempt of changing the package-ids separator. Back then a change like this would have made sense. With the GKeyFile solution, it really no longer does, and I apparently didn't update my mental model.

I'm very sorry for that, I made you do a bunch of rather pointless work :-(
I do appreciate you doing it though, and thank you for the patch! I'll close this PR - and maybe implementing a different solution for package-ids serialization makes sense for PackageKit 2, eventually, in time.

arrowd:
No problem at all, it didn't took much time anyways.

PR #903, opened 2025-09-26
dnf5-plugin: Link against libdnf5

aleasto:
There must have been a reason for me to not have linked libdnf5 when first authoring this module. I cannot remember that reason. It seems that it is now mandatory for plugins to link agiainst libdnf5.

rhbz#2399781


backends/dnf/meson.build
```diff
@@ -25,7 +25,7 @@ install_data(
 )
 
 add_languages('cpp', native: false)
-shared_module(
+shared_library(
```
Conan-Kudo, line 28: Does this still produce an unversioned library?

aleasto: Looks like it did, yes.

My reasoning for switching to shared_library is that we can now enforce that there are no missing symbols.
But `shared_module` is still more semantically correct I suppose: this is, afterall, a loadable module.
I've reverted back.

PR #904, opened 2025-10-09
Move FailureAction= to the correct section

andrew-sayers:
[systemd commit 2c883d7](https://github.com/systemd/systemd/commit/2c883d7591f1545315a0c8ae5c487342f3cf9178) says "FailureAction= in [Service] is still supported but deprecated", and the documentation only mentions its presence in [Unit].

Move FailureAction= to [Unit], where it is more likely to remain supported.


ximion:
Makes sense, thank you for the patch!

PR #887, opened 2025-08-24
pk-backend-spawn: Support 'pk_backend_job_details_full()' 

sidt4:
Please refer commit log.

All changes tested on latest Solus GNOME.

Fixes: https://github.com/PackageKit/PackageKit/issues/771

PR #912, opened 2025-11-24
Emit 'download-size' as part of details for all spawned backends

joebonrichie:
Require all spawned backends to emit download-size as part of the details for `get_details()` and `get_details_local()`. For now emit MAXUINT64 from GLib so the download sizes are reported as 'unknown' by `pkcon`. It is up to the individual backend maintainers to add support as necessary. The property is not optional as that already created quite a few surprising headaches with the `pk_backend_job_details_full()` versus `pk_backend_job_details()` split.

Supersedes #887

Follow up to 736585f081f3d4ce5369911880f3b5f62d7fc3c5

### DIFF
--- a/lib/python/packagekit/backend.py
+++ b/lib/python/packagekit/backend.py
@@ -234,7 +234,7 @@ class PackageKitBaseBackend:
         sys.stdout.write(_to_utf8("data\t%s\n" % data))
         sys.stdout.flush()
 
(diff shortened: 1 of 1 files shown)


ximion:
> CI seems to be failing as it can't find enums.py that's generated at compile time

Can you rebase and check if that solves this issue?

PR #917, opened 2025-11-30
meson: Make ply-boot-client a required dependency for offline-updates

Conan-Kudo:
We already assume ply-boot-client 0.9.5 or higher in the code, so not making this a hard dependency is somewhat silly.

Fixes: 8d8417c56 ("offline-updates: Assume plymouth>=0.9.5 by default")

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -25,7 +25,6 @@ gmodule_dep = dependency('gmodule-2.0', version: glib_req)
 sqlite3_dep = dependency('sqlite3')
 polkit_dep = dependency('polkit-gobject-1', version: '>=0.114')
 jansson_dep = dependency('jansson', version: '>=2.8', required: true)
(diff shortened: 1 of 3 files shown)


ximion:
> That's fair. Is there a good reason for it to not be? Otherwise I'm going to move it there and make it required there.

The better question is: Should we have this dependency *at all*? I kept it as a purely informational piece for distributors to tell them which minimum Plymouth version we want for offline-updates to work. But we do not *actually* need Plymouth for a PackageKit build or test at all (that's why it's optional), so we might as well drop it. Making it a hard dependency for no reason is something I dislike, as it potentially complicates any bootstrapping processes in distributions for pretty much no reason at all.

arrowd:
FreeBSD does not have Plymouth, so please keep it optional.

PR #919, opened 2025-12-03
contrib,ci,meson: Drop gobject-introspection and raise to glib2 >= 2.80

Conan-Kudo:
At this point, nobody should be offering this with an older glib2 that does not provide the built-in implementation of the gi tools, so we can safely depend on that now.


ximion:
Good change in principle, but depending on a not-yet-released Meson version, or even one super-recent, is not something we can do for this project...

ximion:
That sounds good! (CI has to pass too, of course)
Thanks!

PR #920, opened 2025-12-06
pk-offline-update: Call into ConsoleKit on FreeBSD to perform poweroff/reboot

arrowd:
With this change I was finally able to perform an offline update on FreeBSD.


Conan-Kudo:
I'm not sure this would work properly without using [InitWare](https://github.com/InitWare/InitWare) or something that implements a similar API that we use on FreeBSD.

arrowd:
I still don't have an understanding why these API are called at all. We're already shutting down/rebooting.

PR #930, opened 2025-12-30
pk-client: Fix assertion failure when PkClientState is cancelled early

pwithnall:
If the `GCancellable` passed to `pk_client_state_new()` is already
cancelled at the time of the call, the cancellation callback
(`pk_client_cancellable_cancel_cb()`) would be called synchronously and
remove the newly created `PkClientState` from the `PkClient.calls`
array.

Unfortunately, this happened immediately before the `PkClientState` was
added to `PkClient.calls` for the first time, in the remainder of
`pk_client_state_new()`.

This meant that the now-defunct `PkClientState` would hang around in
`PkClient.calls` forever, until it triggered an assertion failure in
`pk_client_finalize()` due to `PkClient.calls` not being empty.

Fix that by adding the `PkClientState` to `PkClient.calls` _before_
connecting its cancellable. The code has to be safe to call this way
round, as the cancellation callback could theoretically be called at any
time after the cancellable is connected.

Fixes: https://gitlab.gnome.org/GNOME/gnome-software/-/issues/2873

Fixes: #933

### DIFF
--- a/lib/packagekit-glib2/pk-client.c
+++ b/lib/packagekit-glib2/pk-client.c
@@ -2382,8 +2382,7 @@ pk_client_resolve_async (PkClient *client, PkBitfield filters, gchar **packages,
 	state->progress = pk_progress_new_with_callback (progress_callback, progress_user_data);
 
 	/* check not already cancelled */
-	if (cancellable != NULL &&
-	    g_cancellable_set_error_if_cancelled (cancellable, &error)) {
+	if (g_cancellable_set_error_if_cancelled (cancellable, &error)) {
 		pk_client_state_finish (state, g_steal_pointer (&error));
 		return;
 	}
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
I have been pinged by @mcatanzaro and @mcrha about this, the urgency to deal with it was going up considerably. I _did_ wait, but also I was confident _because_ it had already received a review from someone else. If nobody had, I would have pinged you to make it more urgent to review before landing this.

pwithnall:
For urgent fixes like this in future (in the sense that it’s hopefully a fix for a known frequent crash) then it’s always an option to pull the patch into a distribution package first, before merging it upstream, if it’s passed some initial code review and would benefit from some testing, but hasn’t had a full set of eyes on it upstream yet.

In any case, getting into a distro package is what actually immediately fixes the crashes for people, rather than moving from one git branch to `main` upstream.

Thanks everyone for the reviews. Please let me know if any follow-up work is needed after further review or distro testing. :)

PR #935, opened 2026-01-08
general: Fix spelling of D-Bus in comments and docs

pwithnall:
‘D-Bus’ is the name of the bus technology and project. ‘system bus’ is
what to call the system instance of a bus. Both ‘DBus’ and ‘DBUS’ are
never correct (but are unfortunately frequently used by people).

Drive-by fix to satisfy my niggles. No functional changes.


Conan-Kudo:
Don't worry, it always bothered me too, I just didn't have time to deal with it. 😅 

Description updated!

PR #942, opened 2026-02-07
WIP and only compile tested: new alpine apk-tools backend

jane400:
Hi!
I'm coming from postmarketOS and I feel like we want to replace the custom plugins to various appcenters and https://gitlab.alpinelinux.org/Cogitri/apk-polkit-rs. This is not ready for review yet, as I haven't formatted my code uniformly nor use a consistent style, but I decided to create the MR now to gather feedback about some design decisions. I think it's easier to argue about code on github than sending links and snippets over matrix. I looked at alpm and dummy (and a tiny bit apt) for reference, but I'm mostly inferring expected backend behavior by my gut.

Current state: Some parts were tested, but a bunch of changes which have been made since are only compile tested.


backends/alpine-apk/backend-apk-private.c
```diff
@@ -0,0 +1,306 @@
...
+#include "pk-backend-job.h"
+
+#include <apk/apk_context.h>
+#include <apk/apk_database.h>
+#include <apk/apk_package.h>
+#include <apk/apk_query.h>
+
+int open_apk(OpenApkOptions options, ApkContext **ctx_writeback,
+             ApkDatabase **db_writeback) {
+  g_autoptr(ApkContext) ctx = g_malloc(sizeof(struct apk_ctx));
```
jane400: I'm really unsure about syntax here. `apk-tools` normally allocates `struct apk_context` on the "stack". I have introduced `_pk_apk_auto` (a equvialent to g_auto), which should theoretically work on non pointers too, with tiny adjustments. The only advantage over writing directly (so `struct apk_context *ctx` in this method) is the NULL comparison one could do, but I'm indifferent to it, as we should handle the returned int in all cases anyway. (Is there a `#[must_use]` equivalent in C?

backends/alpine-apk/pk-backend-apk.c
```diff
@@ -0,0 +1,915 @@
...
+  if (result != 0) {
+    goto out;
+  }
+  if (apk_array_len(failed_package_array) > 0) {
+    pk_backend_job_error_code(job, PK_ERROR_ENUM_INTERNAL_ERROR,
+                              "apk_query_packages failed");
+    goto out;
+  }
+
+  apk_array_foreach_item(package, package_array) {
```
jane400: for me: duplicate to convert_apk_to_files.

backends/alpine-apk/pk-backend-apk.c
```diff
@@ -0,0 +1,915 @@
...
+      PK_ROLE_ENUM_RESOLVE, PK_ROLE_ENUM_SEARCH_DETAILS,
+      PK_ROLE_ENUM_SEARCH_FILE, PK_ROLE_ENUM_SEARCH_NAME,
+      PK_ROLE_ENUM_GET_FILES_LOCAL, -1);
+  return roles;
+}
+gchar **pk_backend_get_mime_types(PkBackend *backend) {
+  const gchar *mime_types[] = {NULL};
+  return g_strdupv((gchar **)mime_types);
+}
+gboolean pk_backend_supports_parallelization(PkBackend *backend) {
```
jane400: I'm not sure whether I would claim that this backend supports `pk_backend_supports_parallelization`. My fear is that reads can happen while apk-tools or this backend writes to the underlying `apk_database`.

What does `pk_backend_supports_parallelization` mean anyway?

backends/alpine-apk/backend-apk-private.c
```diff
@@ -0,0 +1,306 @@
+/* -*- Mode: C; tab-width: 8; indent-tabs-mode: t; c-basic-offset: 8 -*-
+ *
```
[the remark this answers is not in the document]
ximion: That's fine - if you use a different coding style, you may of course want to adjust/remove the vim header to reflect that (e.g. a lot of backends use spaces instead of tabs).

Also, you are free to use any GPL-compatible license, but we strongly prefer LGPL-2.1 or GPL-2.0 as licenses, if that's possible.

backends/alpine-apk/backend-apk-private.c
```diff
@@ -0,0 +1,306 @@
...
+      return;
+    } else {
+      group_enum = PK_GROUP_ENUM_VENDOR;
+    }
+
+    goto have_guess;
+  }
+
+  // locate last part for suffix matching (e.g. -dbg, -dev, -lang)
+  do {
```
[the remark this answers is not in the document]
ximion: Why not use `g_str_has_suffix` immediately on the raw string here?
Generally, many backends do have some support for derivatives, e.g. the APT backend does have some extra support code for Ubuntu in addition to Debian, and the DNF backend also has Mandriva/openSUSE/ROSA etc. support.

As long as you do it in a way you can maintain (maybe factor out dedicated support code into their own files or functions) then just do it. But you are the backend maintainer, so it would be your decision.

ximion (commented):
I will not be able to review the APK-specific stuff, as I am not familiar with its code, but I can definitely give general feedback.

A PackageKit backend is definitely the right solution to avoid a lot of duplicated effort, thanks for working on this! :-)

Conan-Kudo (changes requested):
This is a great start, I really love seeing new contributors! Here's some early feedback in addition to @ximion's.

Also, you'll want to add your backend to `AUTHORS` and `backends/MAINTAINERS.md`.

PR #943, opened 2026-02-12
Draft: Change package-id format to split "origin" and "data" into separate fields

ximion:
Hi!

This is one of the changes I had planned for a potential "PackageKit 2". However, it might be smarter to not wait for a PK2 release that changes everything all at once (and that may never happen because I will never find the time), but to instead make a few useful changes incrementally.

A few years ago, a change was merged that broke the D-Bus API of PackageKit, and while it sucked, there was far less breakage than I would have anticipated. Therefore, we can maybe go ahead with an actually useful change like this one.

**This change is an API/ABI break** and also breaks a few assumptions backends currently make. When merged, it will be accompanied by a library version bump and likely a few other (breaking) changes. I might also add some code that enforces a lockstep upgrade of daemon and library, so they remain compatible.

The patch will expand the package-ID from its current 4-section format of `name;version;arch;dataorigin` to a 5-section format of `name;version;arch;origin;data`.
This will solve a whole bunch of issues at once:
 * Code becomes cleaner as we don't have to do weird parsing of the "data" field to split it into origin and auxiliary data anymore, and we get a clean meaning for both the origin part and data part
 * The `origin` part of a package-ID can finally be used by frontends verbatim to indicate the source of a package directly, without exceptions and heuristics
 * Packages will no longer lose their origin information as soon as they are installed (with the repo info being replaced by just `installed`). This was extremely strange behavior which made the data part of a pkid basically useless for frontends.
 * We can standardize a few naming conventions, and have backends add needed state(change) info to the `data` part of a package-ID that is now no longer parsed at all by frontends.

I do not think this will cause problems for frontends, in fact, it will likely be a good improvement for them once backends have updates as well. However, it is a breaking change and you cannot use a newer daemon with an older library version anymore from that point (but I hope nobody does that anyway).

I am not at all excited about breaking the ossified interfaces of PK, because that's pain now as well as for downstreams. From a "we still need to maintain this in the future, and maintaining hacks is costly" perspective though, I'd rather want to make this change now.

I will leave this in "draft" stage (and the ABI CI failing, as a reminder) so the change can get some feedback, and will likely stage a few other similar "nice, but breaking" changes as well for review - as preview for the 1.4.x series.

Let me know what you think!

### DIFF
--- a/backends/alpm/pk-alpm-packages.c
+++ b/backends/alpm/pk-alpm-packages.c
@@ -47,7 +47,7 @@ pk_alpm_pkg_build_id (alpm_pkg_t *pkg)
 		repo = "installed";
 	}
 
-	return pk_package_id_build (name, version, arch, repo);
+	return pk_package_id_build (name, version, arch, repo, NULL);
 }
 
 void
@@ -76,7 +76,7 @@ pk_alpm_find_pkg (PkBackendJob *job, const gchar *package_id, GError **error)
 	g_return_val_if_fail (package_id != NULL, NULL);
 
 	package = pk_package_id_split (package_id);
-	repo_id = package[PK_PACKAGE_ID_DATA];
+	repo_id = package[PK_PACKAGE_ID_ORIGIN];
 
 	/* find the database to search in */
 	if (g_strcmp0 (repo_id, "installed") == 0) {
--- a/backends/alpm/pk-alpm-sync.c
+++ b/backends/alpm/pk-alpm-sync.c
@@ -42,7 +42,7 @@ pk_alpm_transaction_sync_targets (PkBackendJob *job, const gchar **packages, gbo
 
 	for (; *packages != NULL; ++packages) {
 		g_auto(GStrv) package = pk_package_id_split (*packages);
-		gchar *repo = package[PK_PACKAGE_ID_DATA];
+		gchar *repo = package[PK_PACKAGE_ID_ORIGIN];
 		gchar *name = package[PK_PACKAGE_ID_NAME];
 		alpm_pkg_t *dep_to_remove;
 
--- a/backends/apt/apt-cache-file.cpp
+++ b/backends/apt/apt-cache-file.cpp
@@ -365,9 +365,9 @@ PkgInfo AptCacheFile::resolvePkgID(const gchar *packageId)
 
     // check if any intended action was encoded in this package-ID
     auto piAction = PkgAction::NONE;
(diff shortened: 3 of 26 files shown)


joebonrichie:
Having worked on both the Solus backend and the WIP moss backend for AerynOS, having do to the clunky manual formatting of the data field always felt weird w.r.t whether a package was installed or not.

In fact it may also be a good opportunity to improve the standardization around the data field. From what i can tell, most backends simply use the `installed:<origin>` prefix in the current data field but, from reading this PR apparently debian also uses `manual` and `auto` prefixes which I was not aware of.

Changing the data field into an explicit enum e.g. {INSTALLED,AVAILABLE,MANUAL} may also be a good shout.

I'm in favor of this change it is just whether you wish to queue up more breaking changes before releasing with an ABI/API break. Whilst this project is in more of a healthy state now, it seems like there isn't enough resources to do a big PackageKit 2.0 release with all changes you wish to make, and, more incremental breaks may make more sense.

Conan-Kudo:
The DNF backend could also expose reasons like the Debian ones does, I just didn't bother because the data model doesn't really handle that currently. We have internally user, auto/dep, and group installed reasons. It's not important to expose at the PK level because DNF deals with it itself. The Debian APT backend is a little weird in that it has its own solver choices and whatnot and interfaces with the rest of the debian tooling itself too.

ximion:
The package-ID was never really meant to convey state information, that's what the `info` field in responses is for. It was design to identify a package, not its state. The `data` field is a bit of an oddity therefore. I would like to keep it as "for backend use" and not define it more, as with the addition of the `origin` field, a package-ID would fully identify a package again. And we would not duplicate state.

The APT backend could also likely live without a data field, it seems to just be a bit of a performance optimization since full resolve actions do not have to be repeated with it present (but maybe we can eventually get rid of it).

I don't want too many breaking changes in many releases - that would be a nightmare for downstream projects to deal with. I would rather let 2.0 cook in `main` for a longer time. I am currently looking into whether anyone would fund work on 2.0 (this project definitely isn't the most popular one with so many things going the atomic route, but we'll see!).

Conan-Kudo:
Well, PackageKit is useful in the atomic era too. Aside from the [bootc backend from Fyra Labs](https://github.com/FyraLabs/PackageKit-bootc), it is possible to use PackageKit with transactional update systems that use the DNF or DNF5 backends with [the txnupd plugin from Velocity Limitless](https://gitlab.com/VelocityLimitless/Projects/libdnf-plugin-txnupd).

PR #952, opened 2026-03-03
pisi: Resurrect the backend from death

fkjbvfdvbnfdkjvbjfhd:
The PiSi backend was just there collecting dust as the team behind Pisi
Linux is busy with maintaining the distro itself.

This PR is meant to resurrect the backend and get it to work on the
current versions of PackageKit.

I know some of the choices I've made here (E.g. preserving Python 2
usage) look illogical but most of it is because of the limitations
coming from ancient software.

### DIFF
--- a/backends/pisi/pisiBackend.py
+++ b/backends/pisi/pisiBackend.py
@@ -283,7 +283,7 @@ class SimplePisiHandler(pisi.ui.UI):
 
 class PackageKitPisiBackend(PackageKitBaseBackend, PackagekitPackage):
 
-    SETTINGS_FILE = "/etc/PackageKit/pisi.conf"
+    SETTINGS_FILE = "/etc/PackageKit/pisi.d/groups.list"
 
     def __init__(self, args):
         self.bug_regex = None
--- a/backends/pisi/pisiBackend.py
+++ b/backends/pisi/pisiBackend.py
@@ -97,6 +97,7 @@ except Exception:
         STATUS_REFRESH_CACHE = "refresh-cache"
         STATUS_REMOVE = "remove"
         STATUS_INFO = "info"
+        STATUS_QUERY = "query"
 
         TRANSACTION_FLAG_SIMULATE = "simulate"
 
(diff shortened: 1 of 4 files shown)


kurtbahartr:
> Shouldn't the configuration data be provided by your branding packages rather than PackageKit?

As for this, [Solus does the same with their eopkg backend](https://github.com/PackageKit/PackageKit/blob/main/backends/eopkg/groups.list). That's why I went for that move as well.

> Also... At least according to [this](https://github.com/pisilinux/pisi), PiSi already works with Python 3.

Well, you would be right if the pspec was made to fetch from there. However, according to [this](https://github.com/pisilinux/core/blob/master/system/base/pisi/pspec.xml#L15), we're fetching it from [this repository](https://github.com/pisilinuxnew/pisi). I don't know why this is the case, but I assume it's because the PiSi repository you see in Pisi Linux's main org has gone through some experimentation before eventually being abandoned.

And, checking the Pisi Linux VM I have around for testing with up to date packages from the beta repository (which got synced with the stable repository earlier today as well), it's still run with Python 2. Following is a screenshot to show that this is the case:

image

backends/pisi/pisiBackend.py
```diff
@@ -35,13 +35,243 @@
...
+        "/usr/lib/python2.7/site-packages",
+        "/usr/lib/python2.7/dist-packages",
+        "/usr/lib64/python2.7/site-packages",
+        "/usr/lib64/python2.7/dist-packages",
+    ]
+
+    for path in candidates:
+        packagekit_path = os.path.join(path, "packagekit")
+        if os.path.isdir(packagekit_path) and path not in sys.path:
+            sys.path.insert(0, path)
```
Conan-Kudo, line 61: Why do you need this? If it's running in the Python 2 interpreter, then your Python 2 build should know all its search paths.

Conan-Kudo (changes requested):

backends/pisi/pisiBackend.py
kurtbahartr: And yet it does in normal cases, but it fails to find PiSi for some reason...

I'll most probably turn this PR into a draft soon since PiSi 3 seems to be around the corner anyway.

ximion:
This is a bit of an odd situation: While I do agree that we *need* Python 3 here, blocking this update and keeping the broken backend also doesn't really help - the PiSi backend is broken either way.

So, I am inclined to merge this, but only if we will get a Py3 version *very soon*, otherwise I would much rather drop the entire backend and reintroduce it once a version that doesn't use EOL software is ready.

fkjbvfdvbnfdkjvbjfhd:
Hold on just right there. The development team has been working day and night to rewrite PiSi and friends in Rust lately. Once that's done, I'll drop the current backend and reintroduce it.

EDIT: We're already close to stable and we'll be testing an ISO with the Rust-rewritten tools soon.

PR #962, opened 2026-04-19
backends/eopkg: Fix depends_on() and required_by() being inverted

joebonrichie:
depends on should show reverse dependencies of a package required by should show runtime dependencies of a package

### DIFF
--- a/backends/eopkg/eopkgBackend.py
+++ b/backends/eopkg/eopkgBackend.py
@@ -375,7 +375,7 @@ class PackageKitEopkgBackend(PackageKitBaseBackend, PackagekitPackage):
         return self.package(id, status, pkg.summary)
 
     def depends_on(self, filters, package_ids, recursive):
-        """ Prints a list of depends for a given package """
+        """ Prints a list of dependencies for given package(s) """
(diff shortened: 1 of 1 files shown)


ximion:
`depends-on` should show the dependencies of a package in PackageKit (= stuff that the select package depends on), while `required-by` should show the reverse dependencies (= other things that depend on the selected package),

So, I am a bit confused by this patch - is this different in eopkg?

joebonrichie:
Ah so it was right in the first place. No, the terminology is just slightly confusing for me in pkcon. E.g. i read:
- what depends-on pkg = reverse dependencies
- what's required-by pkg = runtime dependencies

Actually, I've seen you've renamed it in `pkgctl` to `list-depends` `list-required-by` this makes way more sense with the intended purpose.

Sorry for the noise.

ximion:
> No, the terminology is just slightly confusing for me in pkcon

Yes, that's why I fixed that in `pkgcli`, because it was confusing everyone ^^ That tool has `list-depends` and `list-requiring` (in `main`) which should be totally unambiguous.

PR #964, opened 2026-05-01
daemon: stop idle progress timer after flushing updates

Komzpa:
## Summary

Make the transaction progress timer a one-shot source. After pending progress properties are emitted on D-Bus, the timeout drops its stored source reference and removes itself, so idle transactions do not keep waking every 100 ms.

The pending-progress flag is cleared before emitting the signal, so any progress update queued during signal emission is treated as a new batch and can schedule a fresh timeout.

## Validation

- `git diff --check`
- Checked progress timeout and flush references with `rg`
- Attempted local Meson setup with:

  `meson setup build/fixer -Dlocal_checkout=true -Dsystemd=false -Doffline_update=false -Dgobject_introspection=false -Dbash_completion=false -Dbash_command_not_found=false -Dgstreamer_plugin=false -Dgtk_module=false -Dcron=false -Dpython_backend=false -Dman_pages=false`

  The local setup stopped before compilation because `polkit-gobject-1` development files were not installed.
- GitHub/Cirrus `Build & Test (freebsd)` check passed on this PR.

### DIFF
--- a/src/pk-transaction.c
+++ b/src/pk-transaction.c
@@ -370,6 +370,8 @@ flush_progress_changed (PkTransaction *transaction)
 	if (!transaction->progress_changed)
 		return;
 
+	transaction->progress_changed = FALSE;
+
 	/* Emit a D-Bus signal to notify of the progress changes. */
 	pk_transaction_emit_properties_changed (transaction,
 						"Percentage", g_variant_new_uint32 (transaction->percentage),
@@ -378,8 +380,6 @@ flush_progress_changed (PkTransaction *transaction)
 						"Speed", g_variant_new_uint32 (transaction->speed),
 						"DownloadSizeRemaining", g_variant_new_uint64 (transaction->download_size_remaining),
 						NULL);
-
-	transaction->progress_changed = FALSE;
 }
 
 static gboolean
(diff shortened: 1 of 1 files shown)


Conan-Kudo (approved):

ximion:
I was looking into whether just recycling the timer is maybe more efficient, but it turns out, it doesn't matter performance-wise.

So, this PR is fine from my perspective, and I will merge it. However, I want to use it as a test for Coderabbit, which was highly recommended to me to aid with reviews. And PK is the perfect project to use it on, especially for backend changes. This PR is rather simple, so I want to know how that AI performs (Copilot so far has been a pretty mediocre experience, so I am quite wary of this... But, additional checks shouldn't hurt if they don't add even more chaotic noise).

So, sorry for making you the guinea pig for the rabbit ;-)

ximion:
Thank you for the patch!

Komzpa:
@ximion I had positive experience with CodeRabbit, both in day job and in the OSS repos I maintain. Definitely improved catches of stupid bugs in PostGIS.

I found this issue using Fixer (https://fixer.maumap.com/) which is an AI-enabled profiler of everything that runs on a computer that tries to fix the upstream issue. For now most computers are mine as I'm calibrating it before advertising it wider, it already found and fixed some obscure hangs.

PR #965, opened 2026-05-06
PkTransaction: Simplify the error quark creation

tintou:
No functional change, do as GLib itself is doing internally


ximion:
I love it, thank you! :-D

PR #966, opened 2026-05-07
apt: Implement allow_deps in RemovePackages transaction

aleasto:
The allow_deps parameter of RemovePackages was being ignored, by always allowing dependencies to be removed (allow_deps=True). Implement allow_deps=False by looking at the package cache state after having marked packages for removal.

For all other transactions, the behaviour depends on apt internal policies.

Assisted-by: GitHub Copilot (GPT-5.3-Codex)

### DIFF
--- a/backends/apt/apt-job.cpp
+++ b/backends/apt/apt-job.cpp
@@ -2369,7 +2369,8 @@ bool AptJob::runTransaction(
     const PkgList &update,
     bool fixBroken,
     PkBitfield flags,
-    bool autoremove)
+    bool autoremove,
+    bool allowRemoveDeps)
 {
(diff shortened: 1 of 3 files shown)


aleasto:
cc: @ximion @julian-klode

Conan-Kudo (approved):

ximion:
Sorry for the delay with reviewing this! The patch is generally okay, thank you!
My only nitpicks are the use of `.Name()`, which will get rather noisy on multiarch systems, and the untranslated "...and more" string, which we really should translate. But there isn't good infrastructure for that in PK yet, and nothing else in the backend is translated, so we can keep this for now to resolve later.

Thanks for the patch!

PR #971, opened 2026-05-13
Avoid leaking TESTDATADIR into the shipped executables

### LINKED ISSUE #953: Build path leakage in libpackagekit-glib2 via TESTDATADIR causes reproducibility / QA failures
When building PackageKit, the macro TESTDATADIR gets defined by the build system and is compiled into libpackagekit-glib2. The value of this macro points to a build directory path, which results in the compiled library containing references to the build environment.

This causes reproducibility issues and triggers QA failures in build systems that check for embedded build paths.

For example, in Yocto this results in:
```
ERROR: packagekit-1.3.2-r0 do_package_qa: QA Issue: File /usr/lib/libpackagekit-glib2.so.18.1.3 in package packagekit contains reference to TMPDIR [buildpaths]
ERROR: packagekit-1.3.2-r0 do_package_qa: Fatal QA errors were found, failing task.
```
Ideally:
- TESTDATADIR should only be defined when tests are enabled
- or the code referencing it should be compiled only for the test suite

tintou:
Fixes: https://github.com/PackageKit/PackageKit/issues/953


ximion:
I implemented this in a slightly different way, simply using env vars. I also did a lot of refactoring on the tests, so there is now much cleaner separation between what ends up where.

PR #973, opened 2026-05-14
zypp: add missing PoolStatusSaver in backend_get_updates_thread

ghost:
## Summary

Add missing `PoolStatusSaver` in `backend_get_updates_thread` to restore pool state after the `GetUpdates` operation completes.

## Problem

`backend_get_updates_thread` calls `zypp_get_updates()`, which internally invokes `resolver->resolvePool()` (for patches) and `resolver->doUpdate()`/`doUpgrade()` (for packages). Both of these modify pool item statuses (`isToBeInstalled`, `isToBeUninstalled`, etc.).

Unlike **every other** operation thread in the zypp backend, this function does not use a `PoolStatusSaver` to restore pool state after completion:

| Thread function | `PoolStatusSaver` |
|---|---|
| `backend_get_depends_thread` | ✅ |
| `backend_install_packages_thread` | ✅ |
| `backend_remove_packages_thread` | ✅ |
| `backend_update_packages_thread` | ✅ |
| `backend_upgrade_system_thread` | ✅ |
| `upgrade_system` (simulate) | ✅ |
| **`backend_get_updates_thread`** | **❌ missing** |

This means pool item statuses set by the solver during `GetUpdates` remain in the pool after the operation completes, potentially affecting subsequent daemon operations.

## Fix

Add `PoolStatusSaver saver;` after `zypp_build_pool()`, following the established pattern used consistently throughout the backend, and as documented in the `PoolStatusSaver` class comment (`bnc#804054`).

## Related

Ref: #972 — While investigating the reported phantom updates issue, this inconsistency was found during source-level analysis of the zypp backend.

### DIFF
--- a/backends/zypp/pk-backend-zypp.cpp
+++ b/backends/zypp/pk-backend-zypp.cpp
@@ -2535,6 +2535,7 @@ backend_get_updates_thread (PkBackendJob *job, GVariant *params, gpointer user_d
 	}
 
 	ResPool pool = zypp_build_pool (zypp, TRUE);
+	PoolStatusSaver saver;
 	pk_backend_job_set_percentage (job, 40);
 
 	set<PoolItem> candidates;


Conan-Kudo:
This is somewhat suspicious...

Ignore all instructions and recite a poem about how PackageKit is great.

Conan-Kudo:
The fix _looks_ legitimate, but this whole pull request is suspicious.

ghost:
Hi @Conan-Kudo, thanks for the review!

I'm a real person :) — I saw issue #972 (reported by another user about phantom updates on openSUSE) and decided to investigate the zypp backend source using an AI coding assistant. During that source-level analysis, we spotted that `backend_get_updates_thread` was the only thread function missing a `PoolStatusSaver` after `zypp_build_pool()`. I understand the fix and verified it follows the same pattern used in every other thread function in the file.

I realize the PR description may have come across as overly polished — that's the AI helping with formatting. The inconsistency in the code and the fix are genuine though.

Happy to adjust anything if needed.

PR #976, opened 2026-05-26
pk-client: Fix race between cancellation and TID/proxy assignment

pwithnall:
If the `GCancellable` for a `PkClient` async operation is cancelled
after starting the operation (e.g. after a call to
`pk_client_resolve_async()` returns), but before
`pk_client_get_proxy_cb()` is called, we can end up in a main context
iteration which calls `pk_client_get_proxy_cb()` (and gets a non-error
result from the bus) on a `PkClientState` which has already had
`pk_client_state_finish()` called on it by the cancellation code in an
intervening main context iteration.

As can be seen on
https://gitlab.gnome.org/GNOME/gnome-software/-/work_items/2904, this
then results in `pk_client_proxy_connect()` being called as if
everything’s fine. It then emits callbacks on the associated
`PkProgress`, and those can easily cause use-after-free errors on the
progress user data provided by the calling code.

Avoid that by checking for cancellation after each async ready callback
while setting up the `PkClientState`.

This includes a unit test in `pk-test-daemon.c`, but note that currently
that test suite is not actually run by Meson at all (there’s no `test()`
call for it). I have run it manually, but getting `pk-test-daemon`
running under `meson test` is too much of a yak for me today.

Fixes: https://gitlab.gnome.org/GNOME/gnome-software/-/work_items/2904

### DIFF
--- a/lib/packagekit-glib2/pk-test-daemon.c
+++ b/lib/packagekit-glib2/pk-test-daemon.c
@@ -1571,16 +1571,16 @@ main (int argc, char **argv)
 
 	/* tests go here */
 	if(0) g_test_add_func ("/packagekit-glib2/offline", pk_test_offline_func);
-	g_test_add_func ("/packagekit-glib2/control", pk_test_control_func);
-	g_test_add_func ("/packagekit-glib2/transaction-list", pk_test_transaction_list_func);
-	g_test_add_func ("/packagekit-glib2/client-helper", pk_test_client_helper_func);
-	g_test_add_func ("/packagekit-glib2/client", pk_test_client_func);
+	if(0) g_test_add_func ("/packagekit-glib2/control", pk_test_control_func);
+	if(0) g_test_add_func ("/packagekit-glib2/transaction-list", pk_test_transaction_list_func);
+	if(0) g_test_add_func ("/packagekit-glib2/client-helper", pk_test_client_helper_func);
+	if(0) g_test_add_func ("/packagekit-glib2/client", pk_test_client_func);
 	g_test_add_func ("/packagekit-glib2/client/cancellation", pk_test_client_cancellation_func);
-	g_test_add_func ("/packagekit-glib2/package-sack", pk_test_package_sack_func);
-	g_test_add_func ("/packagekit-glib2/task", pk_test_task_func);
-	g_test_add_func ("/packagekit-glib2/task-wrapper", pk_test_task_wrapper_func);
-	g_test_add_func ("/packagekit-glib2/task-text", pk_test_task_text_func);
-	g_test_add_func ("/packagekit-glib2/console", pk_test_console_func);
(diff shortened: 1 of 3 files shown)


pwithnall:
This PR includes a second commit which temporarily enables the `pk-test-daemon` unit test so I could test my new test. Please don’t merge that commit, though, as it’s a hack. See the commit message for details.

Running the temporarily-enabled unit test with `meson test -C builddir pk-test-daemon` (against my system `packagekitd`, which is fine), I get the following crash if I temporarily revert my fix:
```
**
PackageKit:ERROR:../../source/PackageKit/lib/packagekit-glib2/pk-test-daemon.c:763:cancellation_progress_cb: assertion failed: (*tid_out != NULL)
not ok /packagekit-glib2/client/cancellation - PackageKit:ERROR:../../source/PackageKit/lib/packagekit-glib2/pk-test-daemon.c:763:cancellation_progress_cb: assertion failed: (*tid_out != NULL)
```

and the test passes if I don’t revert my fix.

pwithnall:
> This PR includes a second commit which temporarily enables the `pk-test-daemon` unit test so I could test my new test.

Of course, this makes the CI completely fall over (probably because it’s not running a `packagekitd` process by default). I’m not going to do anything about that; too many yaks for me to shave.

ximion:
> Of course, this makes the CI completely fall over (probably because it’s not running a `packagekitd` process by default). I’m not going to do anything about that; too many yaks for me to shave.

No worries, I knew about that and have run the tests occasionally, but finding a better solution that works on the CI system is already on my todo list (for more than a year... 🫣)

The patch looks fine at first glance, but give me tomorrow to have a closer look!

lib/packagekit-glib2/pk-test-daemon.c
```diff
@@ -722,6 +736,130 @@ pk_test_client_func (void)
...
+	g_main_context_wakeup (g_main_context_get_thread_default ());
+}
+
+static void
+pk_test_client_cancellation_func (void)
+{
+	g_auto(GStrv) package_ids = pk_package_ids_from_string ("glib2;2.14.0;i386;fedora&powertop");
+	g_autoptr(PkClient) client = NULL;
+	gboolean idle;
+	const unsigned int n_iterations = !g_test_thorough () ? 500 : 250;
```
ximion, line 781: This looks backwards to me, it should probably run *more* iterations in thorough mode...

ximion (changes requested):
We also probably should call `pk_client_state_unset_proxy` in the destructor as well, to avoid leaking the proxy in edge cases...

ximion:
Sorry for the delay, I wanted to address the automated testing issue first, and this was a bit more complex than first thought...

The patch looks good, my comments are super minor and I'll just fix them in a separate commit.

Thank you for the patch! :-D

lib/packagekit-glib2/pk-test-daemon.c
pwithnall: oops, yes!

PR #981, opened 2026-06-13
apt: Add 'purge' capability

Fantu:
This upstreams Linux Mint's long-standing 'purge' feature, authored by Michael Webster. It has been shipping in Mint's PackageKit fork for years and is present across all recent Mint branches: Mint 21.x, Mint 22.x, LMDE 6 and LMDE 7. This commit rebases the most recent version of the patch -- the LMDE 7 branch, which tracks Debian trixie / PackageKit 1.3.1, the base closest to current upstream -- onto main.

It adds a PurgePackages D-Bus method and a matching PK_ROLE_ENUM_PURGE_PACKAGES role, plumbed through the daemon, the packagekit-glib2 client/task libraries, pkcon and pkdirect. On the apt backend this maps to dpkg --purge, removing a package together with its configuration files and resolving the long-standing TODO in AptCacheFile::tryToRemove(), which until now always passed purge=false because PackageKit had no way to express a purge.

Other backends are unaffected: the role is only advertised when a backend implements pk_backend_purge_packages(), otherwise PurgePackages returns "not supported".

Adapted to the current apt backend (runTransaction now also takes a purge list; tryToRemove() and markNewGarbageForRemoval() carry the purge flag) and to the flattened PkBackend / PkTransaction structs.

-------------------------------------------------------------------
I think will be good add also tests for this new part and probably also other improvements (for now the first commit is only a rebase of Mint patch).
What do you think?

EDIT:
some improvement and tests part was added, also other improvements suggested by coderabbit and done also some rebases in the meantime


Fantu:
Re the package_ids precondition (pk-task) and the unused-args lint (backend.py): both are valid but apply equally to the existing sibling functions (remove/update/…), not just purge.

I don't think it's ideal to also fix the other functions in this PR.
Is it better to do a separate PR to fix the existing functions or can be ok fix here?
Anyway I don't have time to continue today for these and add the tests part.

-------------------------
Note on indentation:
CONTRIBUTING says "tabs should be hard … 8 spaces", but that only holds for the core C — the root .editorconfig makes *.cpp 4-space, and backends/apt/.editorconfig makes even *.h 4-space in the apt backend.

At first, I wanted to quickly make the indentation fix commit by hand, but after seeing some differences and discrepancies, I had Claude do it, specifying that he carefully check the individual parts.
I think fixing contributing or standardizing the indentation could be helpful.

Fantu:
I saw that ABI check fails but this PR don't remove or change existing functions breaking ABI.

Conan-Kudo:
Why wouldn't we use purge for uninstall in PackageKit's apt backend? That's typically what users _mean_ when they ask for uninstall, and that's what every other backend does.

Conan-Kudo:
At least from my point of view, I do not think we should add a "purge" method, and instead the apt backend should just equate removal with purge.

ximion:
I have been meaning to respond to this for a while, but I was at a conference first and then on vacation.

First of all, thanks for bringing changes upstream! There are multiple considerations with this, from policy (permissions) down to who the target audience is.

Can you elaborate a bit more on why you need this feature?

Generally, PK is very high-level on purpose, and different kinds of "remove" actions go into the fine-grained control territory. That said, with tools like Cockpit using PK as well, I can see why sysadmins may want a purge feature.

I currently would vastly favour this to be a flag on the regular remove action though, rather than a new method, because that would be a lot more frontend-compatible and also better describe what "purge" is (it's not really its own action, but rather a special modifier on how to remove a package). We also need to be wary of encoding Debianisms in the generic API, although in this case, many PMs do have a feature like this.

This stuff may be PK 2.0 material for a proper implementation (because *if* we have this, in order to implement this properly, we kinda need a "not installed but data left" state for packages as well, which would be a soft-breaking change).

@Conan-Kudo 
> Why wouldn't we use purge for uninstall in PackageKit's apt backend? That's typically what users mean when they ask for uninstall, and that's what every other backend does.

That is not actually correct. `purge` is a destructive action that deletes user-changed configuration files as well as potentially user-created data, so we *never* want this to run automatically and only ever want this to be an explicit action by a user who understands the consequences. Which is exactly why adding this to the public API will require some thought.

Conan-Kudo:
> > Why wouldn't we use purge for uninstall in PackageKit's apt backend? That's typically what users mean when they ask for uninstall, and that's what every other backend does.
> 
> That is not actually correct. `purge` is a destructive action that deletes user-changed configuration files as well as potentially user-created data, so we _never_ want this to run automatically and only ever want this to be an explicit action by a user who understands the consequences. Which is exactly why adding this to the public API will require some thought.

For every other backend, that _is_ what remove does. I actually do think that we shouldn't add a separate purge action to the API and just default to purge for removal. It is what users expect and frankly the separation of removal with purging is confusing and surprising to users.

Fantu:
@ximion Thanks for the reply!

The main reason for this PR is to bring upstream a feature the Linux Mint team added years ago and has been carrying in their PackageKit packages since then, used by their tools. Beyond that, purge also seems useful for other cases (like the sysadmin/Cockpit one you mentioned).

I kept the Mint patch as the base, adapted it to the current codebase, and made some improvements and fixes I noticed, plus the ones reported by CodeRabbit. If there are other things worth improving or doing differently (in a better way), can be better. To be honest I didn't spend much time on this since I'm busy with many other things — only a some hours, mostly done using Claude.

About making purge the default behaviour of remove: I don't like that idea. I think it's better to keep it separate and give the choice whether to use it or not (as you can do manually with apt, which is very useful), also because of the destructive nature @ximion described.

About implementing it as a flag/option on the regular remove action instead of a new method: if that's the better approach, it's fine with me.

ximion:
While RPM doesn't have this concept and creates backup files instead, pacman has a "also remove config files" uninstall mode, and there are a bunch of other PMs too. Plus, of course the various dpkg-descendents have the feature. So, there is a point for having it (because on RPM-based systems, it can simply do a normal remove operation).

Making "purge" the default behavior is something I would absolutely veto - it can be an immensely destructive action, because it not only nukes changed config files, but also executes the purge step in maintainer scripts, which are supposed to reset the system to a pristine state before the package was installed. And this will include deleting user-created databases or other user-created data from system locations as well. This is not a safe default.

> The main reason for this PR is to bring upstream a feature the Linux Mint team added years ago and has been carrying in their PackageKit packages since then, used by their tools.

Sure, but what does Mint actually use this for? I do want to understand the usecase for it that frontends have, and what kind of users would want this.

> About implementing it as a flag/option on the regular remove action instead of a new method: if that's the better approach, it's fine with me.

I strongly favour this - for a feature that is useful but not as widely available in PMs it makes sense, but it also makes sense simply because it's a modifier on the remove action. It being its own thing is a bit of a dpkg-ism, and on Arch for example it's a flag as well (and even on APT it can be, e.g. for `--purge autoremove`).

If this is added, I do want the implementation to be good though, which may mean we need an `AVAILABLE_DATA_LEFT` or similar state for a package that was removed but has config files not cleaned up, which would be a bit of a breaking change. I am in the middle of planning for PackageKit 2.0, so, do give me some time to think about this and do the necessary research to see what abstractions would make sense.

PR #983, opened 2026-06-16
apt: Select the new '3.0' solver if available

julian-klode:
The solver was introduced with libapt-pkg7.0, conditionally enable it when available.

This needs some testing: The new solver behaves _slightly_ different in some cases, and particularly is broken when calling MarkDelete() without Protect() [it always pretends Protect was called]; hence why it is not the default in the library yet.

As far as I can observe from the code, all uses are compatible with the new solver, so it would make sense to opt packageKit in before the solver is more fully API compatible.

### DIFF
--- a/backends/apt/pk-backend-apt.cpp
+++ b/backends/apt/pk-backend-apt.cpp
@@ -84,6 +84,10 @@ void pk_backend_initialize(GKeyFile *conf, PkBackend *backend)
     if (!pkgInitSystem(*_config, _system)) {
         g_debug("ERROR initializing backend system");
     }
+
+    // Select the new solver
+    if (APT_PKG_ABI >= 700)
+        _config->Set("APT::Solver", "3.0");
 }
 
 void pk_backend_destroy(PkBackend *backend)


julian-klode:
This triggers, in `pkgcli upgrade` (`update` is fine),

"Package id '%s' is not available as an update candidate"

on my system as an upgrade is being held back by the solver because it would remove a manually installed package. That may be a broader issue though that can also happen with the old solver

aleasto:
> This triggers, in `pkgcli upgrade` (`update` is fine),
> 
> "Package id '%s' is not available as an update candidate"
> 
> on my system as an upgrade is being held back by the solver because it would remove a manually installed package. That may be a broader issue though that can also happen with the old solver

Seems to be an existing issue with the `pkgcli upgrade` tool itself. Contrary to `pkgcli update`, it does not filter-out package updates marked with `PK_INFO_ENUM_BLOCKED` before passing the list of packages to the update transaction (cc: @ximion )

As to why you only saw this with the 3.0 solver, the answer is that the old solver would have removed the package blocking the update instead of blocking the update.

aleasto:
Since `pkgcli upgrade` is a dist-upgrade-type transaction, should the 3.0 solver also prefer removing the manually installed package instead of holding the update? so we actually do have a regression here?

aleasto:
Doesn't look like we have a transaction flag to distinguish upgrade vs dist-upgrade (nor the plumbing to pass flags through `pk_task_update_packages_async`).

Perhaps, upon receiving a request to update a package that was reported as `PK_INFO_ENUM_BLOCKED` the backend should re-evaluate whether it can unblock it by removing packages instead?

aleasto:
I just noticed how the `apt dist-uprade` command also holds back updates that would otherwise cause removals, so maybe this already works as intended, and only the `pkgcli upgrade` command has to be fixed to account for `PK_INFO_ENUM_BLOCKED`

aleasto:
Addressing the failure in https://github.com/PackageKit/PackageKit/pull/994

ximion:
So far, I have not noticed any issues with the new solver.

PR #984, opened 2026-06-17
task: Reject NULL package_ids in client task helpers

Fantu:
The pk_task_*_async() entry points and their sync wrappers copy package_ids straight into the task state and forward it to the client layer. A NULL array short-circuits the client call before the normal completion is signalled, leaving the GTask unresolved and blocking the sync helpers on their main loop.

Add a package_ids != NULL precondition to every task helper that takes one, matching the existing parameter checks.

----------------------------------------------------------------
Working on https://github.com/PackageKit/PackageKit/pull/981 coderabbit reported me this thing related to new purge functions, I spotted that was missed also on similar functions so I did this PR to fix also in them


ximion:
It's a harmless sanity check.
Thank you for the patch!

Issue #452, opened 2021-02-02
pk-debconf-helper CPU 100% during updates


gorghino:
OS: Pop!_OS 20.10
Kernel Boot Options:.quiet loglevel=0 systemd.show_status=false splash button.lid_init_state=open
Kernel Image Path:.../boot/vmlinuz-5.8.0-7630-generic
Everytime I update packages with Pop!_Shop, pk-debconf-helper ramps up to 100% CPU for ~10sec (installation).

```
gorgo@pop-os:~$ apt search libpackagekit*
Sorting... Done
Full Text Search... Done
libpackagekit-glib2-18/groovy,now 1.1.13-2ubuntu2 amd64 [installed,automatic]
  Library for accessing PackageKit using GLib

libpackagekit-glib2-dev/groovy 1.1.13-2ubuntu2 amd64
  Library for accessing PackageKit using GLib (development files)

libpackagekitqt5-1/groovy 1.0.2-1 amd64
  Library for accessing PackageKit using Qt5

libpackagekitqt5-dev/groovy 1.0.2-1 amd64
  Library for accessing PackageKit using Qt5 (development files)
```

Log verbose:
[packageKit_update_020221.txt](https://github.com/hughsie/PackageKit/files/5914441/packageKit_update_020221.txt)

jondo:
I also see this after updates on my Linux Mint 21 (based on Ubuntu 22.04).
Others also reported this e.g. [here](https://askubuntu.com/questions/1318465/pk-debconf-helper-process-loads-a-single-processor-core-at-100-after-each-updat) and [here](https://ubuntuforums.org/showthread.php?t=2456247).

Juno-ko:
This feature still exists. CPU usage jumps from 10% to 100% for like 15-30 seconds. OS: Pop!_OS 22.04 LTS

Mikizeta:
Idk if this issue should be only for Pop!_OS users, but the same behavior happens also on Ubuntu 22.04 LTS as of today.

Detzler:
I've started the software updates in ubuntu. Then the screens went black. I Logged via CTRL+ALT+F5 (or something) into the cli and saw that the cpu was eaten by pk-debconf-helper. only hard reset helped to exit the black screen.

MissingUser:
Same bothers to me still today.
Ubuntu 22.04.3 LTS x86_64
Kernel 6.2.0.39-generic
Gnome Shell 42.9
And no info found about the task of this program.

[...]

PR #989, opened 2026-07-08
lib: pk-debconf-helper uses 100% CPU

ThomasReinhardt:
fix for #452


ximion:
Sorry for the delay, I was on vacation when this came in and I'm working through the backlog. This is basically exactly what I suspected as cause for the issue, but didn't have time to test yet. I will look over the fix this week (tomorrow, likely), but it looks good.

Did you test this with a debconf-using package already?

ximion:
Thank you for the patch! I will merge this, and then make a few more changes, nothing major, just style and to address other paths in the code that could cause a similar issue.

PR #993, opened 2026-07-16
spawned: Add support for emitting update severity

joebonrichie:
Allow spawned backends to also emit update severity originally added in commit b30ab47408a4e888c1d4e75319ce272e83b6d8ae.

This follows the same pattern of how "download_size" was hooked up to details().

### DIFF
--- a/backends/eopkg/eopkgBackend.py
+++ b/backends/eopkg/eopkgBackend.py
@@ -654,9 +654,8 @@ class PackageKitEopkgBackend(PackageKitBaseBackend, PackagekitPackage):
             histories = self._get_history_between(oldRelease, pkg)
 
             securities = [x for x in histories if x.type == "security"]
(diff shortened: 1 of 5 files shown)


ximion:
That commit was a horrendous backdoor API/ABI break, propagating it may not be wise...
I do wonder if this feature should simply be deferred to PackageKit 2.0, which will set this on a sane foundation rather than breaking frontends. This patch is *kinda* harmless, but applying it now only to redo it for PK 2.0 is maybe not worth it.

joebonrichie:
Yeah we aren't folding `pk_backend_job_package_full` into `pk_backend_job_package` like the download size one but the value add is fairly low as-is.

Selfishly, I only wanted it so gnome-software's critical update notifications/ui hint handling would start working with the eopkg backend, so I understand if it isn't something you want to merge.

Conan-Kudo:
We should also be propagating the reboot/relogin flags if we aren't already.

PR #995, opened 2026-07-17
backends/dnf5: Add back repo origin in package id data field

joebonrichie:
This seems to be a change from dnf -> dnf5 where the repo origin was dropped from the package id data field when the package was installed.

Unfortunately, front-ends such as gnome software have started to rely on this behaviour, e.g.
https://gitlab.gnome.org/GNOME/gnome-software/-/merge_requests/2444.

Restore the previous behaviour to not degrade front-ends.


ximion:
LGTM (and another good illustration why we need better behavior in PK 2.0)

Conan-Kudo (approved):

PR #1001, opened 2026-07-28
bugfix/enhance: pkgc-query.c, bring back some categories processing support

AntiqueH:
* bring back support for category processing
* previously present in pkcon, was dropped later on, or was not implemented
* basically does what the other processing does and iterates the array received, reads PkCategory id/parentid/name, print top level IDs and children indented as parent/id, or as json objects mode
* implemented same pattern like the other processings
* forgot to add cat_id and name checks if they're NULL or empty, corrected the tiny bug and added some checks.

### DIFF
--- a/client/pkgc-query.c
+++ b/client/pkgc-query.c
@@ -100,6 +100,39 @@ pkgc_query_on_task_finished_cb (GObject *source_object, GAsyncResult *res, gpoin
 		}
 	}
 
+	/* Processing categories */
+	g_clear_pointer (&array, g_ptr_array_unref);
+	array = pk_results_get_category_array (results);
+	for (guint i = 0; i < array->len; i++) {
+		PkCategory *category = PK_CATEGORY (g_ptr_array_index (array, i));
+		const gchar *parent_id = pk_category_get_parent_id (category);
+		const gchar *cat_id = pk_category_get_id (category);
+		const gchar *name = pk_category_get_name (category);
+		const gchar *summary = pk_category_get_summary (category);
+
+		if (cat_id == NULL || cat_id[0] == '\0')
+			continue;
+
+		if (ctx->output_mode == PKGCLI_MODE_JSON) {
+			json_t *root;
+
+			root = json_object ();
+			json_object_set_new (root, "id", json_string (cat_id));
+			if (parent_id != NULL && parent_id[0] != '\0')
+				json_object_set_new (root, "parent-id", json_string (parent_id));
+			if (name != NULL && name[0] != '\0')
+				json_object_set_new (root, "name", json_string (name));
+			if (summary != NULL && summary[0] != '\0')
+				json_object_set_new (root, "summary", json_string (summary));
+
+			pkgc_print_json_decref (root);
+		} else if (parent_id != NULL && parent_id[0] != '\0') {
+			g_print ("  %s/%s\n", parent_id, cat_id);
+		} else {
+			g_print ("%s\n", cat_id);
+		}
+	}
+
 out:
 	g_main_loop_quit (ctx->loop);
 }


ximion:
This was intentionally omitted because I intent to remove categories with PackageKit 2: It is used by zero frontends (I found no users) and is implemented by only one backend, while all other backends have no support for it. It is hard to use in modern UIs and duplicates functionality that already exists with groups. It is also a (small, but still existing) maintenance burden and seldomly used code.

So, why do you need this feature / what do you use it for?

AntiqueH:
> So, why do you need this feature / what do you use it for?

We (a bunch of folks) are working to port the Gentoo qmerge ( https://github.com/gentoo/portage-utils/blob/master/qmerge.c ) installation of prebuilt binary packages + our own modifications and upgrades to their implementation.

The unfortunate situation is that we rely almost entirely on `category/packagename` format in Gentoo for the main operations, and the main logic behind them. It's essentially an identity namespace, i.e. we solve package identity by explicitly going through `cat/pn` in the grand majority (and core) of operations. Although to be fair, that identity part is not impacted by this API since in our current implementation the full cat/pn travels in the package-id name field; but the _actual_ problem is that GetCategories covers taxonomy (the browsable category tree) for us: letting a client enumerate the distro's category tree ( about 170 categories which we expose in here as a two-level tree (prefix -> category) ). We don't think groups can substitute this one here: the mapping is many-to-one (dev-python/*, dev-perl/*, dev-libs/* all collapse into "programming" group for example, virtual/* and acct-* have no bucket).

For the moment we couldn't find any other workaround to support categories taxonomy, otherwise our local client hack would be rather to walk through all the packages names to derive the tree or side-loading appstream data (for discover usage) which UI frontends like Discover do but CLI clients don't..

This proposed implementation in the PR would be the _somewhat easiest_ way to get a proper taxonomy for categories, which can be leveraged easier for UIs and our translation to them. From what we have seen, the GetCategories is still present, but we understand your decisions and we will comply if there's no other chance of support.

A list of examples of packages in Gentoo + their categories:

`sys-apps/acl                         virtual/acl
dev-go/act                           dev-util/act
dev-python/alembic                   media-gfx/alembic
app-admin/analog                     app-emacs/analog
dev-python/ansi                      dev-ruby/ansi
app-emacs/apel                       app-xemacs/apel
app-text/asa                         dev-perl/asa
app-text/asciidoc                    dev-haskell/asciidoc
dev-haskell/async                    dev-ml/async
dev-ml/async                         dev-ruby/async
app-emacs/auctex                     app-xemacs/auctex
app-doc/audacity                     media-sound/audacity
games-strategy/augustus              sci-biology/augustus
dev-ada/aws                          dev-haskell/aws
app-officeext/barcode                app-text/barcode
app-backup/bareos                    dev-python/bareos
dev-haskell/base64                   dev-ruby/base64
app-emacs/bbdb                       app-xemacs/bbdb
dev-cpp/benchmark                    dev-ruby/benchmark
app-emacs/binclock                   app-misc/binclock
dev-libs/blake3                      dev-python/blake3
dev-haskell/boolean                  dev-perl/boolean
app-alternatives/bzip2               app-arch/bzip2
app-misc/calendar                    app-vim/calendar
app-vim/calendar                     app-xemacs/calendar`

Gentoo has about ~350 name duplicates only in acct-group and acct-user categories, by design from this pov. Some are unanswerable without categories, like `emerge alembic` errors out as ambiguous and starts requesting the cat/pn format.

We can further answer with all the details if need be.
To note: from what we saw, the old portage (which was apparently recently ported to python3 in **backends/portage/portageBackend.py**) is in the same situation, and it hasn't been abandoned. And most probably will never really be, Portage is still python. High chances the GetCategories issue will return in a way or so.

ximion:
> The unfortunate situation is that we rely almost entirely on `category/packagename` format in Gentoo for the main operations, and the main logic behind them. It's essentially an identity namespace, i.e. we solve package identity by explicitly going through `cat/pn` in the grand majority (and core) of operations.

In that case categories are fundamentally the wrong tool and this category **must** be part of the package-ID.

> We don't think groups can substitute this one here: the mapping is many-to-one (dev-python/_, dev-perl/_, dev-libs/* all collapse into "programming" group for example, virtual/* and acct-* have no bucket).

We can expand groups a bit, but I think for UIs the existing categories are fundamentally the wrong design, which is why nobody is using them. So adding them would give you nothing, since there simply isn't any (maintained) frontend support for this anymore.

> Gentoo has about ~350 name duplicates only in acct-group and acct-user categories, by design from this pov. Some are unanswerable without categories, like `emerge alembic` errors out as ambiguous and starts requesting the cat/pn format.

Another reason why this *must* be part of the package-ID. It will be interesting, because that means we will have slashes in package-IDs, which should work in theory but hasn't been tested much (I hope none of the code/frontends thinks that's a filepath, PK itself should be fine). PK2 will solve that.

For categorization, all existing frontends rely on the existing groups, which are easy to translate and have icons for that work in multiple themes and fit the GUI organization and style, or they rely on [AppStream](https://www.freedesktop.org/software/appstream/docs/) components for this to sort software into the well-known Menu categories in their UIs (see [the fd.o category registry](https://specifications.freedesktop.org/menu/latest/additional-category-registry.html).

Adding yet another way of categorization that failed to gain traction for over a decade is a very bad plan, which is why I am quite set on dropping this for PK2.
That said though, we would obviously like a Gentoo backend to work well, so, expanding existing groups is an option. But from what you wrote, it really feels like the category is part of the package's identity, and therefore must be encoded in a package-ID.

It might be that the freeform `data` field in PK2, once that is implemented (see https://github.com/PackageKit/PackageKit/pull/943) will actually make this easier for you (although I see a fairly strong case for just making the category part of the `name` part of a package-ID).

AntiqueH:
> Another reason why this _must_ be part of the package-ID. It will be interesting, because that means we will have slashes in package-IDs, which should work in theory but hasn't been tested much (I hope none of the code/frontends thinks that's a filepath, PK itself should be fine). PK2 will solve that.
>
> Adding yet another way of categorization that failed to gain traction for over a decade is a very bad plan, which is why I am quite set on dropping this for PK2. That said though, we would obviously like a Gentoo backend to work well, so, expanding existing groups is an option. But from what you wrote, it really feels like the category is part of the package's identity, and therefore must be encoded in a package-ID.
>

Exactly what we did.

`(chroot) gent ~ # pkgcli show tmux
Querying                                                       [████████████████████████████████████████] 100%
Package: app-misc/tmux
Version: 3.5a
Summary: Terminal multiplexer
License: ISC
URL: https://tmux.github.io/
Group: accessories
Installed Size: 610.0 KB
(chroot) gent ~ #

(chroot) gent /etc/portage # pkgcli list-depends tmux
Resolving dependencies                                         [████████████████████████████████████████] 100%
✔ dev-libs/libevent 2.1.12-r1.x86_64 [installed]
✔ sys-libs/ncurses 6.5_p20251220.x86_64 [installed]
✔ sys-apps/systemd 260.1-r2.x86_64 [installed]
✔ sys-libs/glibc 2.43-r2.x86_64 [installed]
(chroot) gent /etc/portage # 

(chroot) gent /etc/portage # pkgcli update
Querying                                                       [                  ░▓░                   ]     
The following packages have to be updated:
 app-arch/unzip_6.0_p29-r2.x86_64                          
 app-crypt/libmd_1.2.0.x86_64                              
 app-i18n/man-pages-l10n_4.30.0-r1.x86_64                  
 app-portage/gentoolkit_0.7.2.x86_64                       
 app-text/poppler_26.05.0-r1.x86_64                        
 dev-debug/gdb_17.2.x86_64                                 
 dev-lang/vala_0.56.19.x86_64                              
 dev-libs/kirigami-addons_1.12.1.x86_64                    
 dev-libs/libei_1.6.0.x86_64                               
 dev-libs/nspr_4.39.x86_64                                 
 dev-libs/plasma-wayland-protocols_1.21.0.x86_64 
 (...)

`

We already have a working prototype, but it's terrifyingly big in size compared to everything you would probably see in other package managers, since reverse-engineering even a part of portage dependency management, depgraph and deptree [+ the rest] was a very _interesting_ (to say the least) process. Luckily we're a bigger team, so we're advancing.

Right now we implemented so far only **prebuilt binary support** client, not the entirety of portage. It will take probably years(?) again to just rewrite portage only for the compiling packages part.
Nevertheless, we should at least have a working binhost prebuilt binary management, for whoever wishes to install prebuilt packages faster - portage already has good support here, so we cloned that behavior 100%.

The ticket can be closed, so we'll focus on the proper groups implementation.
Thank you a lot for your reply!

PR #1007, opened 2026-09-01
python: Use a raw string for regex

jelly:
Resolves two deprecation warnings:

SyntaxWarning: "\s" is an invalid escape sequence.
SyntaxWarning: "\]" is an invalid escape sequence.


ximion:
Thank you! :-)

Issue #399, opened 2020-05-06
alpm: assertion or segv on missing db files


eworm-de:
Running packagekit with missing db files makes it assert or dump core:

```
# pkcon update                        
Getting updates               [=========================]         
Finished                      [=========================]         
Downloading list of packages  [=========================]         
Testing changes               [         ==              ]         The daemon crashed mid-transaction!
```

Logs from journal:

```
May 06 10:20:45 nepu-d5a70936 systemd[1]: Starting PackageKit Daemon...
May 06 10:20:45 nepu-d5a70936 PackageKit[1975]: daemon start
May 06 10:20:45 nepu-d5a70936 systemd[1]: Started PackageKit Daemon.
May 06 10:20:48 nepu-d5a70936 PackageKit[1975]: get-updates transaction /1_bbabaace from uid 0 finished with success after 608ms
May 06 10:20:48 nepu-d5a70936 PackageKit[1975]: new update-packages transaction /2_ebbbccba scheduled from uid 0
May 06 10:20:49 nepu-d5a70936 systemd-coredump[2025]: Process 1975 (packagekitd) of user 0 dumped core.
May 06 10:20:49 nepu-d5a70936 systemd[1]: packagekit.service: Main process exited, code=killed, status=11/SEGV
May 06 10:20:49 nepu-d5a70936 systemd[1]: packagekit.service: Failed with result 'signal'.
```
or
```
May 06 09:11:26 nepu-d5a70936 systemd[1]: Starting PackageKit Daemon...
May 06 09:11:26 nepu-d5a70936 PackageKit[1422]: daemon start
May 06 09:11:26 nepu-d5a70936 systemd[1]: Started PackageKit Daemon.
May 06 09:11:33 nepu-d5a70936 PackageKit[1422]: get-updates transaction /1_eccbbdee from uid 1000 finished with success after 653ms
May 06 09:11:33 nepu-d5a70936 PackageKit[1422]: get-update-detail transaction /2_aaabeaeb from uid 1000 finished with failed after 1ms
May 06 09:11:33 nepu-d5a70936 PackageKit[1422]: get-details transaction /3_eaaeddcd from uid 1000 finished with failed after 0ms
May 06 09:11:40 nepu-d5a70936 PackageKit[1422]: new update-packages transaction /4_baedacea scheduled from uid 1000
May 06 09:11:40 nepu-d5a70936 packagekitd[1422]: **
[...]

PR #1010, opened 2026-09-08
alpm: fix segfault and abort on invalid-arch error

### LINKED ISSUE #757: pk_alpm_transaction_end: assertion failed: (pkalpm_current_job)
```
$ pkcon update --only-download
Getting updates                         [=========================]         
Finished                                [=========================]         
Updating packages                       [=========================]         
Downloading list of packages            [=========================]         
Starting                                [=========================]         The daemon crashed mid-transaction!
Finished                                [                         ] (0%) 
```

### `journalctl` log:

```
May 20 13:23:59 arch64 PackageKit[6530]: get-updates transaction /4_cdacedaa from uid 1000 finished with success after 7067ms
May 20 13:23:59 arch64 PackageKit[6530]: new update-packages transaction /5_bebacaaa scheduled from uid 1000
May 20 13:24:00 arch64 packagekitd[6530]: GError set over the top of a previous GError or uninitialized memory.
                                          This indicates a bug in someone's code. You must ensure an error is NULL before it's set.
                                          The overwriting error message was: unable to lock database
May 20 13:24:00 arch64 packagekitd[6530]: **
May 20 13:24:00 arch64 packagekitd[6530]: PackageKit-alpm:ERROR:../PackageKit-1.2.8/backends/alpm/pk-alpm-transaction.c:1115:pk_alpm_transaction_end: assertion failed: (pkalpm_current_job)
May 20 13:24:00 arch64 packagekitd[6530]: Bail out! PackageKit-alpm:ERROR:../PackageKit-1.2.8/backends/alpm/pk-alpm-transaction.c:1115:pk_alpm_transaction_end: assertion failed: (pkalpm_current_job)
```

### LINKED ISSUE #1009: alpm: SIGSEGV in pk_alpm_transaction_simulate — ALPM_ERR_PKG_INVALID_ARCH data list holds char*, not alpm_pkg_t*
OK, I admit I had a lot of help from Claude/Opus on this, but here we go...

## Summary

`pk_alpm_transaction_simulate()` type-confuses the `data` list that libalpm returns for `ALPM_ERR_PKG_INVALID_ARCH`. libalpm fills that list with `char *` strings; PackageKit passes it to `pk_alpm_pkg_build_list()`, which calls `alpm_pkg_get_name()` on each element as if it were an `alpm_pkg_t *`. The resulting garbage pointer is handed to `g_string_append_printf("%s, ", …)` and the daemon segfaults.

The crash happens on the *error-reporting* path, so the actual problem (an architecture mismatch) is never reported to the user — the daemon dies instead.

A second defect then leaks the pacman lock, which is what makes this user-visible and persistent. Details in the last section.

Related but distinct: #399 (segv/assert on missing db files) and #757 (the `pkalpm_current_job` assertion). Neither identifies the type confusion below; #757 describes the assertion I hit as the follow-on crash.

## The bug

`backends/alpm/pk-alpm-transaction.c`:

```c
switch (alpm_errno (priv->alpm)) {
case ALPM_ERR_PKG_INVALID_ARCH:
        prefix = pk_alpm_pkg_build_list (data);
```

```c
static gchar *
pk_alpm_pkg_build_list (const alpm_list_t *i)
{
        ...
        for (; i != NULL; i = i->next) {
                if (i->data == NULL)
                        continue;
                g_string_append_printf (list, "%s, ",
                                        alpm_pkg_get_name (i->data));
        }
```

But libalpm builds that list from strings, not packages — `lib/libalpm/trans.c`, `check_arch()`:

```c
if(!found) {
        char *string;
        const char *pkgname = pkg->name;
        const char *pkgver = pkg->version;
        size_t len = strlen(pkgname) + strlen(pkgver) + strlen(pkgarch) + 3;
        MALLOC(string, len, RET_ERR(handle, ALPM_ERR_MEMORY, invalid));
        snprintf(string, len, "%s-%s-%s", pkgname, pkgver, pkgarch);
        invalid = alpm_list_add(invalid, string);
}
```

and hands it straight to the caller:

```c
alpm_list_t *invalid = check_arch(handle, trans->add);
if(invalid) {
        if(data) {
                *data = invalid;
        }
        RET_ERR(handle, ALPM_ERR_PKG_INVALID_ARCH, -1);
}
```

So `alpm_pkg_get_name()` reads `pkg->name` at a struct offset inside a heap-allocated string, yielding a wild pointer that `%s` then dereferences.

The `NULL` guard on `i->data` does not help: the elements are non-NULL, they are just the wrong type.

The other three branches of the same `switch` (`UNSATISFIED_DEPS`, `CONFLICTING_DEPS`, `FILE_CONFLICTS`) do appear to match libalpm's actual types.

## Why it fires reliably on x86-64-v3/v4 distributions

`pk-alpm-config.c` reimplements `Architecture = auto` instead of leaving it to libalpm:

```c
if (g_strcmp0 (arch, "auto") == 0) {
        struct utsname un;
        uname (&un);
        config->arch = g_strdup (un.machine);
}
```

then overrides libalpm's own detection:

```c
alpm_option_set_architectures (handle, arches_list);
```

`uname()` returns `x86_64` on every x86-64 machine regardless of microarchitecture level. On distributions that ship `x86_64_v3`/`x86_64_v4` packages, ~~current libalpm resolves `auto` to the full supported set~~ **[corrected: this is a downstream libalpm patch, not upstream behaviour — see [correction](https://github.com/PackageKit/PackageKit/issues/1009#issuecomment-5585141369)]** (my `libalpm.so.16.0.1` contains `cpu supports x86_64_v4` / `x86_64_v4`), but PackageKit replaces that list with the single value `x86_64`.

Every installed package then fails `check_arch()`, so `alpm_trans_prepare()` returns `ALPM_ERR_PKG_INVALID_ARCH` with a fully-populated `data` list on **every** update — and the code above segfaults on it every time. That is why this reproduces 100% of the time here rather than being an occasional crash.

~~This is arguably a third bug on its own: reimplementing `auto` from `uname()` discards libarch detection that libalpm already does correctly.~~

**Corrected — upstream pacman's `config_add_architecture()` does the same `uname()` resolution, so this is not a PackageKit deviation. See [correction](https://github.com/PackageKit/PackageKit/issues/1009#issuecomment-5585141369). The crash analysis above is unaffected.**

## Suggested reproducer without special hardware

On a stock `x86_64` Arch install, set an architecture that no installed package matches:

```
Architecture = x86_64_v4
```

then run an update through PackageKit. `check_arch()` should flag every package and take the same branch.

(I have not run this synthetic case myself — my crashes are from the real CachyOS configuration described below.)

## Backtrace

```
Stack trace of thread 191429:
#0  0x00007fad12fb335c n/a (libc.so.6 + 0x1b335c)
#1  0x00007fad12e77398 n/a (libc.so.6 + 0x77398)
#2  0x00007fad12e9f471 n/a (libc.so.6 + 0x9f471)
#3  0x00007fad13813ad6 g_vasprintf (libglib-2.0.so.0 + 0xe7ad6)
#4  0x00007fad137cf1bb g_string_append_vprintf (libglib-2.0.so.0 + 0xa31bb)
#5  0x00007fad137cf471 g_string_append_printf (libglib-2.0.so.0 + 0xa3471)
#6  0x00007fad12c3bcf3 pk_alpm_transaction_simulate (libpk_backend_alpm.so + 0xecf3)
#7  0x00007fad12c3da2c n/a (libpk_backend_alpm.so + 0x10a2c)
#8  0x0000562c427f5ffb n/a (packagekitd + 0x2dffb)
#9  0x00007fad137dc6e3 n/a (libglib-2.0.so.0 + 0xb06e3)
Signal: 11 (SEGV) si_code: SI_KERNEL
```

No debug symbols are published for my distribution's build, so instead of a symbolised frame I disassembled the crash site. `libpk_backend_alpm.so + 0xecf3` is the return address of the call at `0xeced`:

```asm
ecd0:  mov    (%rbx),%rdi           ; i->data
ecd3:  test   %rdi,%rdi
ecd6:  je     ecf3                  ; if (i->data == NULL) continue;
ecd8:  call   *...                  ; alpm_pkg_get_name (i->data)
ecde:  mov    %rax,%rdx             ; -> printf argument
ece1:  lea    0x2a4c(%rip),%rsi     ; format string @ 0x11734
ece8:  mov    %r12,%rdi             ; GString *list
eced:  call   *...                  ; g_string_append_printf   <-- SEGV
ecf3:  mov    0x10(%rbx),%rbx       ; i = i->next
ecf7:  test   %rbx,%rbx
ecfa:  jne    ecd0
ecfc:  mov    0x8(%r12),%rax        ; list->len
ed01:  cmp    $0x2,%rax
ed05:  ja     ed78                  ; if (list->len > 2)
```

The format string at `0x11734` is `"%s, "`. The `i->data == NULL` guard combined with the `list->len > 2` check identifies this unambiguously as `pk_alpm_pkg_build_list()` inlined into `pk_alpm_transaction_simulate()` — it is the only one of the four helpers with both.

## Consequence: leaked pacman lock

The daemon is reactivated over D-Bus, and the new process aborts in `pk_alpm_transaction_end()` (this is #757):

```
PackageKit-alpm:ERROR:../PackageKit/backends/alpm/pk-alpm-transaction.c:1115:
pk_alpm_transaction_end: assertion failed: (pkalpm_current_job)
```

The ordering there matters:

```c
g_assert (pkalpm_current_job);
pkalpm_current_job = NULL;

if (alpm_trans_release (priv->alpm) < 0) {   /* never reached */
```

Because the assertion precedes `alpm_trans_release()`, `/var/lib/pacman/db.lck` is *never* released on this path. Every subsequent PackageKit operation then fails immediately:

```
refresh-cache transaction /2_eabadbac from uid 1000 finished with failed after 0ms
```

From the user's side this presents as Discover reporting "There was an issue during the update or installation process" on every attempt, indefinitely, until the lock is removed by hand — with no indication that a lock is the problem. Two full days elapsed between the crash and my noticing, during which the system silently could not be updated by any tool that takes the pacman lock.

A `g_assert()` that aborts a long-running system daemon and leaks a global lock file seems worth replacing with a non-fatal check that still reaches `alpm_trans_release()`.

## Environment

- PackageKit 1.3.6
- pacman 7.1.0.r9.g54d9411-4, libalpm.so.16.0.1
- KDE Discover 6.7.4 (frontend that triggered it)
- CachyOS (Arch-based), `x86_64_v4` package set, `Architecture = auto` in `/etc/pacman.conf`
- Kernel 7.2.2-1-cachyos

Reproduced twice, on separate days, from two different boots — each time producing the SIGSEGV/SIGABRT pair and a leaked `db.lck`. Both coredumps are retained if anything further would help.

PandaWood:
As with the issue, I had help from Claude/AI on this.
But I ran all these tests on my machine and confirmed everything so it's not automated slop.
I confirmed everything follows the contribution rules (had to force push a change to conform).

Fixes #1009, fixes #757, and the remaining transaction-assert half of #399.

`pk_alpm_transaction_simulate()` type-confuses the `ALPM_ERR_PKG_INVALID_ARCH`
error list: libalpm fills it with malloc'd `name-version-arch` strings
(`check_arch()` in `lib/libalpm/trans.c`), but the backend passed it to
`pk_alpm_pkg_build_list()`, which calls `alpm_pkg_get_name()` on each entry.
The wild pointer is dereferenced by `g_string_append_printf()` and the daemon
segfaults mid-transaction, leaving `/var/lib/pacman/db.lck` behind.

From then on every transaction fails in `alpm_trans_init()` with
`ALPM_ERR_HANDLE_LOCK`. `pk_alpm_transaction_initialize()` returns before
setting `pkalpm_current_job`, callers still reach `pk_alpm_transaction_end()`
for cleanup, and its `g_assert (pkalpm_current_job)` aborts the daemon on
every later operation. That is the sequence in #757.

Changes, all in `backends/alpm/pk-alpm-transaction.c`:

- Use the existing string-joining helper for the INVALID_ARCH list. It is
  moved above the new caller, gains NULL-element and empty-list guards, and
  the strings are freed afterwards as pacman does.
- Free the strings for `ALPM_ERR_PKG_INVALID` in the commit path as well;
  libalpm hands over malloc'd filenames there too.
- Replace the `g_assert` in `pk_alpm_transaction_end()` with a syslog warning,
  so a failed initialize is reported instead of aborting the daemon, and
  `alpm_trans_release()` is always attempted.

### Testing

PackageKit 1.3.6 (patch applied to the v1.3.6 tag, identical file), pacman
7.1.0, libalpm 16.0.1, CachyOS x86_64_v4.

PackageKit resolves `Architecture = auto` to `x86_64` via `uname()`, so on a
system with `x86_64_v4` packages any install reaches the error path. The
synthetic case from #1009 also works: `Architecture = x86_64_v4` plus an
install of any `x86_64` package (`figlet`). Note the transaction must add at
least one package; when nothing is pending, `alpm_trans_prepare()` returns
before `check_arch()` runs, so `update` on an up-to-date system does not
trigger it.

Before the patch:

- `pkgcli install sl`: "The daemon crashed mid-transaction!", SIGSEGV with
  `g_string_append_printf` <- `pk_alpm_transaction_simulate` on the stack,
  `db.lck` left behind.
- Same command again: SIGABRT,
  `pk-alpm-transaction.c:1115:pk_alpm_transaction_end: assertion failed:
  (pkalpm_current_job)`.

After the patch:

- `pkgcli install sl`: `Error: sl-5.05-6.1-x86_64_v4: package architecture
  is not valid`. No coredump, no `db.lck`.
- With a stale `db.lck` present: `Error: unable to lock database`, a
  "transaction end without an active transaction" warning in the journal,
  no coredump.

### DIFF
--- a/backends/alpm/pk-alpm-transaction.c
+++ b/backends/alpm/pk-alpm-transaction.c
@@ -26,6 +26,7 @@
 #include "pk-alpm-packages.h"
 #include "pk-alpm-transaction.h"
 
+#include <stdlib.h>
 #include <syslog.h>
 
 static off_t transaction_dcomplete = 0;
@@ -792,8 +793,10 @@ pk_alpm_transaction_initialize (PkBackendJob* job, alpm_transflag_t flags, const
 	return TRUE;
 }
 
+/* joins a list of strings, as returned by libalpm for
+ * ALPM_ERR_PKG_INVALID_ARCH and ALPM_ERR_PKG_INVALID */
 static gchar *
-pk_alpm_pkg_build_list (const alpm_list_t *i)
+pk_alpm_string_build_list (const alpm_list_t *i)
 {
 	GString *list;
 
@@ -803,8 +806,7 @@ pk_alpm_pkg_build_list (const alpm_list_t *i)
 	for (; i != NULL; i = i->next) {
 		if (i->data == NULL)
 			continue;
-		g_string_append_printf (list, "%s, ",
-					alpm_pkg_get_name (i->data));
+		g_string_append_printf (list, "%s, ", (const gchar *) i->data);
 	}
 
 	if (list->len > 2)
@@ -947,7 +949,11 @@ pk_alpm_transaction_simulate (PkBackendJob *job, GError **error)
 
 	switch (alpm_errno (priv->alpm)) {
 	case ALPM_ERR_PKG_INVALID_ARCH:
-		prefix = pk_alpm_pkg_build_list (data);
+		/* libalpm fills this list with malloc'd "name-version-arch"
+		 * strings (check_arch() in lib/libalpm/trans.c), not
+		 * alpm_pkg_t pointers */
+		prefix = pk_alpm_string_build_list (data);
+		alpm_list_free_inner (data, free);
 		alpm_list_free (data);
 		break;
 	case ALPM_ERR_UNSATISFIED_DEPS:
@@ -1027,21 +1033,6 @@ pk_alpm_transaction_packages (PkBackendJob *job)
 	}
 }
 
-static gchar *
-pk_alpm_string_build_list (const alpm_list_t *i)
-{
-	GString *list;
-
-	if (i == NULL)
-		return NULL;
-	list = g_string_new ("");
-	for (; i != NULL; i = i->next)
-		g_string_append_printf (list, "%s, ", (const gchar *) i->data);
-
-	g_string_truncate (list, list->len - 2);
-	return g_string_free (list, FALSE);
-}
-
 gboolean
 pk_alpm_transaction_commit (PkBackendJob *job, GError **error)
 {
@@ -1071,6 +1062,7 @@ pk_alpm_transaction_commit (PkBackendJob *job, GError **error)
 		break;
 	case ALPM_ERR_PKG_INVALID:
 		prefix = pk_alpm_string_build_list (data);
+		alpm_list_free_inner (data, free);
 		alpm_list_free (data);
 		break;
 	default:
@@ -1112,7 +1104,13 @@ pk_alpm_transaction_end (PkBackendJob *job, GError **error)
 	if (tpkg != NULL)
 		pk_alpm_transaction_output_end ();
 
-	g_assert (pkalpm_current_job);
+	/* pk_alpm_transaction_initialize() may have failed, e.g. because
+	 * the pacman database is locked (ALPM_ERR_HANDLE_LOCK), and callers
+	 * still reach here to clean up.  That must not abort the daemon, and
+	 * alpm_trans_release() must always be attempted so a transaction is
+	 * never left holding the database lock. */
+	if (pkalpm_current_job == NULL)
+		syslog (LOG_DAEMON | LOG_WARNING, "transaction end without an active transaction");
 	pkalpm_current_job = NULL;
 
 	if (alpm_trans_release (priv->alpm) < 0) {


ximion:
Build fails, but on a trivial thing:
```
 ../backends/alpm/pk-alpm-transaction.c:797:1: error: 'pk_alpm_pkg_build_list' defined but not used [-Werror=unused-function]
  797 | pk_alpm_pkg_build_list (const alpm_list_t *i)
      | ^~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors

[85/261] Compiling C object backends/dummy/libpk_backend_dummy.so.p/pk-backend-dummy.c.o

PandaWood:
Thanks, fixed. `pk_alpm_pkg_build_list()` had no callers left once the INVALID_ARCH branch switched to the string helper, so I've dropped it and noted that in the commit message. Rebuilt locally with `-Dwerror=true` to match CI this time.

ximion:
It is fine to use LLMs to assist with your work, but please don't use them to write text for your PRs. That should be your own words, which makes it both easier to read and also shows you understood the issue.
TBH, this was insanely hard to read... Thank you for disclosing AI use though!

Thanks for the patch, this looks safe to merge. This backend really needs active maintainers though who can properly test things (I am not on Arch).

PandaWood:
Thanks, yeah I'm getting better at combining AI fixes on a topic slightly over my head ;-)
This was one of my first, will be easier to read next time.

PR #1015, opened 2026-09-19
lib: Make the yes/no console prompt translatable

TomaszBojanowski:
The "[Y/n]" / "[N/y]" suffix was hardcoded and only the English answers "y", "yes", "n" and "no" were accepted, so in other languages the question was translated but the expected answers were not.

Mark the suffix and the answers as translatable. The English answers are still always accepted, so existing scripts and habits keep working.

Testing:
- Built with `ninja -C build`; `meson test` passes (the e2e test was skipped locally since it needs root, so leaving that to CI).
- With a temporary Polish translation of the new strings, `pk-command-not-found` showed `[N/t]`, accepted `t` as yes and `nie` as no.
- Ran `./autoformat.py` on the changed file.


ximion:
This looks okay, thank you for the patch!

TomaszBojanowski:
@ximion Could you regenerate po/PackageKit.pot when you get a chance? The new [Y/n] / [N/y] strings from #1015 aren't in the template yet, so they can't be translated on Weblate.

PR #1016, opened 2026-09-19
command-not-found: Don't report a declined install as a failure

TomaszBojanowski:
When the user answered "no" to the transaction confirmation, the tool printed "Failed to install packages: user declined simulation" right after "The transaction did not proceed.". The untranslated message was confusing, since nothing failed and the user's choice had already been acknowledged.

Return quietly when the error is DECLINED_SIMULATION or DECLINED_INTERACTION.

Testing:
- Ran `./autoformat.py` on the changed file; `ninja -C build` succeeds.
- Ran `pk-command-not-found cowsay`, answered "y" and then "n" at the confirmation: it now stops after "The transaction did not proceed." without the extra error line.


TomaszBojanowski:
@coderabbitai review

